Issue #56, opened 2013-10-02
Feature Request: Ability to have mappings recorded un-gzipped even when the server responds with a gzipped response.


yogurtearl:
It could be useful to have mappings recorded uncompressed. This would allow for easier editing of the body files. Of course, you would want to remove "Content-Encoding" : "gzip" from the response header as well. 

Maybe it could be a command line option --store-mappings-uncompressed

txjohng:
Hi guys, this feature would be a super useful feature where I work. Is there anyway to get this change into the next release?  thanks!

tomakehurst:
GZip support now in 2.0.7-beta

PR #78, opened 2013-12-22
Ungzip recorded responses if --ungzip-recorded-responses cli param given

rowanhill:
Hi,

This PR (hopefully) closes issue #56: it adds a new command line arg, `--ungzip-recorded-responses`, which decompresses gzipped recorded responses (and removes the "Content-Encoding: gzip" header from the stored mapping).

One thought I've had is that this is maybe a bit gzip-specific. Maybe it should just be `--decompress-recorded-responses` or something, in case WireMock ever ends up supporting any other kinds of decompression, so the cli args don't need to change? Or maybe I'm over-thinking that.

As always, happy to implement any changes you might suggest, if I can.

Thanks,
Rowan

### DIFF
--- a/docs/source/record-playback.rst
+++ b/docs/source/record-playback.rst
@@ -32,6 +32,16 @@ for the body content.
 .. note::
     The WireMock recorder will ignore a request with a method and URL identical to those of a stub already recorded.
 
+GZip decompressing
+------------------
+
+WireMock can be configured to decompress any gzipped responses it receives when recording, making editing the resulting
+body files much easier. WireMock also removes the ``Content-Encoding: gzip`` header when playing back the response. To
+enable gzip decompression, add the ``--ungzip-recorded-responses`` argument when starting the standalone runner, e.g.:
+
+.. parsed-literal::
+
+    $ java -jar wiremock-|version|-standalone.jar --proxy-all="http://search.twitter.com" --record-mappings --ungzip-recorded-responses --verbose
 
 Playback
 ========
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -140,9 +140,13 @@ public class WireMockServer {
 		stubRequestHandler.addRequestListener(listener);
 	}
 	
-	public void enableRecordMappings(FileSource mappingsFileSource, FileSource filesFileSource) {
+	public void enableRecordMappings(
+            FileSource mappingsFileSource,
+            FileSource filesFileSource,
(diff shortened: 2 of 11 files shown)


txjohng:
Hi guys, this feature would be a super useful feature where I work. Is there anyway to get this change into the next release?  thanks!

handstandsam:
+1 on this.  It's annoying to manually have to go through the responses and un-gzip them.  I don't have control over the client that is making these requests and therefore can't remove the header.

tomakehurst:
I added gzip support in 2.0.7-beta.

Now when the recorder sees that the response is gzipped it unzips it before saving the file. Also, when ``Accept-Encoding: gzip`` is requested, WireMock will gzip the response.

Does that solve your problem?

handstandsam:
@tomakehurst.  You are correct, it did unzip the gzipped response and that was very helpful with this new 2.0.7-beta version.  I just figured out that I ran into issues because the 'Content-Length' response header was now incorrect.  Is that an issue that should be tracked?

*Issue:* With 2.0.7-beta, when responses are saved in an un-gzipped format, the "Content-Length" response header is not updated and therefore doesn't match the new response body.  *Recommendation:* Remove the 'Content-Length' response header, or calculate it based on the new un-gzipped response payload.  This will allow the "Playback" mode to work without any modification of mapping files.

tomakehurst:
So if I understand correctly: the recording ends up containing the ungzipped content-length header, and because it is present Jetty isn't calculating the gzipped version, but instead just passing the recorded version. So (theoretically) removing Content-Length from the recorded mapping should force Jetty to calculate it itself.

handstandsam:
@tomakehurst That is correct in my usage of *2.0.7-beta*.  Removing the ```"Content-Length"``` header from the recorded response mapping solved this problem.  Jetty then calculated the ```"Content-Length"``` instead of the recorded, un-gzipped JSON file.

tomakehurst:
Cool, I'll fix that before the next release then.

eeichinger:
note that suppressing gzip and thus ensure proper recording also works (with correctly configured webservers!) by setting the "accept-encoding: identity" header using a simple mapping in your "mappings" folder as this:

```
{
  "priority": 10,
  "request": {
    "method": "ANY",
    "urlPattern": ".*"
  },
  "response": {
    "proxyBaseUrl" : "https://other.com",
    "additionalProxyRequestHeaders": {
      "Accept-Encoding": "identity"
    }
  }
}
```

this technique works for all wiremock versions

tomakehurst:
Thanks for sharing that, useful to know.

tomakehurst:
Closing, as gzip is supported in both directions now.

PR #98, opened 2014-02-13
Added remove a mapping feature

shashank-neo:
Have created a new Pull request after rebasing.

Objective:  Allow users to delete a particular mapping.
Implementation:
1.Now whenever a new mapping is created it returns the id of the mapping which is like an index.
eg {"id":0}
2. I have added an end point "__admin/mappings/remove" it takes a body eg {"id":0}, User needs to make a DELETE call on the url, if a mapping exists with the given id it will be deleted else it will return 404.


tomakehurst:
Remove now added via a more recent PR.

PR #103, opened 2014-03-02
Scenario builder generics

rowanhill:
This follows on from #84: instead of throwing an exception when building a mapping that specifies a required / next state without specifying a scenario name, this patch makes use of the type checker to ensure that isn't possible.

I'm not really certain this is the best way to go, but after sitting on it for way too long, I figured I may as well just put up a PR and let there be a discussion about it; I won't be offended if we just chuck this away!

Pros:
 - The API doesn't allow invalid use of itself
 - IDEs will be helpful and only suggest methods that make sense.

Cons:
 - It's overly restrictive, and so not 100% backwards compatible: it'll cause any existing code that calls whenScenarioStateIs / willSetStateTo before inScenario to fail to compile now (although it's a simple fix)
 - It's fiddly generics code that's possibly more complicated than it's worth

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/MappingBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/MappingBuilder.java
@@ -69,6 +69,9 @@ public class MappingBuilder {
 	}
 	
 	public StubMapping build() {
+		if (scenarioName == null && (requiredScenarioState != null || newScenarioState != null)) {
+			throw new IllegalStateException("Scenario name must be specified to require or set a new scenario state");
+		}
 		RequestPattern requestPattern = requestPatternBuilder.build();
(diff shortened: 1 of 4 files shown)


rowanhill:
I notice there's a 2.0 version in the works, with some breaking changes - is it worth reconsidering this PR (or at least the spirit of it, since it's so out of date now)?

tomakehurst:
Yes, absolutely! Was thinking this would be a good thing for v2.

Would you be interested in creating a new PR?

rowanhill:
Yes, if I can find a moment before going back to work. Should the PR be against master, or is there a v2 branch you're working on?

rowanhill:
Okay, there's now a 2.0-friendly PR over at #339, so I'll close this.

PR #145, opened 2014-06-11
Add description field to StubMapping

alins:
the way i figure, if the field is missing in existing mappings the field will get a null value so it should be backward compatible.


tomakehurst:
Closing due to inactivity

PR #178, opened 2014-10-01
Android httpclient is fixed at 4.0-beta1. Use an upgraded client in a different namespace.

objcode:
Use a renamed library to upgrade to a modern httpclient on android.

This solution works (and passes tests) on both Android and the JDK.


indraneel76:
Can i get the updated Jar file . I just need to use in my android project

tomakehurst:
There isn't one yet unfortunately.

Might be worth looking at Roboelectric in the meantime, as I believe people have had some success using this with WireMock.

PR #217, opened 2014-12-29
Making the number of container threads configurable.

chbatey:


k-anil:
Hi Tom/Chbatey
Same I also want to load test and I am getting issues in IO with jetty.
Could you pelase guide us how this issue can be resolved?

Thanks
Anil

PR #218, opened 2014-12-30
Support for parametrization in reponse body files via Velocity templates.

radAdam:
Hi Tom , hope all is well. 

Recently I had the need to support parametrization in stubbed service calls. Essentially I was stubbing part of a user registration flow and using JMeter to help load test.The flow requires a unique user guid to be passed in, and it in turn does a look up on this id, and generates a hash derived from the guid passed via the request. Because I am using an incrementer from within a JMeter plan and invoking tens of thousands of requests concurrently, I was having to generate tens of thousands of mocks. A wild card, catch all response mapping would not work in this scenario.

I added in support for Velocity templates. The request is processed in the handler , and stored in a context object. Before the response is rendered , I do a check to see if the body file is of type .vm , if so the context object is merged with template, and I can process or extrapolate information from the request into the response dynamically. 

Thought perhaps this may benefit others.

Cheers !
-adam

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ copy-admin.sh
 .DS_Store
 **/.DS_Store
 
+tmp
--- a/build.gradle
+++ b/build.gradle
@@ -24,6 +24,7 @@ dependencies {
         	"com.fasterxml.jackson.core:jackson-annotations:2.4.2",
         	"com.fasterxml.jackson.core:jackson-databind:2.4.2"
 	compile "org.apache.httpcomponents:httpclient:4.3.5"
+	compile "org.apache.velocity:velocity:1.7"
     compile "org.skyscreamer:jsonassert:1.2.3"
     compile "xmlunit:xmlunit:1.5"
     compile "com.jayway.jsonpath:json-path:0.8.1"
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -37,6 +37,9 @@ import com.github.tomakehurst.wiremock.standalone.MappingsLoader;
 import com.github.tomakehurst.wiremock.stubbing.StubMappingJsonRecorder;
 import com.github.tomakehurst.wiremock.stubbing.StubMappings;
 import com.github.tomakehurst.wiremock.verification.LoggedRequest;
+
+import org.apache.velocity.VelocityContext;
+import org.apache.velocity.app.Velocity;
 import org.mortbay.log.Log;
 
 import java.util.List;
@@ -57,6 +60,8 @@ public class WireMockServer implements Container, Stubbing {
 	private final Notifier notifier;
 
     private final Options options;
+    
+    private final VelocityContext velocityContext;
 
     protected final WireMock client;
 
@@ -64,6 +69,10 @@ public class WireMockServer implements Container, Stubbing {
         this.options = options;
         this.fileSource = options.filesRoot();
         this.notifier = options.notifier();
+        
+        Velocity.init();
+        velocityContext = new VelocityContext();
+        velocityContext.put( "name", new String("Wiremock Request Context") );
 
         RequestDelayControl requestDelayControl = new ThreadSafeRequestDelayControl();
         MappingsLoader defaultMappingsLoader = makeDefaultMappingsLoader();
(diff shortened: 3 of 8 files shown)


radAdam:
Hey Tom , I've completed a first pass at the extension. There were a couple things I ran into while implementing this. Code is @

https://github.com/radAdam/wiremock-velocity-extension

Firstly, I could not use the cli statement as we originally had discussed. Trying to use -jar and -cp together did not work the way I had hoped. Everything I read essentially said using java -jar will take precedence over java -cp.(unless you know of a way around this ?).

Because of this , I had to switch to using java -cp exclusively, which results in this command 
````
java -cp "tmp/wiremock-velocity-extension-0.1.jar:tmp/wiremock-1.52-standalone.jar" com.github.tomakehurst.wiremock.standalone.WireMockServerRunner --verbose --root-dir tmp/default/ --extensions com.ayork.wiremock.extension.VelocityResponseTransformer
````
It is a bit verbose but manageable. However , using that approach many classes were not found. I had to include all of your dependencies into my gradle build and build a fat jar. It would be great if I didn't have to include those, perhaps you have a better idea how to handle that ?

Another thing I ran into was that initially my response was not being rendered with the velocity context. What I saw happen was that the transformation was occurring correctly on the responseDefinition , however the actual response body was being returned to me as an un-rendered template. I debugged my way through way wiremock and what happens is the transformed responseDefinition is passed into the StubResponeRenderer.renderDirectly method, which in lines 59-61 rebuild the body from the bodyFile; effectively clobbering the transformed responseDefinition if a bodyFileName exists.

What I ended up doing was just setting the bodyFileName to null. That allowed control flow  to direct to  responseBuilder.body(responseDefinition.getByteBody()), which is what I wanted to happen.

Thoughts or suggestions greatly appreciated , thanks !

tomakehurst:
Thanks for the feedback Adam. Sounds like there's some scope for improvement. Let me look into it a bit further.

tomakehurst:
Regarding your second problem, are you using ``ResponseDefinitionBuilder.like(...)``?

I can see a couple of solutions - 1) Make ``ResponseDefinition`` prefer body content if both that and a file are specified (currently it's undefined, but the body file loading happens to be first in the renderer), or 2) Make ``ResponseDefinition`` throw an exception if both are set, and fix ``ResponseDefinitionBuilder`` so that it either prevents you from setting both, or nulls whichever one you didn't set.

2) seems better to me.

I'd forgotten that you can't combine ``-jar`` and ``-cp``. I guess there's potential for adding another CLI param for the extension JAR and then doing the class loading programmatically. That would save some keystrokes, although I'm not sure how much value this would add, as I guess once you've set up your build/written a wrapper script you're likely to stop caring.

radAdam:
No , I am not using ResponseDefinitionBuilder.like(). I am just modifying properties on this ResponseDefinition passed into to the transform method. Should I be creating a new ResponseDefinition
and just referencing value's from the original ?

As for the command line invocation , I agree the verbosity isn't much of an issue , was more hoping you could see a way that I didn't need to include all the wiremock dependencies in my build.

tomakehurst:
FYI, I've now changed (not released yet) ResponseDefinition slightly so that it'll use the body in preference to the bodyFile if both are specified. Hope this helps.

radAdam:
Awesome , thanks Tom. Any idea when you'll be cutting a new release build ? I just need to get some unit tests around the code I wrote and update the README , but I've been holding off until you release this functionality officially.

tomakehurst:
The response transformer extension is actually already in the currently released version, so you could go ahead and release with your workaround if you liked. Really the only change to that extension type I'm planning is the fix to prioritise the body over bodyFile.

Not sure when I'll be releasing unfortunately. Time is scarce at the moment, and I'm starting a new contract tomorrow so I'm sure things are going to get worse in the short term.

radAdam:
Great , thanks Tom. I have completed initial release located here, https://github.com/radAdam/wiremock-velocity-transformer. Eventually ill get this into maven central, but for now its simply available for download from github.Closing the pull request. Thanks again !

tomakehurst:
Nice work! Let me know when you've published it to Maven central and I'll link to it from the doc site.

radAdam:
finally got around to publishing this. In mavencentral,

<dependency>
  <groupId>com.github.radadam</groupId>
  <artifactId>wiremock-velocity-transformer</artifactId>
  <version>1.1</version>
</dependency>

Tested and built with 1.55 wiremock.

tomakehurst:
Nice work! I'll add a link from the docs next time I'm working on it if you like.

radAdam:
Sounds good.

PR #223, opened 2015-01-07
Added path to remove single mapping

nhu313:
I have mix feeling about this. On the one hand, I don't think it is a good practice. It is adding state to Wiremock that doesn't need to be. But in my current situation, I need wiremock to act as an external service to write acceptance test for my service.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -294,6 +294,11 @@ public class WireMockServer implements Container, Stubbing, Admin {
     }
 
     @Override
+    public void removeStubMapping(StubMapping stubMapping) {
+        wireMockApp.removeStubMapping(stubMapping);
+    }
+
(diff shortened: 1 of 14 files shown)


tomakehurst:
If you need to effectively remove a stub, you can always register one that is identical the the old one, but returns a 404. That will have the same effect.

Admittedly this isn't ideal if you're running WireMock for very long periods without restarting, so I can see some value in being able to delete.

I think there are a couple of implementation options that might be a bit easier to work with - 
1) Delete all mappings with a request matching the supplied request pattern (allowing selective bulk deletes but losing a bit of control if you have two identical request patterns with different responses), or
2) Return an ID when creating a stub then using that as a handle for deletion. I think returning an ID has some utility outside of deletion too, e.g. if you want to link stubs to external data.

nhu313:
Good idea. I didn't think about sending 404 as delete.

Request pattern and ID both sounds like good options. For the id route, it might be easier to work off pull request #98. I'm going to close this pull request.

PR #225, opened 2015-01-14
Change JsonFileMappingsLoader.java to not exit wiremock due to bad json in mapping

tedleman:
Hi Tom,

I made a very simple change in JsonFileMappingsLoader.java to not exit wiremock if an exception is thrown due to bad json in a single mapping, instead printing the name of the mapping and error message to standard output. Wiremock would then launch without that mapping. Instead of uploading this small change to my company's internal repo, I was wondering if you would be willing to pull this change so that we can easily keep up with your updates in the future.

Thanks,
Troy

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/JsonFileMappingsLoader.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/JsonFileMappingsLoader.java
@@ -23,6 +23,8 @@ import com.google.common.base.Predicate;
 
 import static com.google.common.collect.Iterables.filter;
 
+import static java.lang.System.out;
+
 public class JsonFileMappingsLoader implements MappingsLoader {
 
 	private final FileSource mappingsFileSource;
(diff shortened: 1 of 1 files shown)


outofrange:
Why should Wiremock start and load a configuration you want to use, if it's not valid?
I would expect it to either load all my configured mappings, or stop and do nothing at all.

tomakehurst:
I have to admit I'm pretty dubious about this one too. Letting failures slip quietly by is usually a recipe for hours of frustrated debugging in my experience.

outofrange:
I guess a non-default cli parameter like --ignoreInvalidMappingFiles (or something like this) could be okay. But I still can't see a use case for this - if the mapping is invalid, I want to know to delete or fix it before I start Wiremock.

tomakehurst:
Yeah, think I'm going to leave this one out. Sorry.

tedleman:
Sorry for the late response, and I completely understand if you don't want this option included. 

The use case I am going for is when multiple developers are trying to access a single instance of wiremock, while mocking calls for entirely different services. When developer A accidentally uploads a mapping with invalid json for his service, developer B (who is trying to mock with a different service) can no longer access wiremock until developer A gets his act together.

tomakehurst:
I can understand the desire to do that certainly, but I tend to view problems like that as being people/process centric at their root, and trying to mitigate them with technical countermeasures tends to make things worse in the long run.

FWIW, I tend not to use JSON files very much any more on my own projects, except sometimes in the first couple of weeks when getting things up and running. 

I've found the most effective approach is to write builder classes to construct JSON or XML payloads and write a wrapper class around WireMock with a load of service methods that take these builders as parameters e.g. ``cmsStub.createNewsStory(myNewsStoryBuilder);``

That way it becomes pretty difficult for people to create broken mappings, and you can have the builders provide sensible defaults for all the properties, meaning you just need to override the ones you care about in any given test case. This also has the effect of de-coupling you further from the system you're integrating with because you only have to change your tests when the data you use changes.

PR #230, opened 2015-02-04
Fix binary data corruption when proxying requests

zregvart:
When proxying binary data, for instance Spring Remoting, byte[] to/from String conversion can corrupt data. This PR introduces getBody() method on the Request class and stores the raw bytes instead of bytes converted to String.
Also I found that writeAndTranslateExceptions does not flush/close the ServletOutputStream, so that was added.


tomakehurst:
Thanks for this. Been meaning to fix this problem for a while. I'll try and find some time to merge it over the weekend.

piersy:
Hi tom, any eta on this, it would be really useful?

tomakehurst:
I'm still looking at this. Doesn't help that the PR doesn't have any tests.

zregvart:
Hi, I've added an acceptance test illustrating the issue, had to resort to using com.sun.net.httpserver.HttpServer to make an impartial test.

tomakehurst:
Thanks for adding the test. Looks good.

PR #233, opened 2015-02-13
NullPointerException when query params provided in actual request are missing when compared with expectations

lucas1000001:
Hi, 

I've come across an issue where if I have a bunch of assertions such as:

            mock.register(
                    get(urlPathEqualTo("/a/b/c"))
                            .withQueryParam("x", equalTo("1"))
                            .withQueryParam("y", equalTo("2"))
                            .willReturn(
                                    aResponse()
                                            .withStatus(200)
                                            .withBody("")
                            )
            );
         
But then go on to put in a request such as :

            /a/b/c?x=1

Then I'd expect the case to not match. But because the query param "y" is missing from the actual request, I'm getting a NullPointerException from the expectations.

See the test RequestPatternMatchesUrlTest#noMatchOnMissingQueryParam for the specific case.

Hope that helps

Cheers

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
@@ -356,7 +356,7 @@ public class RequestPattern {
                 ValuePattern valuePattern = entry.getValue();
                 String key = entry.getKey();
                 QueryParameter queryParam = request.queryParameter(key);
-                boolean match = queryParam.hasValueMatching(valuePattern);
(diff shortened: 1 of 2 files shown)


claudiusb:
Hi, I noticed this also and would like this fix to be included.

tomakehurst:
I've fixed this in commit 026ff487e244d2c10c604ef6ddd0b5329f075868. I didn't use this PR because I'm trying to avoid using null where possible, preferring a null value pattern.

lucas1000001:
OK great thanks, the bug being fixed is the main thing :)

PR #236, opened 2015-02-20
Increasing the request header buffer size.

casanova2020:
Hi,
this is to increase the request header buffer size.
It was doubled to the existing value.


tomakehurst:
A few points:
Could you explain the rationale for doing this?

There seems to be another commit in there doing something else. Could you remove this and create a separate PR if it's something you want to keep?

This is causing the build to fail for JDK 8 (see the Travis page).

tomakehurst:
Closing, as this has been added in another PR.

PR #246, opened 2015-03-06
CLI flag to set Jetty's request header buffer size

tricker:
My apps have very large Headers which were overflowing Jetty's
Request Header Buffer,  this resulted in response: 413 FULL head

To recreate the problem do any GET with a single header larger
than 9k or so.

Also added console output confirmation when Jetty related command
line flags are specified (there are 3 now total).  They previously
weren't output even when applied which leads to confusion.


tricker:
Will submit a separate pull request that includes an acceptance test and is synced with 1.54.

PR #248, opened 2015-03-09
CLI flag to set Jetty's request header buffer size

tricker:
My apps have very large Headers which were overflowing Jetty's
Request Header Buffer, this resulted in response: 413 FULL head

This patch adds the ability to specify a Jetty Header Buffer
Size as a CLI flag when starting standalone Wiremock.

To recreate the problem do any GET with a single header larger
than 9k or so.  I added an acceptance test that demonstrates the
problem.

Also added console output confirmation when Jetty related command
line flags are specified (there are 3 now total). They previously
weren't output even when applied which leads to confusion.

### DIFF
--- a/docs/source/running-standalone.rst
+++ b/docs/source/running-standalone.rst
@@ -87,6 +87,9 @@ The number of threads Jetty uses for accepting requests.
 ``--jetty-accept-queue-size``:
 The Jetty queue size for accepted requests.
 
+``--jetty-header-buffer-size``:
+The Jetty buffer size for request headers, e.g. ``--jetty-header-buffer-size 16384``, defaults to 8192K.
+
 ``--extensions``:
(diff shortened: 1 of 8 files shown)


tomakehurst:
Do you think there's any downside to having a permanently large header buffer, rather than it needing to be configurable?

tricker:
I don't think there is a downside to a large header buffer by default.  However since there is no max header size specified by the HTTP spec, it would be difficult to choose an appropriate size that covers all of the wiremock use cases out there.   To put it another way, whatever size we choose someone will come along who uses an even larger Header.

I do think the default size should be bumped up regardless.  My particular needs are 32k+

tomakehurst:
Yeah, I guess that's true. I've been bitten by this at least once in the past.

PR #255, opened 2015-03-31
Add capability to set up a namespace context so that xpaths with namespa...

wdebeau1:
...ces can be supported without falling back on local-name()

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ValueMatchingStrategy.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ValueMatchingStrategy.java
@@ -19,12 +19,16 @@ import com.github.tomakehurst.wiremock.matching.ValuePattern;
 import com.google.common.base.Function;
 import org.skyscreamer.jsonassert.JSONCompareMode;
 
+import java.util.HashMap;
+import java.util.Map;
+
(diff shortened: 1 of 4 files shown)


tomakehurst:
Please could you make a couple of changes:
1) Restore the indentation in the files changed in the first commit, as I'm having a hard time distinguishing the significantly changed lines from the reformatted ones, and
2) Remove the ``assertEquals`` call. A better solution would be to add a message to the ``assertThat`` call IMO.

wdebeau1:
Sorry about the whitespace stuff, it should be fixed now.

As to the test...
I had changed the test to assertEquals to try to see what status code I was getting back, since assertThat doesn't tell me. I was never able to reproduce the failure locally, and the code I was working on shouldn't have had any impact on that test (StandaloneAcceptanceTest.matchesVeryLongHeader()). It was only failing in one of the 1.7 builds, and when I added the assertEquals, the test started passing, so I was never able to figure out why it was failing in the first place. 
In any case, it should be back to its original form and it passed this time.

davidnewcomb:
Is it pulled in yet?

tomakehurst:
Merged as a new commit. Thanks for this, and sorry for the wait!

PR #265, opened 2015-04-24
Include trailing port information in host header when proxying.

nchaugen:
According to the HTTP/1.1 standard, the "host" header should include trailing port information. If left out, it implies the default port for the service requested (http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html#sec14.23).

The changes in this pull request make WireMock adhere to this by always including trailing port information in the "host" header when proxying.


tomakehurst:
Thanks, looks good!

PR #276, opened 2015-05-25
Modularized to get rid of dependency conflicts on android. Removed chunked mode.

krzysiekbielicki:
On Android I had problem with chunked mode in which wiremock transfers stub responses.
Retrofit wit all clients (apache, okhttp, httpurlconnection) throws SocketTimeoutException.

Also android has a problem with using WireMock as admin client to configure stubs from instrumentation tests.

In this PR contentLength in set for stub responses and code is splitted into two gradle modules:
- client - all client code without apache http-client dependency (AdminClient interface has to be implemented in test code, sample implementation https://github.com/krzysiekbielicki/wiremock-android-sample)
- server - all code and dependecies not needed on ondroid when using client-only mode.

Only differnce between this and upstream implementation is that Wiremock.defaultInstance is not statically initialized (it was needed to get rid of apache HttpClient which is cause of class conflict on android).


tomakehurst:
Apologies for taking so long to review this. Unfortunately, I don't think I can merge it as it's a fairly fundamental design (and therefore breaking) change.

As I mentioned in the README, it's usually a good idea to raise a feature request or start a thread on the mailing list if you're thinking of submitting a non-trivial PR.

I'd like to support the Android community better (and I've been saying this for well over a year admittedly), but how that's achieved needs some more thinking IMO.

My (limited) take on it is that it should be possible to run both client and server on Android with a bit of work.

jakubkrolewski:
Is there any update on Android support? As I saw in the changelog for version 2.0, the support for Java 6 and 7 was dropped. Does it mean that the support for Android, which can't handle Java 8, was dropped as well?

tomakehurst:
Java 7 support has been restored, but not Java 6.

No other progress on the Android front, but I've been giving it some thought recently. Would you find it useful to have WireMock installable as an app?

PR #280, opened 2015-06-05
allow using single quotes for quoting Strings in JSON files

kamilszymanski:
This PR allows you to use single quotes so that you don't have to escape double quotes (e.g. for JSON response body)


tomakehurst:
Thanks!

Issue #90, opened 2014-01-13
Race condition in recording


gustavonalle:
It seems like that sending two almost simultaneous identical requests to the recording standalone server, neither of them is persisted to disk.

tomakehurst:
New recorder coming in #674

PR #281, opened 2015-06-05
Test and provisional fix for issue #90: Race condition in recording

twystd:
Issue #90 is not a true race condition - it happens when multiple 'similar' requests are waiting for a response from a slow server. StubMappingJsonRecorder checks against a list of already received requests (rather than the recorded requests/response stubs) so discards the delayed response, assuming that it has already been received and processed.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
@@ -67,15 +67,16 @@ public class StubMappingJsonRecorder implements RequestListener {
 //    }
 
     /* 
-     * Updated <code>requestReceived</code> to correctly handle multiple concurrent identical 
-     * requests. The initial implementation discarded any/all responses if a pending request
-     * matched the request/response.
+     * Provisional fix for issue #90 to correctly handle multiple concurrent identical 
+     * requests by checking against a list of recorded requests rather than a list of
+     * received requests.
      * 
-     * And 'synchronized' ensures that only one of the request/response pairs is recorded 
-     * for the case where responses are received 'simultaneously' (i.e. within the execution
-     * time of this method).
+     * 'synchronized' is required to ensure that only one of the request/response pairs is
+     * recorded for the case where responses are received 'simultaneously' (i.e. within the
+     * execution time of this method).
      * 
      * @see com.github.tomakehurst.wiremock.ConcurrentDelayedResponsesFixTest
+     *
      */
     @Override
     public synchronized void requestReceived(Request request, Response response) {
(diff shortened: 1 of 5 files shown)


tomakehurst:
Hi,

Apologies for taking so long to review this.

A few comments:
1) You've marked the fix as provisional. What in your view still needs to happen to make it final?
2) I think there's probably a better way to support concurrency in the stub recorder. Perhaps rather than synchronizing ``requestReceived()``, you might look at using something like ``java.util.concurrent.CopyOnWriteArrayList`` as the recorded list implementation?
3) Rather than cloning a tweaking ``WireMockServer``, ``HttpClientFactory`` and ``ProxyResponseRenderer``, maybe subclassing (with the necessary extension points exposed) would be better? Or even just adding the connection manager into the main client factory?
4) The test case is pretty complex. I suspect you could simplify this by making the same request from two threads simultaneously, configuring the target service to respond very slowly to the first request then quickly to the second.

twystd:
Hi Tom,

No problem :-)

1. As far as I'm concerned it just needs somebody who knows the codebase to review it to be sure there weren't side-effects - it's a fairly subtle behaviour change so I wasn't 100% sure. I've been using it for a couple of months and haven't noticed anything but that doesn't prove anything.

2. The 'synchronized' is there because that whole section of the code needs atomic compare-and-set semantics  to avoid writing away two stubs for the 'same' request (I actually had that show up running the tests prior to putting in the synchronized). If writing duplicate stubs doesn't matter I can update the unit tests and then a CopyOnWriteArray will do it - otherwise I could look at moving the synchronized to a guard variable or see if I can get the same semantics some other way if you prefer ?

3. I tried subclassing WiremockServer etc, but adding in the connection manager functionality involved sweeping changes to the core code and I was reluctant to do that just to get a test case to run. Am totally happy to add the connection manager into the core code if you think it will be useful - maybe it should be a separate issue/pull request, though ?

4. Uh, duh - let me try that (the test case mirrors my real-world setup)

tomakehurst:
2. I generally try to avoid using low-level concurrency primitives. There's pretty much always something in java.util.concurrent that suits what I'm trying to do. Doing your own synchronisation can be tricky, so I prefer to use more abstract building blocks that fit my use case. It's a fair point about duplicates, perhaps one of the other concurrent collections in the JDK e.g. ``ConcurrentSkipListSet``?

3. Looking at the internals of ``HttpClientBuilder``, it seems that it already creates a connection pool. Are you absolutely certain it's forcing a single connection without your changes?

twystd:
1. No problem - I'll restructure it to avoid the 'synchronized'. There are lots of ways to do the same thing, using 'synchronized' just seemed the simplest :-)

2. Yeah, HttpClientBuilder does use a connection pool but it leaves the max connections at the default of 2 - I found this out the hard way when I found the requests being mysteriously serialized and not really testing the concurrency at all. It's a bit hazy now, but I think the 'synchronized' issue only showed up intermittently at first and increasing the number of concurrent requests made it easier to verify the fix.

Digging around in the HttpClientBuilder code yesterday, I found this:

            if (systemProperties) {
                String s = System.getProperty("http.keepAlive", "true");
                if ("true".equalsIgnoreCase(s)) {
                    s = System.getProperty("http.maxConnections", "5");
                    final int max = Integer.parseInt(s);
                    poolingmgr.setDefaultMaxPerRoute(max);
                    poolingmgr.setMaxTotal(2 * max);
                }

so it may be possible to increase the max connections using system properties. I seem to vaguely remember trying it (unsuccessfully) but I guess let me play a bit.

tomakehurst:
New recording coming in #674

twystd:
Whew :-)

PR #283, opened 2015-06-12
allow specifying response body as JSON

mlc:
Sometimes, when writing mappings as JSON, you want to mock a JSON response body. Having to escape the JSON inside the JSON can be a bit obnoxious, so this patch lets you specify a JSON object instead.


tomakehurst:
Thanks. Enjoyed the supporting meme :-)

PR #294, opened 2015-07-30
WireMock to support parallel unit testing

ashley-taylor:
Changed the WireMock class to use a thread local variable instead of a static variable.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -39,7 +39,12 @@ public class WireMock {
 
 	private final Admin admin;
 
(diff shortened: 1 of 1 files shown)


tomakehurst:
Can you elaborate why you'd want to have a client per thread? Not sure I can see the benefit when they're all pointing at the same server.

ashley-taylor:
when you have two or more test classes that contain a WireMockRule and you execute your tests in parallel.

The WireMockRule will call the static configureFor method in the WireMock class.

This causes the stubbing and verifying to happening to a single wiremock server from all testing threads.

tomakehurst:
OK, that's a fair point.

Incidentally, you can avoid this problem if you use the ``Stubbing`` interface on the rule, and you'll get a bit of extra performance as it'll configure via a direct call rather than over the wire.

Issue #296, opened 2015-08-31
Consider downgrading to Jetty 9.2


jkubrynski:
Jetty 9.3 is known to cause problems with Jersey and it requires Java 8, which causes many problems with projects depending on WireMock

marcingrzejszczak:
This includes this ticket: https://java.net/jira/browse/JERSEY-2940

tomakehurst:
Java 7 went EOL 5 months ago,  so deprecating support for it doesn't seem unreasonable at this point. Keeping it just means having to stick with more obsolete libs e.g. Jetty.

Can you elaborate why Jersey compatibility is a concern (given that it's always possible to use the standalone jar as a dependency if want to avoid wiremock imposing its version on your classpath)? 

Also it looks like they're working on fixing it. I'd be surprised if they remained incompatible for long.

jkubrynski:
Java 7 is EOL, but there is still a lot of projects which use it. 

The problem with Jersey is that if you use Jersey in your application it gets break down in tests due to wiremock dependency. Which means wiremock 2.0 doesn't work in projects using Jersey.

tomakehurst:
That's true, but if everyone took that line it'd have to be supported in perpetuity. Someone has to be first (or at least not last) to drop support. I'm guessing you have a problem with it because you'd like to keep accurest at JDK 7? Is there a specific reason you need to do this?

I'm still not clear why using the standalone dependency isn't a solution to the Jetty/Jersey problem. Have you found this doesn't work for you?

marcingrzejszczak:
@tomakehurst in micro-infra-spring we're starting in memory zookeeper with in memory wiremocks that we feed with stubs downloaded from nexus. We need that to be available at runtime.

jkubrynski:
There are some people using accurest with jdk 7. We dont want to drop
compatibility for now.

I'm not sure what you mean with standalone jar. Usually you have wiremock
in your dependencies, which causes using jetty 9.3. In fact we re even not
able to exclude jetty-server,  as wiremock uses classes introduced in 9.3.

tomakehurst:
So what criteria would you use to decide when it's time to drop support for a JDK, if not EOL?

I mean adding the standalone classifier to your wiremock dependency, then excluding all its transitive dependencies as documented on the site. This will get easier shortly as I'm planning to start releasing the standalone version under a different artifact ID with a minimal POM.

[...]

PR #298, opened 2015-09-03
[#296] Fix missing depndencies in pom.xml

szpak:
Standard pom.xml has all requied dependencies. `packaging` in standalone pom.xml is `jar` instead of `pom`.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -217,17 +217,27 @@ publishing {
         }
     }
     publications {
-        thinJar(MavenPublication) {
-            artifact jar
+        mavenJava(MavenPublication) {
+            from components.java
+            artifact sourcesJar
+            artifact javadocJar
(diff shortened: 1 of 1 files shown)


tomakehurst:
I've just found a terser way of fixing it. Check out the latest on the 2.0.0-beta branch.

szpak:
Unfortunately your version seems to not work correctly. does not work
1. There are no source and javadoc jars.
2. The scope is `runtime` for the dependencies - shuld be `compile`.
3. `packaging` in standalone pom.xml is `pom` - should be `jar`.

These issues are not available in my version. You may consider backporting solution from my branch.

tomakehurst:
I've finally got this working. There's now a 2.0.1-beta release on Sonatype with separate artifacts and correct POMs for each.

PR #304, opened 2015-09-05
In order to allow for observing the matching process I have extracted…

wojciechbulaty:
… the matching behaviour from RequestPattern to a new class RequestPatternMatcher and  introduced a MatcherObserver.

Similar to https://github.com/tomakehurst/wiremock/pull/303 but trying to avoid breaking changes.


tomakehurst:
I'm not going to merge this as we're not taking PRs for the 1.x version any more. Details are in the README:
https://github.com/tomakehurst/wiremock/blob/master/README.md

I generally agree that diagnosing non-matches could be easier, but I'm not sure how this particular solution gives substantially more than you currently get with verbose logging turned on.

Can I suggest starting a mailing list thread to discuss options first, before diving into code? I've got a few ideas that I haven't got round to trying that might help here too, so it'd be good to get it all out on the table before committing to a specific solution.

wojciechbulaty:
Thanks Tom from the prompt reply.

I was hoping to have support for Java 7, therefore the merge to 1.x, but I can see now you have restored Java 7 in 2.0.2-beta, so I am cool now :)

I will have a look at that verbose logging, I did not know there was such a thing. I will also raise another discussion on the mailing list.

I hope I am not causing too much trouble :)
Thank for your help!

PR #306, opened 2015-09-14
Add edit mapping api and related changes

dlaha21:
This adds edit mapping functionality and some related changes

* Adds UUID to `StubMapping` to uniquely identify one
* Adds `mappingFileName` to `StubMapping` to persist the edited mapping to the original file
* Fixes the Comparator for the `ConcurrentSkipListSet`. It was not possible to remove anything from the set as the comparator never returned "0"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
 language: java
 jdk:
   - oraclejdk8
+  - oraclejdk7
+  - openjdk7
--- a/build.gradle
+++ b/build.gradle
@@ -26,12 +26,13 @@ apply plugin: 'project-report'
 apply plugin: 'com.github.johnrengelman.shadow'
 
 sourceCompatibility = 1.7
+targetCompatibility = 1.7
 group = 'com.github.tomakehurst'
-version = '2.0.1-beta'
+version = '2.0.2-beta'
 
 def versions = [
         jackson: '2.6.1',
-        jetty  : '9.3.2.v20150730'
+        jetty  : '9.2.8.v20150217'
 ]
 
 repositories {
@@ -318,7 +319,7 @@ task 'bump-version' << {
             'docs/source/getting-started.rst'   : [{ "<version>2.0.${it}-beta</version>" }, {
                 "wiremock/2.0.${it}-beta/wiremock-2.0.${it}-beta-standalone.jar"
             }],
-            'docs/source/running-standalone.rst': { "wiremock/2.0.${it}-beta/wiremock-2.0.${it}-beta-standalone.jar" },
+            'docs/source/running-standalone.rst': { "wiremock/2.0.${it}-beta/wiremock-standalone-2.0.${it}-beta.jar" },
             'sample-war/build.gradle'           : { "com.github.tomakehurst:wiremock:2.0.${it}-beta" },
     ]
 
--- a/build.gradle
+++ b/build.gradle
@@ -56,7 +56,6 @@ dependencies {
     compile("junit:junit:4.12") {
         exclude group: "org.hamcrest", module: "hamcrest-core"
     }
-    compile "org.apache.commons:commons-lang3:3.4"
 
     testCompile "org.hamcrest:hamcrest-all:1.3"
     testCompile("org.jmock:jmock:2.5.1") {
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ apply plugin: 'com.github.johnrengelman.shadow'
 sourceCompatibility = 1.7
 targetCompatibility = 1.7
 group = 'com.github.tomakehurst'
-version = '2.0.2-beta'
+version = '2.0.5-beta'
 
 def versions = [
         jackson: '2.6.1',
--- a/build.gradle
+++ b/build.gradle
@@ -55,6 +55,7 @@ dependencies {
     compile("junit:junit:4.12") {
         exclude group: "org.hamcrest", module: "hamcrest-core"
     }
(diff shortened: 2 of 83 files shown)


tomakehurst:
Hi, apologies for the delay reviewing this.

A few comments:
1) There are no acceptance tests for this. I generally find the acceptance tests to have the most value over the long term as they're a lot more amenable to refactoring while still running pretty fast. Also, acceptance testing via the HTTP admin interface has the effect of asserting that the JSON API is OK.
2) There are some changes I'm having trouble reading because it looks like your IDE has changed the tabs. Hard to distinguish genuine changes vs. formatting.
3) I'd prefer not to add commons lang if at all possible. I realise it's already there transitively, but I'd prefer not to depend on any additional libs unless they're adding significant value. I usually find there's something close enough in Guava when considering commons packages, and the code quality is generally better.
4) I need to think about this a bit more, but I don't think ``StubMapping`` should have any knowledge of mapping files. Seems like a coupling in the wrong place. Would you say the file saving part is essential to this change? Could it be pulled out into a separate PR?
4.5) A better option here might be use the UUID of mapping as the file name (or part of it). That way you wouldn't need to maintain the relationship explicitly at all.
5) Was there any intention to be able to remove stub mappings? Since there's no admin task class for this, I can't see how this could be done via the API.

Other than that I think you're on the right track. I'll try and have a think about saving mappings and how that might be better abstracted from the core domain. It might be nice to try and decouple it from the underlying storage mechanism i.e. put it behind a ``StubRepository`` interface, with a default file-backed impl, but with the future option to plug in databases etc.

dlaha21:
Hey, no worries for the delay. Please find my replies to the comment below. 

1. Yes, I'll add acceptance tests for this, I was actually not aware of them
2.  As Wiremocks has a mix of tab and space separated formatting, it becomes very difficult to configure the project formatting (in the IDE) for these cases. For example `AcceptanceTestBase` has both tabs and spaces. I'll fix the formatting changes, but it would greatly help if this was consistent throughout the project.
3. That's fine, I'll change that.
4. I agree, it felt wrong in putting the fileName in StubMapping. The save mapping would be needed by the UI, but I can pull it out in a separate PR. I did think of using the UUID as the file name, but then it would not work for users with an existing set of mapping files or if someone generated the files through other means. I like your idea of a `StubRepository` and then the default file-backed one can keep a map of the uuid vs the filename. While in the case of a future DB based repo, the UUID itself can be used as the key. If you are ok with this, I'll do this as a part of a different PR and pull out the file saving changes from this one. Let me know...
5. Yes, but I though I would raise it as a separate PR. Did not want this one to get too big

tomakehurst:
Agree the formatting is a bit of a mess (that's 4 years of PRs for you!). I'll try and make that consistent going forward and possibly fail the build if it gets broken.

dlaha21:
Hey, I have made the changes and pulled out file name from StubMapping. I will try and create a separated PR for that feature. Let me know what you think.

dlaha21:
Hey Tom, did you get a chance to look at the updated PR? I have made the changes that you mentioned.

tomakehurst:
Apologies, I've been struggling to find a long enough block of time recently. I'll pull it down and try and review it on the train (if I can get a seat).

tomakehurst:
OK, this is generally looking very good.

Just a couple of remaining points:
1) The acceptance test isn't really end-to-end in the way that the others are - i.e. it's hitting the JSON API directly rather than via the DSL, so the DSL isn't being exercised.
2) I'm dubious about the ``VisibleForTesting`` annotation on ``StubMapping`` - the setter needs to be present for JSON binding to work.

dlaha21:
Find my replies below

1. I don't quite understand this. The edit API doesn't have a corresponding DSL. Do you mean that we setup the mock using a DSL, then edit it through the API and assert that it has been modified as expected? But then this would be testing the API for which we already have a test
2. You are right, will remove that

tomakehurst:
In that case could you add editing to the DSL? Currently there's nothing that you can only do via the JSON API, and I think it would be good to keep it that way.

dlaha21:
Sorry for the late reply, was on vacation. imho the edit API does not fit well into the DSL as you would need to have the UUID to uniquely identify the mapping. 
If we want to do this then the add mapping DSL would need to return the UUID and `WireMock#stubFor`, `WireMock#givenThat` and `WireMock#register`'s return type would change from void to String. What do you think?

tomakehurst:
The alternative (which I started working on in a spare moment) is to specify the ID yourself both on creating and editing stubs.

e.g.
```java
UUID id = UUID.randomUUID();

wireMockServer.stubFor(get(urlEqualTo("/edit-this"))
                .withId(id)
                .willReturn(aResponse()
                        .withBody("Original")));

wireMockServer.editStub(get(urlEqualTo("/edit-this"))
                .withId(id)
                .willReturn(aResponse()
                        .withBody("Modified")));
```

Can you see any issues with this?

dlaha21:
This looks good... I will add these in. Just one point, since its a UUID, I am more inclined to have
the method as `withUuid()` than `withId()`. Let me know your thoughts...

tomakehurst:
I marginally prefer ``withId`` because the fact it's a uuid is made explicit by the parameter type, but I don't feel that strongly about it.

tomakehurst:
Hey, I've added edit to the DSL now and pushed it to the 2.0-beta branch. I ended up rebasing, hence the fact your individual commits aren't there. Please take a look and let me know if I've missed anything.

PR #316, opened 2015-10-16
Implemented equals for HttpHeader

adrianmkng:
I've implemented equals for HttpHeader as promised.


tomakehurst:
I've pushed a release with this.

2.0.6-beta will have it next time Maven syncs.

PR #320, opened 2015-10-29
Enable HttpAdminClient to be used for standalone process wiremock server

sixtysecond:
Need: ability to write tests for standalone wiremock using same syntax as for in-process wiremock.
Changes:
* Add support for https
* overload addStubMapping to also accept MappingBuilder
* overload countRequestsMatching to also accept MappingBuilder.

After these changes, or functionally similar ones have been merged to master, I will update the documentation to show how to use HttpAdminClient for remote testing.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
@@ -40,18 +40,18 @@ import static org.apache.http.entity.ContentType.APPLICATION_JSON;
 
 public class HttpAdminClient implements Admin {
 
-    private static final String ADMIN_URL_PREFIX = "%s://%s:%d%s/__admin";
+	private static final String ADMIN_URL_PREFIX = "%s://%s:%d%s/__admin";
 
-    private final String scheme;
-    private final String host;
-    private final int port;
-    private final String urlPathPrefix;
+	private final String scheme;
+	private final String host;
+	private final int port;
+	private final String urlPathPrefix;
 
-    private final HttpClient httpClient;
+	private final HttpClient httpClient;
 
-    protected String getScheme() {
+	protected String getScheme() {
         return scheme;
-    }
+	}
 
     protected String getHost() {
         return host;
@@ -94,13 +94,13 @@ public class HttpAdminClient implements Admin {
         this(host, port, "");
     }
 
-    @Override
-    public void addStubMapping(StubMapping stubMapping) {
+	@Override
+	public void addStubMapping(StubMapping stubMapping) {
         postJsonAssertOkAndReturnBody(
                 urlFor(NewStubMappingTask.class),
                 Json.write(stubMapping),
                 HTTP_CREATED);
-    }
+	}
 
     public void addStubMapping( MappingBuilder mappingBuilder) {
         addStubMapping(mappingBuilder.build());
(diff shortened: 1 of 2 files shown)


tomakehurst:
Agree HTTPS support for admin is definitely a good idea!

Could you elaborate the use cases for being able to pass ``MappingBuilders`` into those two additional overloads?

As per the readme, the master branch (i.e. v1.x) is frozen at the moment, and I'm only merging PRs for the 2.0-beta branch. Please could you re-issue the PR against that branch?

Also, the test case diverges a bit stylistically from the existing test suite. Could I suggest ditching it in favour of an acceptance test in style of the ones already there?

sixtysecond:
> As per the readme, the master branch (i.e. v1.x) is frozen at the moment, and I'm only merging PRs for the 2.0-beta branch. Please could you re-issue the PR against that branch?

I will do so.

> Also, the test case diverges a bit stylistically from the existing test suite. Could I suggest ditching it in favour of an acceptance test in style of the ones already there?

I will review the existing tests for style, but more specific feedback would be greatly appreciated.

> Could you elaborate the use cases for being able to pass MappingBuilders into those two additional overloads?

The goal is to be able to use the same syntax for stubbing using the admin client as when using stubbing and verification. Therefore code written 

Without the change, the syntax has an extraneous build method which differs from the in-memory syntax:
client.addStubMapping(get(urlEqualTo("/some/thing")).willReturn(aResponse().withHeader("Content-Type", "text/plain").withBody("Hello world!")).build());

with the change:
client.addStubMapping(get(urlEqualTo("/some/thing")).willReturn(aResponse().withHeader("Content-Type", "text/plain").withBody("Hello world!")));

in-memory for comparison:
 stubFor(get(urlEqualTo("/some/thing")).willReturn(aResponse().withHeader("Content-Type", "text/plain").withBody("Hello world!")));

An alternative to overloading would be for both the MappingBuilder and StubMapping to implement a common interface, but this seemed like a more minimal change.

sixtysecond:
Closing pull request since resubmitted with requested changes on beta branch without overloading stub and verify methods with builders.

tomakehurst:
Is ``client.register(get(urlEqualTo(...)))`` no good for what you're attempting?

sixtysecond:
I assume you're referring to using the WireMock class, including register.
The static default instance and static methods seem a confusing interface when you're dealing with remote management. While I can appreciate the succinctness of static methods for fluid dsl, in this case it seems that the initial method should be an instance method belonging to a class with only instance methods to make it clear which wiremock is being communicated with.

One approach would be to implement a remote wiremock class which mimics much of the behavior and interface of the WireMock class but with no static state or methods. In this way, all of the example code in the documentation could be used for remote management with the minor addition of needing to prepend the call with the instance name, e.g. remoteWireMock.stubFor(.....)

tomakehurst:
Yeah, I agree it is a bit confusing having the static and instance methods named differently. Having them on the same class wasn't a great choice but would've been quite hard to change once people started using it without causing some pain.

Perhaps it would be a good idea to use the beta to deprecate that class and split them out for consistent naming. Adding more methods feels a bit like a workaround TBH.

sixtysecond:
I completely agree. How about if WireMock implements a new interface, e.g. WireMockClient. WireMock remains the class to use for localhost testing. Then another class implements the same interface for remote testing, but with constructors appropriate for remote testing.

PR #321, opened 2015-11-03
Added https support and builder for HttpAdminClient

sixtysecond:
Resubmitting change on 2.0 beta branch as requested.
Added support for https for admin client.
Added builder to allow  selecting http or https scheme and custom http client.


sixtysecond:
An alternative to this approach would be to have the HttpAdminClient take the complete url as a parameter so that scheme becomes a non-issue. This would probably be cleaner than requiring people to set an optional builder parameter for https.

PR #331, opened 2015-12-08
Support randomly distributed delays for responses

jsravn:
This expands on the existing fixed delay functionality by providing a
more general feature to draw values from a random distribution.

This allows teams to simulate long tails in delays from downstream
systems, a common scenario in load and resiliency tests.

I've provided a log normal implementation out of the box, which gives a
pretty good long tail distribution for modeling delays. It also has the
nice benefit of being able to specify the 50th percentile which is a
commonly tracked metric.

Here is an example priming using the log normal delay:

```json
{
    "request": {
            "method": "GET",
            "url": "/some/thing"
    },
    "response": {
            "status": 200,
            "body": "Hello world!",
            "headers": {
                    "Content-Type": "text/plain"
            },
            "delayDistribution": {
                    "type": "lognormal",
                    "median": 80,
                    "sigma": 0.4
            }

    }
}
```

This example can be viewed visually at wolfram alpha:
https://www.wolframalpha.com/input/?i=lognormaldistribution%28log%2890%29%2C+0.4%29

New distributions can be added by implementing `DelayDistribution` and
adding it to the @JsonTypeInfo in the interface.

### DIFF
--- a/docs/source/simulating-faults.rst
+++ b/docs/source/simulating-faults.rst
@@ -128,6 +128,25 @@ To use, instantiate a ``new LogNormal(median, sigma)``, or via JSON:
             "sigma": 0.4
     }
 
+Uniform delay
+^^^^^^^^^^^^^
+
+A uniform distribution can be used for simulating a stable latency with a fixed amount of jitter. It takes two parameters:
+
+* lower - Lower bound of the range, inclusive.
+* upper - Upper bound of the range, inclusive.
+
+For instance, to simulate a stable latency of 20ms +/- 5ms, use lower = 15 and upper = 25.
+
+To use, instantiate a ``new UniformDistribution(15, 25)``, or via JSON:
+
+.. code-block:: javascript
+
+    "delayDistribution": {
+            "type": "uniform",
+            "lower": 15,
+            "upper": 25
+    }
 
 .. _simulating-faults-bad-responses:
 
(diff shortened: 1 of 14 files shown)


jsravn:
Just some background, this is a 2.0 implementation of behaviour I wrote for a custom fork of an old version of wiremock, used to load test a relatively large set of backend microservices at 5-10k tps, and specific failure scenarios. I've been meaning to try and get this in upstream so we can get people on 2.0 when it comes out. So here you go :).

tomakehurst:
James - this is awesome, I've been meaning to add this feature for a very long time!

Would you mind adding a small section to the docs explaining how to use this?

Thanks,
Tom

jsravn:
Hi Tom, I've added documentation, feel free to edit it as you wish.

Although doing the docs made me realize I should probably add this to the global config as well, just like fixed delay. And it may be worth adding a couple more distributions (at least fixed) for illustrative purposes. I'll try to work on that later.

tomakehurst:
Thanks again, I'll merge what you've done.

I was thinking I'd add a uniform distribution at least. I've heard that Weibull distributions are good models for some situations, so that might be worth considering also.

jsravn:
Weibull would be good, you can probably model longer tails better than a lognormal. The main reason we ended up using lognormal is the nice property that one of its parameters is the 50th percentile, easy to gather from existing metrics. Then just a single parameter needs to be tuned.

PR #339, opened 2016-01-01
2.0 scenario builder generics

rowanhill:
As discussed, this is just a reworking of #103 but on the 2.0-beta branch, now we can make some breaking changes.

A few small differences from the original PR:
- I've added a bunch of `@SuppressWarnings("unchecked")` annotations where I have to cast to (T)
- MappingBuilder is abstract; we always want to create a ScenarioMappingBuilder, even if the type system only thinks it's a MappingBuilder...
- I've removed the check in `build()` for required/new state without scenario name, and instead thrown an exception when trying to set a null scenario name (as that's now the only way you can end up with states without a scenario name, thanks to the type system)
- I've made the same change to LocalMappingBuilder, although I'm not entirely sure what it's used for - I couldn't find any tests for it, or any uses of `requestMatching(RequestMatcher)` in the code base. I'm a bit confused by it in general, as it only seems to forward calls to its internal MappingBuilder (so why not just delete it and use MappingBuilder?), but I figure it should behave in the same way as MappingBuilder.

Any thoughts?


tomakehurst:
Thanks for this. While I was looking over it I had a bit of a brainwave as to how to avoid a class hierarchy explosion.

The ``LocalMappingBuilder`` class exists so that the compiler will prevent you trying to register custom matchers (i.e. code that the server needs to execute) when you're talking to a remote ``WireMock`` instance. As such it's a marker interface of sorts, and seeing what you've had to do here to accommodate ``ScenarioMappingBuilder`` made me realise it's not a great solution.

So I've pulled all the impl (including for scenarios) back into ``MappingBuilder`` and created a set of interfaces pretty much corresponding to the classes in your PR that vary the returned interface.

For now I've pushed the result to this branch:
https://github.com/tomakehurst/wiremock/tree/rowanhill-2.0-scenario-builder-generics

I've knocked this up pretty hastily and it seems to work, but I'd definitely appreciate your feedback!

Thanks,
Tom

rowanhill:
Sorry, just spotted this, and don't have time to take a proper look just now.

I did try and do something with interfaces rather than class hierarchies, but I didn't try with _generic_ interfaces like you have. That meant it wasn't possible to do `.inScenario("foo").atPriority(1).willSetStateTo("bar")` or similar - i.e. all scenario related calls had to immediately follow inScenario, with nothing interrupting them. I think the use of generic types solve that, right?

I'll try and take a proper look through soon.

tomakehurst:
Yes, the I believe the behaviour is equivalent to what you'd done. 

No rush reviewing it. Suspect I'll be neck deep in client work from tomorrow so I won't be working on it much either.

rowanhill:
I've had a look through - it looks good!

I do have some suggestions though; I've just pushed them, and split them out as separate commits so you can easily pick and choose which you want.

Some brief notes on them:
- I think there was a genuine bug: LocalMappingBuilder should return LocalMappingBuilder, not RemoteMappingBuilder; I've added a compile-only "test" for this.
- Tidy up: SuppressWarnings annotations aren't needed in this brave new world of interfaces
- Tidy up: The scenario methods in MappingBuilder can return MappingBuilder, not the ScenarioMappingBuilder interface. I've made MappingBuilder package private (aka "default") to ensure the implementation type doesn't accidentally leak out anywhere.
- Minor improvement: ScenarioMappingBuilder doesn't need to be generic. It also doesn't need to redefine RemoteMappingBuilder's methods, since we're extending it, and will automagically inherit them.
- Slightly crazy improvement: Unfortunately, we can't do the same trick with LocalMappingBuilder, since then MappingBuilder implements two interfaces with irreconcilable types (the compiler doesn't know whether it returns a LocalMappingBuilder or a ScenarioMappingBuilder). It's annoying that we can't get some type safety to guarantee that LocalMappingBuilder will always have the same interface (other than return type) as RemoteMappingBuilder. To provide that guarantee, I've written a slightly hacky reflection based test. I suspect that the order of methods returned by reflection isn't guaranteed, so I guess there's a small chance it might break on other JVMs. It'd be possible to write something more intelligent, but it didn't seem worth the effort.

tomakehurst:
Thanks for this feedback, and apologies for not having reviewed it yet!

tomakehurst:
Merged, finally. Thanks again for your help with this!

PR #341, opened 2016-01-05
Upgrade to Jetty 9.3

slandelle:
Latest Jetty is 9.3.

Changes:
- Upgrade to JDK8. Jetty 9.3 is now built against JDK8, as cyphers used for HTTP/2.0 are not available in JDK7.
- `org.eclipse.jetty.server.Request#getUri` was renamed into `getHttpURI`
- `GzipFilter` no longer works. `GzipHandler` must be used instead.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -25,8 +25,8 @@ apply plugin: 'idea'
 apply plugin: 'project-report'
 apply plugin: 'com.github.johnrengelman.shadow'
 
-sourceCompatibility = 1.7
-targetCompatibility = 1.7
+sourceCompatibility = 1.8
+targetCompatibility = 1.8
 group = 'com.github.tomakehurst'
 version = '2.0.8-beta'
 
@@ -34,7 +34,7 @@ def shouldPublishLocally = System.getProperty('LOCAL_PUBLISH')
 
 def versions = [
         jackson: '2.6.1',
-        jetty  : '9.2.13.v20150730'
+        jetty  : '9.3.6.v20151106'
 ]
 
 repositories {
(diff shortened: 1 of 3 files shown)


tomakehurst:
We made a decision a couple of months back to hold off the upgrade to Jetty 9.3 in order to give folks a bit more time to get off JDK 7.

At the time there didn't seem to be any specific advantage in 9.3 over 9.2.

Since WireMock doesn't (yet) support HTTP/2, is there any urgency to this?

Can the JDK 8 cyphers be used with Jetty 9.2?

slandelle:
As the developper of [AsyncHttpClient](https://github.com/AsyncHttpClient/async-http-client), I really want to use the latest Jetty version. Every time I upgraded, like from 9.2 to 9.3, I could uncover some bugs on my side.

Currently, my integration test suite directly uses Jetty, but a contributor was suggesting using WireMock, so I was evaluating it once again.
I perfectly understand your position, but I think I'll keep on going with a tailor made solution so I'm not bound to any dependency's agenda.

Feel free to close this PR, or keep it around for future reference when you'll want to upgrade.

Regards

tomakehurst:
Right, I see.

Unfortunate that we can't make everyone happy at the moment. I wonder whether it'd be worth considering adding an extra build target that produces another WireMock artefact with Jetty 9.3?

slandelle:
> Unfortunate that we can't make everyone happy at the moment.

Perfectly understandable, no pro.

> I wonder whether it'd be worth considering adding an extra build target that produces another WireMock artefact with Jetty 9.3?

You'd have to isolate incompatible code, as Jetty 9.2 and 9.3 are not compatible (see my commit).
You can probably do that with Gradle, but I can't help there, sorry.

yogurtearl:
Upgrading to Jetty 9.3 would also drop Android support, since Android does not include JDK 8-ish classes.

tomakehurst:
We definitely won't upgrade exclusively to Jetty 9.3/JDK 8 yet. What I'm considering (if I can't find the time) is to split out the sources specific to the webserver and split the build so that an additional build target can be created supporting whatever the Jetty bleeding edge is at the time without harming backwards compatibility in main artefact.

PR #347, opened 2016-01-17
Add support for requiring absent query parameters

yinonavraham:
There are some use cases when we would like to verify a query parameter is not being sent.
This pull request adds this ability by adding a `withoutQueryParam` method to the `RequestPatternBuilder` (similar to `withoutHeader`).

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/RequestPatternBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/RequestPatternBuilder.java
@@ -39,6 +39,7 @@ public class RequestPatternBuilder {
 	private Map<String, ValueMatchingStrategy> headers = newLinkedHashMap();
     private Map<String, ValueMatchingStrategy> queryParameters = newLinkedHashMap();
     private Set<String> withoutHeaders = newHashSet();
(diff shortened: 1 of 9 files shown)


tomakehurst:
The reason I haven't merged this is that I'm in the middle of a major refactor of the matching code. Part of the objective of this is to make the matching code a lot more modular and composable, and matching on absent query params will become possible as an effect of this.

Unfortunately, as usual having to fit it around client work, it's taking a while.

tomakehurst:
Added in v2.

dzielnykompanion:
I met the need for withoutQueryParam method, but I don't see this feature in Wiremock, despite it should be added starting from v2.... 
May i know what is the status of it? If this idea has been forgotten, I can refresh it and create a new PR @tomakehurst

tomakehurst:
@dzielnykompanion `withoutQueryParam` is just syntactic sugar.

You can already specify that a query parameter must be absent as seen here:
https://github.com/wiremock/wiremock/blob/master/src/test/java/com/github/tomakehurst/wiremock/StubbingAcceptanceTest.java#L266

dzielnykompanion:
Yep, I missed that, thanks ! I was confused because I was expecting syntactic sugar similar as withoutHeader method

PR #348, opened 2016-01-19
Convenient method to load JSON mappings on classpath for JUnit

phonypianist:
Added classpathFileSource() method.


tomakehurst:
Please could you:
a) Add a test for this
b) Preserve the formatting of the class so that only the affected lines appear in the merge

phonypianist:
Sorry, I had missed the similar function has already implemented. So I cancel this pull request.

PR #351, opened 2016-01-23
jetty-mbeans

eostermueller:
Tom,

With this pr, if you add --jetty-mbeans to the command line, jconsole will show you the jetty MBeans.
I plan on using these for some performance testing.  Specifically, I want document how to monitor current vs. max concurrency.

Two other things about this pr:

1) I created a junit test (JettyMBeansTest.java) that correctly validates the functionality that I added.  Unfortunately, some manual test setup is required to run it.  I documented that setup in JettyMBeansTest#canFindJettyMBeans().
If you have any bright ideas on making my junit run without this manual setup, I would love to hear them.
I documented the obstacles I ran into in JettyMBeansTest#canFindJettyMBeans().

2) It looks like someone has made a few minor changes _after_ I created my clone/fork.  It seems like those changes (ExtensionLoader.java is one example) are included in my pr.

Let me know if you've got any questions.

--Erik

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,7 +28,7 @@ apply plugin: 'com.github.johnrengelman.shadow'
 sourceCompatibility = 1.7
 targetCompatibility = 1.7
 group = 'com.github.tomakehurst'
-version = '2.0.8-beta'
+version = '2.0.9-beta'
 
 def shouldPublishLocally = System.getProperty('LOCAL_PUBLISH')
 
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
@@ -21,7 +21,6 @@ import com.github.tomakehurst.wiremock.extension.Extension;
 import com.github.tomakehurst.wiremock.extension.ExtensionLoader;
 import com.github.tomakehurst.wiremock.http.CaseInsensitiveKey;
 import com.github.tomakehurst.wiremock.jetty9.JettyHttpServerFactory;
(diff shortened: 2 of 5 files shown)


tomakehurst:
Thanks for this. I'll have a think about the manual test problem before I merge it, as I've found that tests that require manual steps never end up being run at all!

tomakehurst:
I've been tinkering with this for a while, and I'm troubled by the fact that the JVM's JMX server seems to start (and therefore read the system properties) before you've had a chance to configure it programmatically. Not only is this hard to test, but it limits its usefulness generally, particularly to anyone working entirely programmatically e.g. via the JUnit rule.

A while back I was considering integrating [Dropwizard Metrics](https://dropwizard.github.io/metrics/3.1.0), which provides many more metrics of more types than Jetty's JMX does, and has more options for exporting them.

Do you think this would this solve the specific problem you have?

eostermueller:
There is more to dropwizard than I thought, thanks for the link.
Let me play around with that.  I'm writing a short 150-200 page java
performance book.  The idea is to show that java performance is not as
complicated as everyone makes it out to be, so gathering server side
metrics needs to be real easy.

I was expecting to use this to display data:
http://jmeter-plugins.org/wiki/JMXMon/

But with your suggestion, I think the following will work, graphing a data
item from an XML over HTTP response:
http://jmeter-plugins.org/wiki/PageDataExtractor/

....and just need to see if the right metrics are available.  The main ones
I need are:
* web container max threads
* web container current threads
* access time for getting thread from pool of web container threads
* access time for getting jdbc connection from a jdbc DataSource.

I'll get back to you, thanks Tom.
--Erik

PR #354, opened 2016-01-26
Add Dockerfile to run wiremock. Standalone mode

gabrielpjordao:
I've pushed it to Dockerhub (under gabrielpjordao/wiremock) as well. It would be interesting if you could push it with the "original" username, though.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,6 @@
+FROM java:7-jre
+
+ADD http://repo1.maven.org/maven2/com/github/tomakehurst/wiremock/1.57/wiremock-1.57-standalone.jar .
+
+EXPOSE 8080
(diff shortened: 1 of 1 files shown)


gabrielpjordao:
I've pushed it to Dockerhub (under gabrielpjordao/wiremock) as well. It would be interesting if you could push it with the "original" username, though.

tomakehurst:
I think a Docker configuration for WireMock is a great idea, but I'm hesitant about keeping one in with WireMock itself.

I think deployment wrappers (and generally things that yield new build targets) ought to have their own release cycle, the main reason being that they can then be freely maintained by someone expert in that technology. I'm not particularly Docker literate, hence my reluctance to fold maintenance of a Dockerfile into WireMock's core.

There are a couple of WireMock Docker projects out there already. It'd be really great to see one of these become the focus of attention and therefore be kept up to date.

PR #355, opened 2016-01-27
Now using a trust strategy which accepts all certificates instead of …

phxql:
…the TrustSelfSignedStrategy, which checks if the chain length is exactly 1.


tomakehurst:
Awesome, thanks for this.

PR #360, opened 2016-02-09
add content-length to response header

yaronyam:
Some HTTP clients read or validate the response according to the content-length header.
Although there are ways to workaround this by pre-calculating the content-length and adding it to a stubbed response or maybe customizing a ResponseTransformer - it's nicer to have it supported transparently

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/Response.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/Response.java
@@ -16,13 +16,12 @@
 package com.github.tomakehurst.wiremock.http;
 
 import com.google.common.base.Optional;
+import com.sun.deploy.net.HttpRequest;
 
 import java.nio.charset.Charset;
-import java.util.Arrays;
 
 import static com.github.tomakehurst.wiremock.http.HttpHeaders.noHeaders;
 import static com.google.common.base.Charsets.UTF_8;
-import static com.google.common.collect.Iterables.transform;
 import static java.net.HttpURLConnection.HTTP_NOT_FOUND;
 import static java.net.HttpURLConnection.HTTP_OK;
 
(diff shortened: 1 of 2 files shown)


Mahoney:
I think this needs to be approached slightly differently. As it stands it's not backwards compatible. A lot of people use wiremock for simulating wrong or unusual HTTP behaviour, including not sending the content-length header, or sending an inaccurate content length header, and this would break them.

I'd favour a new method on the stub response DSL - something like 'withCorrectContentLength()', defaulting to false but setting it to true will make the equivalent of the work you've done happen.

yaronyam:
I understand the concern. This PR implementation will not override the content-length if it was already set by a stubbed response definition, so there's no impact for the use case of stubbing with a wrong content-length intentionally. The other use case, of simulating the lack of content-length header for some negative testing will break, but that may not be a common usage.
You're suggestion to explicitly define such behavior on each stub response definition is a viable solution, but I'd like to stay DRY. How about adding a conf option such as --add-content-length ( or --set-content-length?!)  which defaults to false and is override-able by stub response definitions? I guess that in such case the implementation would fit better as a Filter (such as ContentTypeSettingFilter)

yaronyam:
On second thought, perhaps a global header transformer, plugable as an extension, would be more appropriate

tomakehurst:
I think the problem here is that WireMock defaults to chunked encoding, and Content-Length isn't valid in this context (since you don't know the content length in advance when returning a chunked response).

Or are you seeing it return a non-chunked response without content-length?

yaronyam:
Yes, WireMock does return "Transfer-Encoding: chunked" header. The receiving end in my use case insists on getting a Content-Length. Would be nice if I could configure WireMock to return Content-Length instead of Transfer-Encoding.

tomakehurst:
I agree it'd be good to be able to control which encoding gets returned. I'll look into how easy this is.

Issue #181, opened 2014-10-10
Header items get doubled


jagin:
Hi,

i'm using WireMock as a browser proxy and I can see that request header is getting changed before send in a strang way.
To check this out i have set the WireMock connection through another proxy (it was Fiddler on port 8888, http://www.telerik.com/fiddler).
To start WireMock i run the following command:

`java -jar wiremock-1.51-standalone.jar --enable-browser-proxying --proxy-via 127.0.0.1:8888 --verbose`

For the first request the verbose param of the WireMock shows:

```
2014-10-10 10:52:11 Received request: POST http://localhost:8000/api/autologin HTTP/1.1
Host: localhost:8000
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:32.0) Gecko/20100101 Firefox/32.0
Accept: */*
Accept-Language: pl,en-US;q=0.7,en;q=0.3
Accept-Encoding: gzip, deflate
Content-Type: application/json; charset=UTF-8
X-Wolf-Protection: 0.5396181126404231
X-Requested-With: XMLHttpRequest
Referer: http://localhost:8000/
Content-Length: 67
Cookie: connect.sid=s%3ASduRYg0x_QBFDTNsRp0ojmI9AaqQjvuf.WUpVrpQuCGBqWrNY9bfhF8UFXPb5Us0a5%2FCRmMALOtc
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache


2014-10-10 10:52:11 Request received:
POST http://localhost:8000/api/autologin HTTP/1.1
Host: localhost:8000
User-Agent: Mozilla/5.0 (Windows NT 6.1; WOW64; rv:32.0) Gecko/20100101 Firefox/32.0
Accept: */*
Accept-Language: pl,en-US;q=0.7,en;q=0.3
Accept-Encoding: gzip, deflate
Content-Type: application/json; charset=UTF-8
X-Wolf-Protection: 0.5396181126404231
X-Requested-With: XMLHttpRequest
Referer: http://localhost:8000/
Content-Length: 67
Cookie: connect.sid=s%3ASduRYg0x_QBFDTNsRp0ojmI9AaqQjvuf.WUpVrpQuCGBqWrNY9bfhF8UFXPb5Us0a5%2FCRmMALOtc
Connection: keep-alive
Pragma: no-cache
Cache-Control: no-cache

{"token":"","trace":"","data":{"login":"balgab9576"},"locale":"pl"}
[...]

PR #363, opened 2016-02-17
Added bodyAsBase64 element to LoggedRequest JSON in requests find output

RaceCondition:
In order to pass back a binary body such as a zip or jpeg exactly as received in the original request for verification, we need a new `bodyAsBase64` element in the `LoggedRequest` JSON. The original string element `body` corrupts the stored byte array so this was needed as there was no way to retrieve it otherwise.
So calling the usual `http://<host>:<port>/__admin/requests/find` will return the same JSON as before but now it will also have something like this example...

`"bodyAsBase64": "TXkgdGV4dAo=",`

... at the same level as `"body": "My text"` element.

For the moment, this is only in the REST API, not the Java API. We tried the Java API after finding that the REST API didn't return binary but that has the same issue.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
@@ -28,6 +28,8 @@ import java.util.Date;
 import java.util.Map;
 import java.util.Set;
 
+import org.apache.commons.codec.binary.Base64;
+
 import static com.github.tomakehurst.wiremock.common.Strings.stringFromBytes;
 import static com.github.tomakehurst.wiremock.common.Urls.splitQuery;
 import static com.github.tomakehurst.wiremock.http.HttpHeaders.copyOf;
@@ -135,6 +137,12 @@ public class LoggedRequest implements Request {
     @JsonProperty("body")
 	public String getBodyAsString() {
         return stringFromBytes(body);
+	}
+	
+	@Override
+	@JsonProperty("bodyAsBase64")
+	public String getBodyAsBase64(){
+		return Base64.encodeBase64String(body);
 	}
 
 	@Override
--- a/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
@@ -20,7 +20,6 @@ import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.github.tomakehurst.wiremock.http.*;
-import com.google.common.base.Charsets;
 
 import java.net.URI;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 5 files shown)


RaceCondition:
Oh dear. Not sure what the failure with #181.2 is caused by. The tests ran successfully locally when I did a `gradle clean build`.

RaceCondition:
I've tried the same Oracle JVM version used by the failing Travis CI job, being jdk1.7.0_76 and all tests pass. I'm still puzzled as to why they have failed here.

tomakehurst:
Looks like port numbers clashing. I don't think it's a result of anything you've done - since Travis builds run in Docker containers I'm guessing this is a random clash with some other process whose TCP stack is not fully isolated.

Running it again seems to have made the problem go away.

tomakehurst:
Can you elaborate what you mean by this being the REST API only and not the Java API?

RaceCondition:
The REST API was just referring to calling WireMock via JSON+HTTP so thinking about it, it's not really REST so sorry for the confusion there.

The effort to get bodyAsBase64 working was concentrated around that rather than the Java side. I'm not in a position to be able to try the Java Client right now and it's not something we need at this stage given that the JSON+HTTP approach is working nicely with this change.

I hope this is enough of a claification/elaboration for you but if you need more info then ask away.

tomakehurst:
I'm just wondering what you think is missing on the Java side. I can't see anything obvious I'd want to add, since there's no DSL component to this.

RaceCondition:
Ah OK, maybe the Java side is OK now with this change. All I can remember is that without this change there was no way to get the byte array back unmangled (due to UTF-8 conversion I think it was).

The recent addition in v1.58 and v2.0.8 (I think) of the `bodyAsBase64` request element meant that we just needed to amend the code to return it in the response from `__admin/requests/find` and all was good. We didn't look at the Java side anymore - but if you think it'll work with this pull request then that's good. I can't try it out until Friday but a colleague of mine might try it tomorrow.

tomakehurst:
Ah, I think I see the potential problem: because the ``LoggedRequest`` constructor takes body as a String, you'll get a bad version of it back if you use the Java API remotely.

tomakehurst:
Yeah, I think it'd be a bit misleading to leave it in its current state.

I think all that would need to happen would be to modify the ``LoggedRequest`` constructor that takes body as a string to take the base64 version instead (also changing the JSON property name appropriately).

As a bonus, it might be nice to remove the ``byte[]`` based constructor and just use the base64 version everywhere.

RaceCondition:
Right - thanks for the advice. I'll leave it for my colleague to look at tomorrow or look at it myself on Friday when I'm back.
Thanks again.

tomakehurst:
Cool, appreciate you both putting the effort in!

RaceCondition:
I've done a second commit that includes the changes for the Java API as discussed. Both interfaces work and all of the tests pass. If there is anything for me to do beyond this then please let me know.

src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
```diff
@@ -76,15 +77,18 @@ public LoggedRequest(String url,
...
-                         @JsonProperty("loggedDate") Date loggedDate) {
-        this(url, absoluteUrl, method, headers, body.getBytes(Charsets.UTF_8), isBrowserProxyRequest, loggedDate);
+    		@JsonProperty("absoluteUrl") String absoluteUrl,
+    		@JsonProperty("method") RequestMethod method,
+    		@JsonProperty("headers") HttpHeaders headers,
+    		@JsonProperty("browserProxyRequest") boolean isBrowserProxyRequest,
+    		@JsonProperty("loggedDate") Date loggedDate,
+    		@JsonProperty("body") String body,
+    		@JsonProperty("loggedDateString") String loggedDateString,
+    		@JsonProperty("bodyAsBase64") String bodyAsBase64)
```
RaceCondition: I had to add all of the extra JSON properties even though they are unused as test `com.github.tomakehurst.wiremock.RequestQueryAcceptanceTest.requestBodyEncodingRemainsUtf8()` was failing with `java.lang.RuntimeException at RequestQueryAcceptanceTest.java:119`. It was trying to do a base64 decode against the body, which is not base64 encoded. After plenty of digging, these extra properties got it working. I have to admit that it is a bit of a mystery why that is.

PR #371, opened 2016-03-08
adding filename limitation

rousseau-christopher:
There is an error then we have a long url, the UniqueFilenameGenerator don't trim the filename and usual filesystem cannot handle file longer than 255 (windows, linux, mac included).
In that case wiremock throw a write error exception and failed.

I propose to trim the filename to 240 character + extension.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/UniqueFilenameGenerator.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/UniqueFilenameGenerator.java
@@ -17,9 +17,6 @@ package com.github.tomakehurst.wiremock.common;
 
 import com.github.tomakehurst.wiremock.http.Request;
 import com.google.common.base.Joiner;
-import com.google.common.base.Splitter;
-import com.google.common.collect.FluentIterable;
-import com.google.common.collect.Iterables;
 
 import java.net.URI;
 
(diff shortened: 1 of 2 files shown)


rousseau-christopher:
Looks like build on jdk8 failed because port binding ? I don't know how to restart it without pushing modification. Could you restart this build. There is no reason of failure.

tomakehurst:
Yeah, this is a bit of a pain with Travis (and containerised CI in general). Seems there's much higher likelihood of port clashes, despite requesting a random, free one.

rousseau-christopher:
I'm not very good with gradle, but is there any plugin equivalent to this one ?
It offer possibility to reserve a random unused port with maven.

http://www.mojohaus.org/build-helper-maven-plugin/usage.html

tomakehurst:
Reserving random ports seems to be quite tricky. WM delegates to java's ServerSocket as I suspect the maven plugin does. But some OS configurations seem to occasionally allocate the same port twice (race condition perhaps).

SimiJo:
Hi, I have a long url and when recorded using a proxy, the mapping files are not created by wiremock 2.3.1. The fix for the filename too long issue is not available in the latest too, Any workaround?

tomakehurst:
No workaround currently, other than copying/writing your own recorder.

Total rewrite of the recorder in the pipeline though.

tomakehurst:
New recorder coming in #674

PR #381, opened 2016-04-01
Create new client every time, to work around obscure and I/O-level multi-threading issues

kjetilv:
Yes, it's true. On OSX, I encountered an obscure bug. 

To uncover it, I added print statements that I implemented by decorating breakpoints in IDEA:

This is all in apache's httpcore version 4.4.1. In org.apache.http.protocol.HttpRequestExecutor, I added non-suspending and logging breakpoints as follows:

Line 122: "REQ:  " + System.identityHashCode(conn) + ": " + conn + " => " + request
Line 126: "OK:   " + System.identityHashCode(conn) + ": " + conn + " => " + request
Line 128: "FAIL: " + System.identityHashCode(conn) + ": " + conn + " => " + request + ": " + ex

A typical output, which shows the bug at the end:

REQ:  120493608: CPoolProxy{148.121.152.75:56553<->10.179.121.244:8084} => [URL]
REQ:  1402900645: CPoolProxy{148.121.152.75:56552<->10.179.121.244:8084} => [URL]
OK:   120493608: CPoolProxy{148.121.152.75:56553<->10.179.121.244:8084} => [URL]
OK:   1402900645: CPoolProxy{148.121.152.75:56552<->10.179.121.244:8084} => [URL]
REQ:  1848886687: CPoolProxy{148.121.152.75:56552<->10.179.121.244:8084} => [URL]
FAIL: 1848886687: CPoolProxy{0:0:0:0:0:0:9479:984b:56552<->10.179.121.244:8084} => [URL]: org.apache.http.NoHttpResponseException: The target server failed to respond

(The [URL]'s are request objects I have edited out, for clarity.)

The System.identityHashCodes ensure I am printing the same object. We see object #1848886687 goes from one state at request time, and to a new state after. The new state is that the local address has mysteriously become 0:0:0:0:0:0:9479:984b!

What is that, you say? It is an IPv6 address format. However, it describes the same IPv4 IP address!  You can work out what 94.79.98.4b is in decimal for yourself ...

When does this change occur? When the socket connection flushes the outgoing data. (I.e. the request.) This happens in org.apache.http.impl.io.SessionOutputBufferImpl, line 126:

       this.outstream.write(b, off, len);

Don't ask me how it happens, but before we call write here, the connection is fine. And SOMETIMES, when the call returns, it is screwed, and unable to receive incoming data. (Manifesting as NoHttpResponseException: The target server failed to respond, as above, and the FAIL breakpoint hitting.) Not every time, but pretty often. This is why I suspect multi-threading issues.

I am serious. Do NOT ask me how this happens. Could be an issue with http-core and/or my network setup. I just know the pull request makes the FAILs go away.

PS: Sorry about the whitespace/formatting diff, that's my opinionated IDEA. The change is only creating a new client every time. I can clean that up, but it'll have to be a little later.

PPS: This is NOT an april fool's!

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
@@ -51,18 +51,20 @@ public class ProxyResponseRenderer implements ResponseRenderer {
     private static final String CONTENT_LENGTH = "content-length";
     private static final String HOST_HEADER = "host";
 
-    private final HttpClient client;
     private final boolean preserveHostHeader;
     private final String hostHeaderValue;
-	
+	private final ProxySettings proxySettings;
+	private final KeyStoreSettings trustStoreSettings;
+
 	public ProxyResponseRenderer(ProxySettings proxySettings, KeyStoreSettings trustStoreSettings, boolean preserveHostHeader, String hostHeaderValue) {
-		client = HttpClientFactory.createClient(1000, 5 * MINUTES, proxySettings, trustStoreSettings);
+		this.proxySettings = proxySettings;
+		this.trustStoreSettings = trustStoreSettings;
 
         this.preserveHostHeader = preserveHostHeader;
         this.hostHeaderValue = hostHeaderValue;
 	}
 
-    public ProxyResponseRenderer() {
+	public ProxyResponseRenderer() {
         this(ProxySettings.NO_PROXY, KeyStoreSettings.NO_STORE, false, null);
     }
 
@@ -73,7 +75,7 @@ public class ProxyResponseRenderer implements ResponseRenderer {
 
 		try {
 			addBodyIfPostPutOrPatch(httpRequest, responseDefinition);
-			HttpResponse httpResponse = client.execute(httpRequest);
+			HttpResponse httpResponse = client().execute(httpRequest);
 
             return response()
                     .status(httpResponse.getStatusLine().getStatusCode())
@@ -95,15 +97,19 @@ public class ProxyResponseRenderer implements ResponseRenderer {
         if (responseDefinition.getHeaders() != null) {
             httpHeaders.addAll(responseDefinition.getHeaders().all());
         }
-	    
+
 	    return new HttpHeaders(httpHeaders);
     }
+
+	private HttpClient client() {
+		return HttpClientFactory.createClient(1000, 5 * MINUTES, this.proxySettings, this.trustStoreSettings);
+	}
 
     private static HttpUriRequest getHttpRequestFor(ResponseDefinition response) {
 		final RequestMethod method = response.getOriginalRequest().getMethod();
 		final String url = response.getProxyUrl();
 		notifier().info("Proxying: " + method + " " + url);
-		
+
 		if (method.equals(GET))
 			return new HttpGet(url);
 		else if (method.equals(POST))
@@ -123,9 +129,9 @@ public class ProxyResponseRenderer implements ResponseRenderer {
 		else
 			return new GenericHttpUriRequest(method.toString(), url);
 	}
-	
+
 	private void addRequestHeaders(HttpRequest httpRequest, ResponseDefinition response) {
-		Request originalRequest = response.getOriginalRequest(); 
+		Request originalRequest = response.getOriginalRequest();
 		for (String key: originalRequest.getAllHeaderKeys()) {
 			if (headerShouldBeTransferred(key)) {
                 if (!HOST_HEADER.equalsIgnoreCase(key) || preserveHostHeader) {
@@ -142,11 +148,11 @@ public class ProxyResponseRenderer implements ResponseRenderer {
                 }
 			}
 		}
-				
+
 		if (response.getAdditionalProxyRequestHeaders() != null) {
 			for (String key: response.getAdditionalProxyRequestHeaders().keys()) {
 				httpRequest.setHeader(key, response.getAdditionalProxyRequestHeaders().getHeader(key).firstValue());
-			}			
+			}
 		}
 	}
 


tomakehurst:
Are you able to write a test that reliably replicates this (on OSX at least)?

I'm a bit concerned about the cost of recreating the client each time, both in terms of the setup cost and the loss of pooled connections. Any thoughts on how that could be mitigated?

kjetilv:
I could try, but I don't know if it would succeed. On such a low level, I'm not sure what could be the most likely factors. I don't usually meddle in the socket nether regions. And, after all, this is pretty basic behavior, and I'm sure I'm not the first to use it on OSX.

To mitigate: You could use a thread local HttpClient, of course. Then again, that introduces management of those clients, and potential leaks.

I would be very happy if I could turn on this behavior as a configuration option. HttpClient pooling is unlikely to be the big performance win for my application.

tomakehurst:
So if you can't reliably reproduce it in a unit test, under what circumstances does the error happen?

How would we be sure that creating a new client each time fixes the problem?

kjetilv:
Those are both good questions! Especially since we don't know what the problem is, and it occurs unreliably.

Generally, with third-party code that works on a low level, I would advocate a conservative approach to sharing objects and the state that they represent, as it leaves you less vulnerable to these types of errors. In this case, one conservative approach is a fresh client that is exclusive to its thread. It seems to fix my issue. 

For me and others who feel likewise, or who encounter similar weirdness, it would be nice to be able switch to this behavior. Should I add a new configuration option instead, leaving current behavior as the default?

tomakehurst:
Have you filed a bug report with apache?

It'd be great to get this fixed at source rather than increasing complexity with a workaround.

kjetilv:
Nope, sorry. I agree with you, but I have only worked with the WireMock level so far, to analyze and work around it. Not sure how I would proceed to isolate the bug in Apache's code, since it's pretty low-level. The state change occurs during a call to a JDK class, so it could even be in the I/O libraries.

You already pass two primitive parameters to the ProxyResponseRenderer (along with the settings objects), and I agree that adding a third is a slippery slope. If they were wrapped up in a Parameter Object containing those two a new flag (connectionSharing or something - or trustApache!), I don't think complexity would suffer unduly.

tomakehurst:
I think I'd still like to see if we can get this fixed in the apache client before committing to add a workaround to WireMock.

I appreciate this is frustrating when you're just trying to get your work done, but I think it would set a bad precedent to start adding workarounds for every library bug.

tomakehurst:
Bit of a long shot, but have you tried bumping the apache client version to the latest?

kjetilv:
First of all, thanks for being incredibly responsive yesterday.

I have tried a newer version, but I think I may have messed up the Maven dependencies so the change didn't "take". I will try again.

Lastly, I have limited time to go chasing this bug (it took some time to isolate even on the WireMock level of the stack), and first I need to figure out whether it's actually in the Apache library, or if it is in the JDK. So I'm considering running my forked version until I can get a closer look at it.

tomakehurst:
No problem,  sorry I couldn't be more help.

Good luck with your upgrade/workaround - hope it solves your problem.

One last question: which OSX version are you running?

guspower:
I had a quick attempt to replicate this problem under Linux  but with no success. [Here is the set up I used](https://github.com/guspower/wiremock-httpclient) - @tomakehurst @kjetilv would either of you be able to run this on your Macs and let me know if you can replicate the problem?

kjetilv:
Impressive effort, thank you. 

This code does not replicate the bug on my machine either. I've made a similar unit test, which I will push to my forked branch later.

I have been able to reproduce and isolate the behavior, but so far only against a particular proxied server.  (The one that displayed the problem in the first place.) When I run the test against other targets (eg. public websites), the bug does not appear. Something about that server is causing bad IPv6-ness to creep into the I/O level. The server in question is a commercial product running on a Tomcat. (It responds with Apache-Coyote/1.1 - not sure if I am at liberty to reveal the exact product.)

I have done some more research using this case. For instance, the conditions for the bug to arise is simply that a connection is reused. A CPoolProxy (implementation of ManagedHttpClientConnection) receives a CPoolEntry with a LoggingManagedHttpClientConnection where the socketHolder is already set to a socket. This seems to be the case for all connections that encounter the issue. (And they are many.)

I have also pinpointed the state change to happen during a native (!) call to SocketOutputStream. I'm not going to dive into that, so I am trying to understand the conditions that cause the error instead.

kjetilv:
As to Tom's question, I am running the latest OSX.

guspower:
Ah that's interesting. Do you know what version of Tomcat it's running under-the-hood?

kjetilv:
I think it might be a Tomcat 4 – I downloaded it and it responds with Apache/Coyote 1.1. But it does not provoke the error, either.

guspower:
wow that's properly old skool! - could you verify the tomcat version using something like:

    java -cp lib/catalina.jar org.apache.catalina.util.ServerInfo 

that'd be a big help

kjetilv:
I'll have a closer look at it tomorrow. Need to go through some gatekeepers at work.

kjetilv:
Turns out it's 6.0.29. Still not reproducible on my machine, even when I run it with the same WAR file that provokes the issue in the field.

Not really that surprising, since I suspect it's an issue on a lower level.

kjetilv:
In case anyone is still following this pull request, some updates:

- My problem can be solved if WireMock can be changed to call useSystemProperties() on the HttpClientBuilder it uses. This makes the builder respect system properties, which in turns opens up ways of configuration already supported in the HttpClient library. One of these is http.keepAlive, which selects the correct pooling strategy.
- Gave up on diagnosing the bug. Life is too short!

guspower:
Ah ok this is interesting. So - just to be clear - when HttpClient is configured to use http.keepAlive you no longer see the NoHttpResponseException against the application's embedded Tomcat 6.0.29?

kjetilv:
That's right. Setting this property makes it select the no-pooling strategy, which doesn't provoke the error.

WireMock just needs to call useSystemProperties() on the HttpClientBuilder, so that it will be respected.

tomakehurst:
Just so I understand - adding ``useSystemProperties()`` would simply allow you to disable pooling via system properties if you wished to?

If so, I'm happy for this to be included. Could you raise a PR?

kjetilv:
Raised https://github.com/tomakehurst/wiremock/pull/392

tomakehurst:
Closing as effectively fixed by #392.

PR #400, opened 2016-05-06
Added ability to record a particular request/response repeatedly

selangley:
This is my proposed addition for:

Two ideas for record/playback of repeated requests and response
https://groups.google.com/d/topic/wiremock-user/tcX9TuCLCeE/discussion

It allows for specifying an optional command-line parameter:
    --record-repeats
that can be used as an addition to the:
    --record-mappings 
command-line parameter.

The first request/response recorded matching a particular request is saved with the file ending .1.json.  The second will end in .2.json. The third will end in .3.json, etc.

Playback assumes the requests will be received and played back to the client in the same order as they were recorded.  If you want different behavior, you should modify the scenario values in the corresponding mapping files according.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -158,9 +158,9 @@ public class WireMockServer implements Container, LocalStubbing, Admin {
 		stubRequestHandler.addRequestListener(listener);
 	}
 	
-	public void enableRecordMappings(FileSource mappingsFileSource, FileSource filesFileSource) {
+	public void enableRecordMappings(FileSource mappingsFileSource, FileSource filesFileSource,	boolean recordRepeatsEnabled) {
 	    addMockServiceRequestListener(
-                new StubMappingJsonRecorder(mappingsFileSource, filesFileSource, wireMockApp, options.matchingHeaders()));
+                new StubMappingJsonRecorder(mappingsFileSource, filesFileSource, wireMockApp, options.matchingHeaders(), recordRepeatsEnabled));
         notifier.info("Recording mappings to " + mappingsFileSource.getPath());
 	}
 
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
@@ -57,6 +57,7 @@ public class CommandLineOptions implements Options {
 	
 	private static final String HELP = "help";
 	private static final String RECORD_MAPPINGS = "record-mappings";
+	private static final String RECORD_REPEATS = "record-repeats";
 	private static final String MATCH_HEADERS = "match-headers";
 	private static final String PROXY_ALL = "proxy-all";
     private static final String PRESERVE_HOST_HEADER = "preserve-host-header";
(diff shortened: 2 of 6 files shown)


tomakehurst:
Thanks for creating this, and apologies for not reviewing it yet. I'm not ignoring it - just have a large backlog to clear!

tomakehurst:
Hi, apologies again for having left this so long.

I've had one or two requests recently for a multiple mapping file format (akin to Betamax's tapes). I've also been asked about a shorthand for scenarios when you've just a series of responses for a given GET or whatever.

I'm wondering whether there's some intersection between your ideas and this.

The main advantage of single files is the ability to guarantee that they're loaded in the order they were recorded (although in fairness this could be done by sorting on creation date in the current setup).

How would something like ``--single-file-recording --record-scenarios`` work for you, where the second parameter indicates that repeated identical requests should form a chain of scenario states?

It seems like that might satisfy your use case with a lot less work on your part, but perhaps I'm not fully understanding what you're trying to do.

selangley:
Thanks for looking at this.

I believe that what you are suggesting would end up giving me the same functionality I’m after.

By single-file-recording, do you literally mean an entire  WireMock recording session would be recorded into a single file?  If so, navigating and editing that file would get a little unwieldy after a few dozen kilobytes.

If you mean just the mapping  files for a recording session are condensed into a single mapping file, while the body response files remain as they are, one for each distinct response, I like that better.

Or you might mean that for a particular request mapping, all the scenarios and  corresponding  pointers to body responses are contained in a single mapping file.  And each particular request mapping gets its own mapping file.   I’d be happy with that, too.

I don’t have any out-of-order issues with my current PR  when it comes to response playback because it relies on the sequentially generated scenario state transitions.  It’s true that the order the recorded responses are loaded into WireMock is not guaranteed to match the order in which they were recorded.  But for my purposes, loading order doesn’t matter since the scenario state transitions were defined correctly at recording time.

Regards,

Scott

tomakehurst:
OK, thanks for the insight.

I agree that keeping body files separate is a good idea. The main benefit from recording to single files is the mappings loading in a predictable order, although I take your point that if they're all within scenarios it doesn't matter. I think there are probably circumstances where recording into a scenario isn't the right choice though, but you still want to get out what you put in without worrying about de-duplication.

Let me think about this a little more, and review the other similar requests I've had. I think it's quite important to get this right, so I don't want to rush to a design.

laser:
@selangley - Out of curiosity, is there a way to enable `record-repeats` without using a command-line flag? 

I ask because the `WireMockServer` constructor has a parameter of type [`Options`](https://github.com/selangley/wiremock/blob/2.0-beta/src/main/java/com/github/tomakehurst/wiremock/core/Options.java#L33) - but your PR seems to affect a change to [`CommandLineOptions`](https://github.com/selangley/wiremock/blob/2.0-beta/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java#L102) only.

tomakehurst:
FYI, I've decided I'm not going to merge this PR.

It's a special case of a more general complaint I've heard a few times now so I'm going to write a new recorder that addresses these properly.

selangley:
I'll look forward to to the new recorder.  

As for the options, record-repeats is set as a command-line option.  It's an optional, companion parameter to record-mappings which is also set via the command-line.

PR #411, opened 2016-05-19
proper encoding

mariuszs:
fix encoding for posted mappings


tomakehurst:
I take that to mean you think this isn't required anymore?

mariuszs:
Yes, it is required, because we have many problems related to stub definitions defined in json with country specific characters inside. @twroniak is preparing complete PR for this issue.

mariuszs:
https://github.com/tomakehurst/wiremock/pull/412/files

PR #412, opened 2016-05-20
json encoding is not lost when mocking with java api

twroniak:


mariuszs:
:+1:  excellent :)

src/test/java/com/github/tomakehurst/wiremock/client/WireMockClientAcceptanceTest.java
```diff
@@ -65,4 +65,17 @@ public void buildsMappingFromStaticSyntax() {
 		
 		assertThat(testClient.get("/my/new/resource").statusCode(), is(304));
 	}
+	
+	@Test
+    public void buildsMappingWithUrlOnyRequestAndResponseWithJsonBodyWithDiacriticSigns() {
```
mariuszs: please format

src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
```diff
@@ -165,8 +164,7 @@ private String postJsonAssertOkAndReturnBody(String url, String json, int expect
         HttpPost post = new HttpPost(url);
         try {
             if (json != null) {
-                StringEntity stringEntity = new StringEntity(json);
-                stringEntity.setContentType(APPLICATION_JSON.getMimeType());
+                StringEntity stringEntity = new StringEntity(json, APPLICATION_JSON);
```
szpak: Nice catch. I didn't know that, but from the implementation point of view (in StringEntity) it is easier.

Btw, @mariuszs maybe you should report it upstream (as an possible engancement)?

mariuszs: Do you mean master?

tomakehurst:
Hey apologies for the delay getting to this.

Can you provide me with some context for this? I'm struggling to understand exactly what problem this solves.

Thanks.

twroniak:
Hello,
Takie a look at the test I've provided, if you revert the change, the test will fail. The bug is caused by single-argument StringEntity constructor - it sets the content type to ISO-8859-1 regardless of the original encoding.

Cheers!

tomakehurst:
Ah, OK so it's actually setting up the stub incorrectly without the encoding. Very odd.

Thanks!

mariuszs:
@tomakehurst why orginal author is erased with your commit?

tomakehurst:
This is standard behaviour when Github merges from a PR AFAIK. All I did was hit the Merge button on this occasion.

tomakehurst:
Ah, that's just the extra merge commit that gets generated. Here's the original with author intact:
0305fa43cbe6aeed60997534f3fec0f35fca40cf

mariuszs:
Ok, I see :beer:

PR #419, opened 2016-06-02
Allow responses to contain request parameters

### LINKED ISSUE #405: Add a ResponseTransformer to support response templating
One of the most common feature requests is a way to copy request parameters into the response.

I propose embedding a very simple template engine - possibly StringTemplate but can be persuaded otherwise - allowing response headers and body to do simple substitutions e.g.

``` javascript
{
    "request": {
            "method": "GET",
            "url": "/some/thing"
    },
    "response": {
            "status": 200,
            "body": "Request: $$request.body$$",
            "headers": {
                    "Content-Type": "$$request.headers.Accept$$"
            }
    }
}
```

SethBresnettSoftwire:
Fixes #405 

A couple of things to note:
- This only takes the first query param or header value corresponding to the given key
- Headers have to be referred to with '-' replaced by '_' - this is because StringTemplate needs the keys to be valid Java identifiers. I don't think there's any other templating library that's suitable for this and doesn't have that restriction, so I think this is the best workaround.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -59,6 +59,7 @@ dependencies {
     compile("junit:junit:4.12") {
         exclude group: "org.hamcrest", module: "hamcrest-core"
     }
+    compile "org.antlr:ST4:4.0.8"
 
     testCompile "org.hamcrest:hamcrest-all:1.3"
     testCompile("org.jmock:jmock:2.5.1") {
@@ -131,6 +132,7 @@ shadowJar {
     relocate "com.jayway", 'wiremock.com.jayway'
     relocate "org.objectweb", 'wiremock.org.objectweb'
     relocate "org.custommonkey", "wiremock.org.custommonkey"
+    relocate "org.antlr", "wiremock.org.antlr"
 
     dependencies {
         exclude(dependency('junit:junit'))
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/TemplateRequestAdapter.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/TemplateRequestAdapter.java
@@ -1,0 +1,66 @@
+package com.github.tomakehurst.wiremock.extension;
+
+import com.github.tomakehurst.wiremock.http.HttpHeader;
+import com.github.tomakehurst.wiremock.http.QueryParameter;
+import com.github.tomakehurst.wiremock.http.Request;
+import com.github.tomakehurst.wiremock.http.RequestMethod;
+import com.google.common.collect.Maps;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.Map;
+
+import static com.github.tomakehurst.wiremock.common.Urls.splitQuery;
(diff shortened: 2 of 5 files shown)


tomakehurst:
The headers thing troubles me a bit, and I admit I probably haven't thought it through properly. It seems there are other characters e.g. # that are valid for header names, but not for Java identifiers.

Would it be possible to address values within a multi-valued query param/header with [x] type syntax?

SethBresnettSoftwire:
I think in practice the only characters that are going to appear in headers are letters and '-', so while it's a bit unpleasant, I think the real world use-cases are covered here.

For the multiple values for headers/query strings, I think the best that can be done with StringTemplate is something like $request.query.id.("0")$ (treating the multiple values as a map from an index to each value). I can add that if you think it would be a good improvement?

tomakehurst:
Could the same trick you're suggesting for accessing values by index also be used to specify quirky header names?

Also, could you wrap the header value in such a way that ``$request.query.id$`` would show the first item, but ``$request.query.id.("<n>")$`` also works?

pjotal:
this is an awesome feature. when are you guys planning to release it? it's very useful

scottaj:
I'm confused, how does this relate to the documentation [here](http://wiremock.org/docs/response-templating/)? I was trying to get the docs in that link to work and couldn't, which led me here and to thinking this is not actually implemented?

tomakehurst:
Response templating is a solution to this. Just haven't closed this issue off yet.

Can you elaborate why you couldn't get it to work?

scottaj:
Ah okay, that makes sense. I was having trouble getting it to work with a body file (it isn't clear if that should work from the docs)

I have a response section in a stub like this: 
```
"response" : {
    "status" : 200,
    "bodyFileName" : "partially-approved.json",
    "transformers": ["response-template"],
    "headers" : {
      "Cache-Control" : "private, max-age=0",
      "Content-Type" : "text/json; charset=utf-8",
      "Server" : "Microsoft-IIS/8.5",
      "X-AspNet-Version" : "4.0.30319",
      "X-Powered-By" : "ASP.NET",
      "Access-Control-Allow-Origin" : "*",
      "Date" : "Mon, 27 Feb 2017 20:09:29 GMT",
      "Connection" : "close"
    }
  }
```

And then my body file is a JSON file with properties that are templated out with handlebars. I'm not sure if I should even expect that to work, but right now it just sends back the literal `{{templateVars}}` in the response. 

e.g. this is a snippet from my body file:

```
{
  "TerminalID": "{{request.query.terminalId}}",
  "ResultFinal": true,
  "AuthCode": "28063A",
  "TransType": "sale",
...
```

When it comes back it has the literal string `{{request.query.terminalId}}`

tomakehurst:
Have you enabled response templating, either via the CLI argument or by adding the extension if you're starting it programmatically? Just adding the transformer name to the stub isn't enough on its own.

scottaj:
That was my problem. I had several wiremock rules and added the transformer to the wrong one

PR #420, opened 2016-06-02
add multipart support

policybased:
Hi,

This is a basic multipart support implementation.

Maybe it would be a good idea to have stub the form data in the ReqeustPattern class. Something like withFormData. It should not matter how the form data are received (Url encoded or multi part).

To handle the form data for now, I implemented a wiremock extension. But it would be perfect to have this feature included in Wiremock.

Best regards

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/HttpsSettings.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/HttpsSettings.java
@@ -22,16 +22,20 @@ public class HttpsSettings {
     private final int port;
     private final String keyStorePath;
     private final String keyStorePassword;
+    private final String keyStoreType;
     private final String trustStorePath;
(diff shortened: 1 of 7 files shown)


tomakehurst:
I think this is a great idea in principle, but needs a bit more discussion before committing to an implementation.

There are a couple of reasons I'd be reluctant to merge this at the moment:
1) There are no tests
2) You've exposed a servlet API class in the ``Request`` interface, which isn't a great idea as this is supposed to abstract away completely from any specific implementation. I'd actually like to remove the servlet API dependency completely at some point, so this would be a step in the wrong direction.

Agree that some thought needs to be put into how to stub and verify based on form parts. At the moment there's not much value added by having it on  ``Request`` and its impls.

tomakehurst:
Closing this, as it's been dormant for a while.

PR #421, opened 2016-06-03
Accept gzipped request bodies

### LINKED ISSUE #404: Support gzipped request bodies
Although WireMock now supports gzipped responses, gzipped request bodies are not. Would be great to have this too.

SethBresnettSoftwire:
Fixes #404 

My IDE's done a bit of tidying up in JettyHttpServletRequestAdapter, I can take that out if you'd prefer a cleaner diff.

Also we can maybe just rely on the Content-Encoding header rather than looking at the magic bytes. (That would also mean we can do it without turning the response back and forth between a stream and an array), let me know if you'd prefer it that way.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Gzip.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Gzip.java
@@ -15,7 +15,6 @@
  */
 package com.github.tomakehurst.wiremock.common;
 
-import com.google.common.base.Charsets;
 import com.google.common.io.ByteStreams;
 
 import java.io.ByteArrayInputStream;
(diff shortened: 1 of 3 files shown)


tomakehurst:
Thanks for this Seth. I'll try and get this reviewed by EOP today.

On the IDE front, things are a bit of a mess at the moment. The unofficial standard is tab=4 spaces, and using space characters rather than tabs. However, there's still code in there from PRs and probably also from me using Eclipse back in the day that uses tabs.

I'm planning to start enforcing this in the build with Checkstyle at some point, then do a one-time cleanup so that this is less painful in future.

tomakehurst:
Excellent stuff, thanks!

PR #423, opened 2016-06-07
Reverts to WireMock 1 behaviour - Date header only returned if stubbed

Mahoney:
In WireMock v1 no Date header is returned unless explicitly stubbed.

This PR returns WireMock v2 to that behaviour; previously it generated a Date header field regardless of stubbing, meaning that if you stubbed the Date header field you got two, with the stubbed one later in the header field order so ignored by most clients.


tomakehurst:
Thanks!

PR #428, opened 2016-06-09
Add Admin UI and server the same from __admin_ui

dlaha21:
Sorry for such a delay, I should have created this PR earlier as we already had all the prerequisites merged for this. 

This PR adds the Admin UI and servers it from `/__admin_ui/`. Looking forward to your feedback. There are not tests for the UI (couldn't find enough time for it) and we can add them incrementally if you are fine with that approach.


handstandsam:
Thanks Dip.  I checked this branch out a few months back, and it looked nice.  I haven't looked at it again now, but looked through the PR.

Should the vendor JS files be the minified versions to reduce size, or is there benefit in the full source versions?

tomakehurst:
The lack of tests is a dealbreaker for me I'm afraid.

The reason is that as soon as I merge this (or any change for that matter) I become the primary maintainer. Two things I know from painful experience are that tests rarely get added afterwards and that having to manually regression test a UI on every release inevitably leads to a loss of productivity, quality or both.

tomakehurst:
How about this: rather than making this a core feature of WireMock (which I've always been a bit ambivalent about TBH), how about spinning it out into its own project? 

This would mean you'd retain ownership of it and be able to evolve it at a pace that suited you. I'd happily promote it from wiremock.org.

Making it its own project would have a number of advantages such as making it usable as an embeddable component on another site/app. I've seen several teams build their own UIs for WireMock and embed them into mini testing sites. Having a pre-made component to drop in might be very useful in this instance.

dlaha21:
I completely understand your apprehension due to the lack of tests. Spinning this off as a separate project makes sense and if you can mention that in wiremock.org, even better.

The easiest way which comes to my mind is having a Node.js express app serving the UI and proxying the wiremock admin requests to the actual server. Can you think of any other way it can be dropped in as a component while running wiremock?

tomakehurst:
If I add CORS headers to the ``__admin`` range of URLs, then it should be possible for the UI to call it without any proxying required. CORS support is very good these days, just older versions of IE that can't cope with it.

tomakehurst:
Hah, in fact there was a PR to do this a while back so it's already there!

PR #430, opened 2016-06-13
Adding Remove Stub API 

planetakshay:
I am adding a new API in stand alone mode to be able to remove a stub mapping - you can contact me

akshay.deodhar@gmail.com /  @akshaydeodhar


tomakehurst:
This is a good start. I suggest a few tweaks:

1) Provide an implementation for ``HttpAdminClient.removeStubMapping()``, rather than throwing an exception.
2) Write an acceptance test for the case where the stub you wish to delete is present, and one for the case where it isn't
3) Add ``removeStubMapping()`` to the ``Stubbing`` interface so that it's available everywhere local stubbing is.
4) Use the UUID for the mapping to identify it for deletion rather than checking for request pattern equality, as this is how the edit function already works.

planetakshay:
Tom,
I have implemented Stubbing and HttpClient admin changes but i have question about UUID.
in 2.0 beta is client who is using in stand alone mode and is making post req to configure stubs, expected to pass UUID? i dont think thats the intention. 

As a client when i set up stub i only post something like this with no UUID and i should be able to post the same thing to be able to remove it.
{
  "request": {
    "method": "GET",
    "url": "/v9/something/accounts/25"
  },
  "response": {
    "status": 202,
    "body": "response for test",
    "headers": {
      "Content-Type": "text/plain"
    }
  }
}

So when i call remove, system [ WM ] seems to generate a UUID on its own thats unique and so remove never happens since it does not match with any of existing stubs.

tomakehurst:
You don't need to check for a custom matcher and throw the exception in this case, so I suggest removing that.

As a client you can optionally add a ``uuid`` attribute when you create a stub. If you don't it'll still be assigned one, so you can read it back then use it for deletion if you need to.

planetakshay:
Tom,
As i mentioned, I am not adding UUID as a client when i set it up. So when i make a Remove Call WM assigns a UUID. *Note that this UUID that is assigned by WM to remove is again a unique one* and is not same as the one previously matching mapping-- 

so even though mapping.getRequest() finds a match - UUID still does not match and it never results in removing a mapping is what i am saying.

tomakehurst:
Adding the UUID when creating the stub is pretty trivial:
```java
  UUID myId = UUID.randomUUID();
  stubFor(get(urlEqualTo("/"))
            .withId(myId)
            .willReturn(aResponse()));
```

I suggest doing this if you know you need to delete a stub later.

planetakshay:
Tom,
Yes i understand and i could add a stub like how you have mentioned.
However, its not about me, about users.

Do you want clients to remember UUID for all the stubs they have created? I am a test engineer and as a client/user, i would not want to do that.
How the uniqueness for a stubb is maintained should be transparent to me as a user i think. 
As a client all i should know is i set up stub-1 and i can remove stub-1.

Here whats happening is client sets 
stub-1 and stub-2 and system assigns uuid-1 and uuid-2 respectively.
Then i send a remove call for stub-1 and system assigns uuid-3 and match never happens and stub never gets deleted.

tomakehurst:
How is having to remember the whole request pattern any simpler than remembering the UUID?

By analogy - if I'm going to delete a row from a SQL table, I'm probably going to use the primary key in the where clause rather than than each non-key column.

And I think maintaining consistency with the way editing works is definitely a good idea.

tomakehurst:
One further point - it's possible to register multiple stub mappings with identical request patterns, where the newest one wins. Your version of remove would delete all of them, whereas a UUID based version would only delete one you're targeting.

tomakehurst:
Here's an idea:
Change all of the stub creation methods (``givenThat() stubFor()`` etc.) to return the ``StubMapping`` just created. Keep ``StubMapping`` as the parameter type for the remove method, but internally use the UUID to find the object to remove.

That way users don't need to explicitly set the UUID, and the edit and remove methods are the same syntactically and semantically.

planetakshay:
Tom thanks for your perspective,

From my tests, requests are "pre-known" and are read from the file so we have nothing to maintain and so do most users i believe.Where generating a UUID and maintaining is an additional burden..

 I am not using givenThat()/s stubFor() - my wiremock runs in AWS EC2 and i make RestAssured -> POST to configure it. 

Also my remove will also delete the first request that matches and not all. i return with true.

Going back to primary key argument, it would be like asking a customer to maintain primary keys of all their records if they want to retrieve info about themselves instead of just giving the system their first name last name and then system giving them all relevant info. kinda like asking amazon customer to remember all their order ids rather then being able to search with just the items they bought!

I do't know, i can implement it the way you want but i figured to make WiremockMock more useful, i should put forward my perspective.

tomakehurst:
OK, true you're deleting the first match not them all. You'd be unable to delete any other stubs with identical requests (at least until you'd deleted more recently added ones).

I'm also thinking about the case where you've created, edited, then try to remove the stub. You have to keep track of your edits if you're going to successfully delete at the end i.e. you need to keep a reference to your most recent version, not the initial one.

planetakshay:
How about we give precedence to UUID - if provided delete only that one or else delete the first one that matches.

later we can have removeAll - where if request pattern matches then delete all stubs irrespective of uuid.

tomakehurst:
Yeah OK, that sounds like a good solution.

planetakshay:
For now,

i have removed custom matcher look up and removing stubs based on UuID to begin with.
This is an easier change and should serve what we need to do however, i still plan to implement the request based removal change to that remove method.

What are you guidelines on commenting. i do not see any comments anywhere in code.

tomakehurst:
I'm of the opinion that comments should be a last resort. If the code's intent isn't clear from its structure and naming, then it should be refactored to improve this.

One thing I forgot to mention - in your final commit please can you also update the docs, including examples of how to remove stubs via the Java and JSON APIs.

planetakshay:
Tom,
I added logic to remove stubs based on request if the UUID do not match. also fixed a NPE
Added new test too. For my last commit it says it can not automatically merge.

also, i am not sure where to add docs..sorry i am new to this open source contribution thing.

tomakehurst:
Look under docs/source. The doc engine is Sphinx which uses restructured text - similar to Markdown.

planetakshay:
Tom,  
i synced my fork and resolved conflict, build it locally ran all unit tests and then did a commit and push. but after merging that to my pull request seems to cause an error...i may have checked in some git file not quite sure what it is..
Can you see..

tomakehurst:
It looks like you've blown away some of the new classes I introduced. Might be worth re-doing the merge (I usually find this easier than trying to fix one that's gone wrong).

planetakshay:
Tom,

i saw that in my intelliJ there were 74 files or something that were not pushed. i pushed them and not pull request build is successful.

Please check. **Also looks like i have checked in .gitIgnore** and that also got merged.

planetakshay:
Added the documentation as well.

Can you please look over my commit and accept it under 2.0-beta branch?

planetakshay:
@tomakehurst  ping!

tomakehurst:
Thanks for this.

FYI I'm going to refactor the remove logic to a functional style, for brevity and consistency.

planetakshay:
Thanks Tom.

This one was kind of a bug mess, I have submitted much cleaner #433 for you to merge yesterday. This is in case that adds value/saves you time.

tomakehurst:
Ah damn it, I hadn't spotted that you'd opened a new PR.

planetakshay:
Also i noticed that your remove method seems to use some google library that shows
" Removes, from an iterable, every element that satisfies the provided
   * predicate."
so now the documentation says that it only removes one on first match basis..

tomakehurst:
Good catch. Although now I think of it, perhaps it's more intuitive to delete all the matching stubs?

planetakshay:
does new create multiple stubs if provided? i was designing remove to do exactly opposite of new but i am ok either way. i can change the document if you decide to keep it this way.

did you look at  **.ignore** file i merged by mistake..would that be a problem?

planetakshay:
i have changed the documentation to reflect that it would remove all matching mappings and also resolved a rst example viewing issue in RC1. See pull request #453

PR #432, opened 2016-06-16
refine stubbing and verification acceptance tests for some matchers

### LINKED ISSUE #408: Matching with urlPathMatching is not exact
( The same problem existed for `urlPathEqualTo()` and was fixed in https://github.com/tomakehurst/wiremock/issues/323. )

The following stub

    stubFor(post(urlPathMatching("/endpoint/user/12345/success")).willReturn(aResponse().withStatus(500)));

will return 500 also on shorter URLs like `http://localhost:8090/endpoint/user/12345`

Following the documentation and also in regard to the sister functionality of `urlPathEqualTo()`, I think `urlPathMatching()` should do an exact match as well.

As a fix, it should be sufficient to get the actual path first like in the `urlPathEqualTo()` case and then match without appending `.*`.

zuckel:
matchers: urlEqualTo, urlPathEqualTo, urlPathMatching

I wanted to fix #408 but it turned out that the issue is already fixed in 2.0-beta.
These are the refined tests which I think should be added as enhanced regression tests anyway.

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/StubbingAcceptanceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/StubbingAcceptanceTest.java
@@ -143,6 +143,13 @@ public class StubbingAcceptanceTest extends AcceptanceTestBase {
 	}
 
 	@Test
+	public void doesNotMatchOnUrlPathWhenPathShorter() {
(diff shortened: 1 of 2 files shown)


tomakehurst:
Thanks!

PR #433, opened 2016-06-16
Cleaner Change for Feature "Remove Stub API"

akshaydeodhar:
Tom this change request is in lieu of #430 
This is much cleaner and has only 14 file changes. This is to add remove stub apis.


tomakehurst:
I've re-written quite a bit of this now, rendering this change obsolete.

PR #436, opened 2016-06-20
Android https support for wiremock

abhagupta:
PR contains:
 - patch for Android support for HTTPS communication, which needs the Keystore to be 'BKS' type instead of 'JKS'.  
 - The PR is intended to give users an additional option to send KeyStoreType and TrustStoreType as Wiremock configuration when starting the server. For example,
```
WireMockServer wireMockServer  = new WireMockServer(wireMockConfig().port(9988).httpsPort(9943).keyStoreType("BKS").keystorePath("path/to/keystore")
```

Please note that the acceptance test could not be created for this addition, because the normal JVM only supports JKS keystore type and the acceptance test fails if the test is executed on desktop.  I was trying to forcefully set the  JVM security to support BKS just for the test, but couldn't find how to. The JVM settings have to be set manually by adding Boucy Castle Provider in the Java_Hom>jre>lib>security>java.security file.  @tomakehurst  please let me know if lack of acceptance test is ok in this situation?

I am going to make the change for the standalone process as well so that both standalone and non-standalone versions match. Probably in a seperate PR. (or let me know if that is also  required for acceptance of this one.)  Also, documentation is in progress. Please let me know if any questions.

Thanks for all the help.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/HttpsSettings.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/HttpsSettings.java
@@ -22,16 +22,20 @@ public class HttpsSettings {
     private final int port;
     private final String keyStorePath;
     private final String keyStorePassword;
+    private final String keyStoreType;
     private final String trustStorePath;
     private final String trustStorePassword;
+    private final String trustStoreType;
     private final boolean needClientAuth;
 
-    public HttpsSettings(int port, String keyStorePath, String keyStorePassword, String trustStorePath, String trustStorePassword, boolean needClientAuth) {
+    public HttpsSettings(int port, String keyStorePath, String keyStorePassword, String keyStoreType,  String trustStorePath, String trustStorePassword, String trustStoreType, boolean needClientAuth) {
         this.port = port;
         this.keyStorePath = keyStorePath;
         this.keyStorePassword = keyStorePassword;
+        this.keyStoreType = keyStoreType;
         this.trustStorePath = trustStorePath;
         this.trustStorePassword = trustStorePassword;
+        this.trustStoreType = trustStoreType;
         this.needClientAuth = needClientAuth;
     }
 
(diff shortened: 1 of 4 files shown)


abhagupta:
BTW.. @tomakehurst   This issue needs to be addressed https://groups.google.com/forum/#!topic/wiremock-user/qDy0eZ2RjbI 
before new version of wiremock can work on Android. This PR is going to work on the version 2.0.8

tomakehurst:
Thanks for taking the time to work on this. I'll try and review it tomorrow.

abhagupta:
@tomakehurst  While I am still completing the blog and docs for this,  I forked @handstandsam 's repo to support the HTTPS version of endpoints. The instructions are very raw but might help you test the PR if you need. please let me know if you have any questions.

tomakehurst:
Thanks for this. Looks fine so far, and I agree that an acceptance test isn't really practical in this instance.

I'll hold off merging until you've got the doc commit in there and sorted out the base64 issue.

abhagupta:
https://medium.com/@abhaguptatweet/android-testing-with-mocking-fcec5b9f71c2#.yodirp954  is the document that can be used for instructions

tomakehurst:
I've merged this to master now and made some tweaks that appear to fix the encoding problems. I've managed to get the test to pass in your example repo, and got WireMock to start with HTTPS enabled.

Would appreciate if you could confirm whether I've actually fixed all the issues you encountered!

abhagupta:
Thanks Tom. Yes I will check the merged code by tomorrow.

abhagupta:
@tomakehurst  Apologies for the delay. I have tested the wiremock-2.1.5-rc4.jar in Android with both http and https protocol. Everything is working great!!! Acceptance tests are passing too , so I am assuming the non-android version of Android is working as well... Thanks again and accepting the PR.

abhagupta:
Just realized that we should also provide the `--keystoretype` option in the standalone version as well. I will see if I can submit a PR for it. It might take a while though ..(probably by end of this week). If you are able to integrate this, that is fine too..

tomakehurst:
Excellent, thanks for the validation. I can release 2.0 for real now!

I'd spotted the lack of a CLI arg. Probably good to have it long term, but I don't see it as a showstopper since Android the the main use case for switching it, and that's always going to be initiated programmatically.

Issue #407, opened 2016-05-07
Build additional artefact for Java 8+, Jetty 9.3 and HTTP/2


tomakehurst:
To maintain Java 7 support (which the Android community highly value, along with others in orgs that can't upgrade yet) Jetty 9.2 is as high as we can go.

It would be great to start supporting HTTP/2 as soon as possible though, so I propose creating a 2nd standalone JAR artefact in the build, specifically targeting this configuration.

This would probably necessitate splitting the source tree up into core/ java7/ and java8/ (or something along those lines) and composing core/ with one of the other two at build time.

dsyer:
From what I can tell, there's actually not a *very* deep dependency on Jetty for the core features of Wiremock. If we were to split it up along the lines suggested and remove the Jetty dependencies to separate libraries, it would make it possible to use any Servlet container. Probably it would make the "deployed war" option a bit safer as well, since you could be sure that the main code paths do not depend on container code. I'm quite interested in a version that works flexibly with Spring Boot, and just uses the container that the project under test already has, instead of forcing it to use Jetty (i.e. Tomcat 8, Jetty 9, Undertow). I could help with implementing that here or host it somewhere else, but it would probably be a hack until the code is modularlized a bit. I have some proof of concept code that shows the level of hackery required, and it isn't much, so the modularization is probably not too hard.

tomakehurst:
Hi Dave,

Thanks for the offer to help with this.

We've worked pretty hard to isolate the Jetty dependency. A while back I did a scratch refactoring to get it working with the core Java web server, and I got the tests for all the core features passing.

Unfortunately there are a few less-than-obvious things that depend on some combination of Jetty or the Servlet API such as gzipped responses, file serving and fault injection. It would be great to eliminate these dependencies, if only to remove the Servlet API requirement (which seems to be a regular source of difficulty).

Could you elaborate your motivation for wanting to use your own container with it? Is it simply about dependency (in)compatibility, or something else?

I'm just wondering if dropping the Servlet API would solve your problem.

dsyer:
[...]

PR #440, opened 2016-06-27
Isolate jetty dependencies into a single package

dsyer:
The servlet and request adapters have moved to the "servlet" package.
The only dependency they had on jetty was conditional anyway, so
the condition was also modified to not even require jetty on the
classpath.

See gh-407. Not a complete modularization yet, but splits up the
servlet-only stuff from the jetty server.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/servlet/WireMockHttpServletRequestAdapter.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/servlet/WireMockHttpServletRequestAdapter.java
@@ -13,12 +13,15 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package com.github.tomakehurst.wiremock.jetty6;
+package com.github.tomakehurst.wiremock.servlet;
 
 import com.github.tomakehurst.wiremock.http.*;
 import com.google.common.base.Optional;
 
 import javax.servlet.http.HttpServletRequest;
+
+import org.apache.commons.lang.ClassUtils;
+
 import java.io.IOException;
 import java.net.URI;
 import java.util.Enumeration;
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/tomakehurst/wiremock/servlet/WireMockHttpServletRequestAdapter.java
```diff
@@ -172,6 +175,25 @@ public boolean isBrowserProxyRequest() {
...
+        }
+        if (request instanceof org.mortbay.jetty.Request) {
+            return true;
+        }
+        return false;
+    }
+
+    private boolean isPresent(String className) {
+        try {
+            ClassUtils.getClass(className);
```
dsyer: Used commons-lang here (it was on the classpath anyway). I guess it could be done without that dependency if needed.

tomakehurst:
As I mentioned on the other thread, this is against master which is frozen currently.

Most of the work to isolate Jetty has already been done on the 2.0-beta branch (which is where all the current activity is).

dsyer:
Sorry, I didn't realize. Is this redundant then?

tomakehurst:
The branch pretty much has these parts covered I think. The Servlet API and Jetty bits are all in their own packages and there's an ``HttpServerFactory`` abstraction as an entry point.

The config object doesn't expose a setter for the factory yet, so that probably needs to change. I suspect creating another factory impl isn't as simple as it might be either e.g. degrading gracefully when fault injection isn't available, but it should be possible to do it I think.

dsyer:
It looks like the basic outline of the change in this PR could still apply to 2.0-beta (the servlet is not jetty-specific). Maybe I'll re-work it. Need to check my PoC still works with 2.0 first.

dsyer:
See #442

PR #442, opened 2016-06-28
Isolate jetty dependencies into a single package

dsyer:
The servlet and request adapters have moved to the "servlet" package.
The only dependency they had on jetty was conditional anyway, so
the condition was also modified to not even require jetty on the
classpath.

See gh-407. Not a complete modularization yet, but splits up the
servlet-only stuff from the jetty server.

### DIFF
--- a/sample-war/src/main/webapp/WEB-INF/web.xml
+++ b/sample-war/src/main/webapp/WEB-INF/web.xml
@@ -23,7 +23,7 @@
 
   <servlet>
   	<servlet-name>wiremock-mock-service-handler-servlet</servlet-name>
-  	<servlet-class>com.github.tomakehurst.wiremock.jetty9.JettyHandlerDispatchingServlet</servlet-class>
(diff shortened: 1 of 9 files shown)


tomakehurst:
I think it might be necessary to split out the code that builds the fault injector from ``WireMockHandlerDispatchingServlet`` as at the moment it still indirectly references Jetty classes, so I think it's unlikely to run if Jetty is taken off the classpath.

dsyer:
Interestingly it does work fine, at least for simple use cases without faults I guess. I added an extra abstraction to fix the problem in general (see last commit).

tomakehurst:
I've merged this onto a branch and changed a few bits as I discovered that it'll throw an NPE if there's no fault injector specified. There are also a couple of tests for an alternative HTTP server. See what you think:

https://github.com/tomakehurst/wiremock/tree/dsyer-feature/jetty-isolation

dsyer:
LGTM. Thanks.

PR #443, opened 2016-06-28
Fix android codec dependencies

abhagupta:
@tomakehurst  The Guava libraries work in android. I have made 2 changes that were needed to make encoding and decoding work in Android. I have tested the changes in Android project. Also I tested the changes for Desktop in wiremock standalone by running `wiremock-standalone-2.1.1-beta.jar` and testing a basic stubbing. Both worked.


abhagupta:
Checking the failure of acceptance tests.

abhagupta:
The UTF-8 characters need to be decoded differently with Guava libraries .. which will need to be looked at before this PR can be merged. Hence closing it.

Issue #445, opened 2016-06-29
The new matching system refactor has removed short circuiting behaviour


theangrydev:
As mentioned [here](https://github.com/tomakehurst/wiremock/commit/acc9871a6e1648ea7ba26b82de6970c9c836f1b2#commitcomment-18068132), the change to RequestPattern means that it is not possible to send non-XML bodies when there is at least one XML matcher configured.

This is a BIG problem for any non-trivial use of wiremock.

I suggest either putting the short circuiting behaviour back, or making sure that all matchers can handle unexpected input gracefully and say "no match" rather than blowing up. 

Perhaps the simplest thing would be to wrap all the eager match computations in a try-catch block (e.g. in MatchResult.aggregate).

tomakehurst:
Thanks for pointing this out. Are you able to build WM locally so you can test this fix on your code?

Obviously I'll release this, but it takes a little while to sync with M2.

theangrydev:
In addition to the returnsNoMatchAnd1DistanceWhenActualIsNull test, we also need a returnsNoMatchAnd1DistanceWhenActualIsNotXml

For example, what happens if the body is JSON and EqualToXmlPattern is called?

tomakehurst:
Yes, good point. Same treatment for the JSON matcher too. rc3 coming right up...

theangrydev:
That too, but what I meant was if a JSON string or any other non XML string ends up in the XML matcher.

For example I know that XMLUnit.buildControlDocument("not xml") will blow up

theangrydev:
And you will probably have similar issues the other way around too. If you are doing any JSON parsing and you pass in something that is not JSON, the parser will probably blow up

tomakehurst:
I'm writing tests covering both scenarios. The JSON matcher ones already passed as there was already a catch block in the right place. I'd just missed the equivalent in the distance calculation in the XML matcher.

theangrydev:
Great, thanks for the quick response! I'll get back to you once the new version is released and I've tried it on a real codebase

tomakehurst:
All committed and pushed. Release happening as we speak.

PR #446, opened 2016-06-29
RequestPattern should not match when there is a body pattern but no request body

theangrydev:
This solves one of the issues I have seen as a result of #445


tomakehurst:
Closing this since I put my own fix in.

PR #470, opened 2016-07-26
Update request-matching.md

bnorrish:
Added escape of '?' character ('\\') for regex matching on path and query.


tomakehurst:
Thanks

PR #479, opened 2016-08-15
Feature/near miss via junit rule

reftel:


tomakehurst:
Thanks for submitting this. Quick question - why split the tests over so many classes? I can see that you might want one for ClassRule and one for Rule, but beyond that I'm not sure I understand the value.

magnusr:
Well, `WireMockClassRule` should get to do its thing before and after each test, so each test for `WireMockClassRule` needs be a separate class, right?
But anyway, they don't, need to be separate source files. Fixed that, and rebased to master.

reftel:
@tomakehurst : Anything left to improve in this PR, or can it be merged?

tomakehurst:
Apologies for not merging yet. I wanted to review the tests a little more as I think there might be some duplication. It's on my list!

PR #488, opened 2016-08-30
Add extra information to the mock files : Record request body AS IS to binary file

sotnikdv:
BUSINESS CASE

Wiremock intended to be used as a tool to analyze request/response exchange between different parts of application.
For this we need additional information about request, like request body AS IS.

WHY THIS IS NEW FEATURE

Currently we are writing request body patterns only

IMPLEMENTATION DETAILS

New option added to trigger extra information recording.
Tests are added

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -169,9 +169,9 @@ public class WireMockServer implements Container, Stubbing, Admin {
 		stubRequestHandler.addRequestListener(listener);
 	}
 	
-	public void enableRecordMappings(FileSource mappingsFileSource, FileSource filesFileSource, boolean recordRequestBody) {
+	public void enableRecordMappings(FileSource mappingsFileSource, FileSource filesFileSource, boolean recordMappingsExtra) {
 	    addMockServiceRequestListener(
-                new StubMappingJsonRecorder(mappingsFileSource, filesFileSource, wireMockApp, options.matchingHeaders(), recordRequestBody));
+                new StubMappingJsonRecorder(mappingsFileSource, filesFileSource, wireMockApp, options.matchingHeaders(), recordMappingsExtra));
         notifier.info("Recording mappings to " + mappingsFileSource.getPath());
 	}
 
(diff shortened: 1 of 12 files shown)


tomakehurst:
I'm not sure about this. WireMock isn't really an analysis tool, so this seems to be a bit out of it's scope.

It's also quite a lot of extra code for a fairly marginal amount of value.

Is there any reason you can't just add your own ``RequestListener`` implementation that drops request bodies onto the file system?

sotnikdv:
Hi Tom,

the business requirement is to create tests with custom (manually created of modified) mocks, to cover edge cases testing on integration with other parts. For example, emulate invalid\damaged REST responses from remote service.

To make this possible we need:
- add extra information for analyze (this commit is about this feature)
- make JSON human readable (optional, by the key) and editable (pretty formatting for JSON bodies)
- make wiremock dynamically switchable between modes or check that it's safe if two instances will be pointed to the same data dir.

Since this is the real request from business; we already did very similar thing for one of major eCommerce in US, I decided that this is important feature I'd like to add to Wiremock.

As for the code, I followed the code style and test coverage, in fact, the actual change is pretty simple, it's just ~20 lines of the actual code. The rest is tests and interfaces change.

Hope that this will make sense.

Sincerely,
  Dmitry

tomakehurst:
I'm sorry, I still don't really understand what you're attempting to achieve, how this change would get you there or why this can't be done with your own listener. Perhaps a concrete example would help?

sotnikdv:
Hi Tom,

we have an application, divided onto multiple projects and communicating by the REST protocol. For example "panel" and "business processor". Wiremock may help to stub "business processor" for tests.

Nevertheless, we need to test also edge cases like "business processor" sent wrong/incorrect/damaged response and check how "panel" will handle this. We can't just record queries and then reply, we need to create stub files manually or manually change existing to emulate the case.

The easiest way is to:
- record exchange on the real system
- review requests/responses
- modify response to emulate the case

In our situation, we want to use Wiremock to run security-related test cases.

This pull request will help to get more information about the request recorded. Then we need to add pretty formatting for JSON and some information on ordering (or timestamp). Feature is switchable.

The reason why it can't be implemented as the listener - we need to conditionally (command line option) to alter mapping info, request and response to cover such cases.

I've forked Wiremock and this will be enough to make things happen, but I'd like to backport the changes, because the case is frequently used.

Sincerely,
  Dmitry

tomakehurst:
Unless I'm totally misunderstanding, WireMock can do everything you're attempting already without modification. I regularly use it for testing scenarios like the one you've described.

Since you're essentially just outputting additional diagnostic information, a listener should work just fine. At a stretch, there's nothing to stop you writing your own recorder (using the existing one as a starting point, which is itself just a listener) and decorating its output rather than altering the core model.

Sorry, but I'm going to have to pass on this one.

PR #493, opened 2016-09-12
Xpath matcher fix

senpay:
We have faced nasty issue after updating from Wiremock 1.57 to 2.x. In case there're mock mappings that has only xpath pattern and wiremock gets request with non-xml body (even for get request I think), it will generate a log record:
_[Fatal Error] :1:1: Content is not allowed in prolog._
for each xpath matching attempt.

So, lets say we have 30 mocks waiting for xpath and 5 waiting for some json, in case wiremock gets json body, it will write _[Fatal Error] :1:1: Content is not allowed in prolog._ 30 times. In our environment it produces lots of log garbage and impacts performance.

The message itself comes from attempt to build Document from non-xml thing, so I am suggesting to check with simple regexp if request body at least looks like xml.

There's a possible (although hackish) workaround to check for content-type first, but I though it would be great if we could move without it.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesXPathPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesXPathPattern.java
@@ -38,6 +38,8 @@ import static com.google.common.base.MoreObjects.firstNonNull;
 
 public class MatchesXPathPattern extends StringValuePattern {
 
+    private final static  Pattern xmlPattern = Pattern.compile("<(\\S+?)(.*?)>(.*?)</\\1>",
+                                                            Pattern.CASE_INSENSITIVE | Pattern.DOTALL );
     private final Map<String, String> xpathNamespaces;
 
     public MatchesXPathPattern(@JsonProperty("matchesXPath") String expectedValue,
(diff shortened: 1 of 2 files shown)


senpay:
@tomakehurst 
Could you please look at this pull request and either merge it or reject? We need to know what future steps we should take in order to fix the situation we face now.

tomakehurst:
I agree this is a problem worth fixing, but I'm not keen on attempting to detect whether the expectation is XML via regexes.

The log message in question is output by `com.sun.org.apache.xerces.internal.impl.XMLErrorReporter` calling `com.sun.org.apache.xerces.internal.util.DefaultErrorHandler`. It looks like there are a couple of ways this could be suppressed, either by swapping the impl of either of these, or using a config toggle.

Could you look at disabling it that way instead?

senpay:
@tomakehurst 
Tom,

Thank you for coming back with this suggestion. I also don't really like using regex for this. I will try to investigate suggested ways once have a time for that.

Cheers!

senpay:
@tomakehurst 

Cheers Tom. It doesn't seem to be easy to fix the error logging occurring in XMLUnit without changing it itself. I thought I can add a simple method that will try to build DOM from xml with custom error reporter which will suppress this error logging and then use it instead of RegExp check.

Do you think it may be an option? On the other hand, if we have DOM we may not need to use XMLUnit to evaluate XPath.

Your thoughts would be appreciated.

Regards,
Alex.

senpay:
I think I was able to fix this nicely. Check this out:

`            DocumentBuilder documentBuilder = XMLUnit.newControlParser();
            documentBuilder.setErrorHandler(new SilentErrorHanlder());
            Document inDocument = XMLUnit.buildDocument(documentBuilder, new StringReader(value));`

`public class SilentErrorHanlder implements ErrorHandler {
        @Override
        public void warning(final SAXParseException exception) throws SAXException {
            throw exception;
        }

        @Override
        public void error(final SAXParseException exception) throws SAXException {
            throw exception;
        }

        @Override
        public void fatalError(final SAXParseException exception) throws SAXException {
            throw exception;
        }
    }`

tomakehurst:
Looks great. Thanks for investigating this. Happy to merge this solution.

PR #494, opened 2016-09-12
Add PUSH and PURGE HTTP method constants

dtanner:
Apache Traffic Server and some other cache servers use the PUSH and PURGE methods.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
@@ -33,6 +33,8 @@ public class RequestMethod implements ValueMatcher<RequestMethod> {
     public static final RequestMethod HEAD = new RequestMethod("HEAD");
     public static final RequestMethod TRACE = new RequestMethod("TRACE");
     public static final RequestMethod ANY = new RequestMethod("ANY");
+    public static final RequestMethod PUSH = new RequestMethod("PUSH");
+    public static final RequestMethod PURGE = new RequestMethod("PURGE");
 
     private final String name;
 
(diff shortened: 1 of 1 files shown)


tomakehurst:
Are you aware that you can new up ``RequestMethod`` with any verb you want?

dtanner:
Yeah - here's what I'm currently doing in the tests:
`stubFor(new BasicMappingBuilder(new RequestMethod("PURGE"), urlEqualTo("/"))`
It's not a huge deal, but it feels a little dirty since I'm exceeding my access rights to the BasicMappingBuilder constructor (I'm using Groovy).  I was hoping to use a more concise helper method like:
`stubFor(purge("/"))`  

But it's not the end of the world either way - my tests are working great, so a big huge thanks for building the tool.  It doesn't look like those methods are in the spec, so I understand if you don't want to add them.  I just realized my PR didn't add that helper method.  Let me know what you think, and if you do want to add them, let me know if there are other places I should check to add.

dtanner:
p.s. And maybe I missed an available method that isn't package private, to avoid the exceeded access.?

tomakehurst:
The DSL makes non-standard methods available like this:

```java
stubFor(request("PURGE", urlEqualTo("/purge/this"))
```

I'm a bit reluctant to start adding more (less/non standard) verbs. To be consistent with the rest it'd also be necessary to add extra DSL methods and a few other bits here and there. What I'm trying to avoid is adding maintenance overhead for niche cases that are achievable via some other route.

dtanner:
Sure - I understand.  Thanks for the usage tip - I'll switch to that.

tomakehurst:
No probs

PR #502, opened 2016-09-25
Added support for multivalued cookies

garrettbts:
Have a use case where the same cookie key gets set with path=/ and path=/api

When loading /api from browser the same cookie key gets sent twice, and WireMock explodes with IllegalArgumentException when invoking the immutable map builder.build().

Should probably add a getCookiesAsList() method for multivalued cookies within the com.github.tomakehurst.wiremock.http.Request interface.


tomakehurst:
Please could you add an acceptance test case demonstrating this? Happy to merge once that's there.

Thanks.

garrettbts:
@tomakehurst Added acceptance test.

tomakehurst:
Fixed in 827e7ccb3edcb15bb78e68add74adee536e921b2

PR #505, opened 2016-09-28
Upgrade to gradle 3.1

rhart:


sta-szek:
Did you miss change in [gradle wrapper](https://github.com/tomakehurst/wiremock/blob/master/build.gradle#L405)  build.gradle to 3.1?

rhart:
Yes, my bad.  I don't tend to use that task so forgot to check if it was there.  Fixed now.

PR #513, opened 2016-10-12
In order to be able to see raw traffic from and to Wiremock for debug…

wojciechbulaty:
…ging purposes in acceptance tests I have added a NetworkTrafficListener to the ServerConnector and exposed it via an interface WiremockNetworkTrafficListener. The root cause of this was that com.github.tomakehurst.wiremock.http.RequestListener will not record things Jetty does to the request, for example appending "--gzip" to an E-TAG header if the content is gzipped, see org.eclipse.jetty.servlets.gzip.AbstractCompressedStream line 26:  setHeader("ETag",etag.substring(0,end)+"--"+_encoding+'"').

Let me know what you think Tom.


tomakehurst:
I can definitely see the usefulness of exposing this information.

I'm wondering whether it might make more sense to attach it to the existing ``LoggedRequest`` as  ``rawRequest`` or something like that?

Also, I could see myself just wanting to pipe it out to the console. Maybe worth including a logging toggle (standalone also)?

wojciechbulaty:
Thanks Tom. I might be missing something obvious here but I cannot see how I would put rawRequst in the LoggedRequest since LoggedRequest is in the Wiremock layer and the functionality I introduced has to be added in to the Jetty layer if you want to take full advantage of it. 
I would like to log what Jetty does after Wiremock has finished its job and we are back in Jetty call stack (gzip etag issue, etc.). To do that org.eclipse.jetty.io.NetworkTrafficListener has to be plugged in to the ServerConnector before starting up Jetty. 

As for the console output, good idea, I am happy to add that once we have pulled this commit into master. You have to bear in mind it will log individual bytes/chars not strings, so that means I would probably log it to a special file on the disk instead of the console. I will have a look at that.

Let me know what you think.

wojciechbulaty:
I have added a toggle to log to the console. Let me know what you think.

tomakehurst:
This looks good. Happy to merge, but would you mind adding a brief section to the docs explaining how to use it?

wojciechbulaty:
Sure. Let me know what you think.

tomakehurst:
Perfect, thanks.

PR #519, opened 2016-10-25
Fixing issue where ignoring array ordering did not work properly. 

nebevay:
This is because there is a bug in zjsonpatch, using jsonassert instead.

The downside is that distance will be smaller when subtree is missing from actual, had to remove that test as json-assert does not provide enough information in the diff to easily count size of missing subtree.
Adding a test that was failing before and a test for nested array.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -62,6 +62,7 @@ dependencies {
     }
     compile 'org.apache.commons:commons-lang3:3.4'
     compile 'com.flipkart.zjsonpatch:zjsonpatch:0.2.1'
+    compile 'org.skyscreamer:jsonassert:1.3.0'
 
     testCompile "org.hamcrest:hamcrest-all:1.3"
     testCompile("org.jmock:jmock:2.5.1") {
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToJsonPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToJsonPattern.java
@@ -17,21 +17,15 @@ package com.github.tomakehurst.wiremock.matching;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.databind.JsonNode;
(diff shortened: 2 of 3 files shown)


tomakehurst:
Have you considered fixing the zjsonpatch bug then upgrading?

Might be a better choice than working around it in wiremock.

nebevay:
Hi Tom,

First of all thanks a lot for this great library, many teams at Ocado Technology are finding it really useful for testing including mine :)

Yes I had considered fixing zjonpatch. It seems there is already a bug report for this https://github.com/flipkart-incubator/zjsonpatch/issues/18. I might add a comment to that one explaining what the issue is precisely because I've had a look and I think I know where the problem lies, however I'm not sure when it will be resolved.

Sadly I don't have enough time right now to code the fix myself, I could potentially do it at a later time. The way I see it my change just updates to a library that performs the diff correctly. The caveat is that  need to sacrifice the precision of the distance calculation. Correct me if I'm wrong but I think this is only used to say things like "request was not matched, closest match is...".  My personal opinion is that loosing some precision in that area is not a high price to pay for fixing an actual bug that will present itself even with simple arrays. The version using zjsonpatch will remain in the source code and when the issue is sorted out you can always revert back to it. But ultimately it's up to you!

Best,
Grzegorz

tomakehurst:
Let me give this a bit of thought. Obviously it's preferable to fix thing at source, but if that's going to cause too much friction, then clearly working around it in WM is better than nothing.

Thanks for the input anyway. Interesting to learn that you're using it at Ocado!

matpom:
Is it going to be merged? Is anything stopping you, that I could help you with?

tomakehurst:
If you can fix the root zjsonpatch bug and raise a PR with Flipkart that would be absolutely awesome.

Insomnium:
Any updates, guys?

phalves-bbm:
Are you planning to merge that fix?

iljapavlovs:
any updates?

tomakehurst:
Fixed now that we've moved to Json-Unit

PR #525, opened 2016-11-04
Support for providing URL pattern that need to be recorded while in r…

kkleokrish:
…ecording mode.


MasonM:
FYI: My [wiremock-snapshot](https://github.com/MasonM/wiremock-snapshot/) extension provides this functionality via the "filters" parameter.

tomakehurst:
As per Mason's comment, see #674

PR #535, opened 2016-11-23
adding silent error handler to DocumentBuilder to suppress Fatal Error Logging

senpay:
adding silent error handler to DocumentBuilder to suppress Fatal Error Logging occurring on attempt to evaluate XPath on non XML string


tomakehurst:
Thanks

Issue #545, opened 2016-12-11
Can we add the shutdown API command to the API Reference?


halfer:
The API reportedly supports `__admin/shutdown` as per the [notes here](http://wiremock.org/docs/running-standalone/) but it is not mentioned [here at all](http://wiremock.org/docs/api/). Is this an oversight, and if so can it be rectified?

Thanks!

tomakehurst:
It is, and yes it can.

halfer:
Great, thanks. If it is something a non-Java person can help with, do let me know. Or is it just a case of regenerating things in Swagger?

tomakehurst:
It's just a case of adding it to the RAML spec (from which the Swagger spec is generated). At the moment I'm the only one who can re-publish the site, but if you'd like to update the RAML, I'll happily accept a  PR.

halfer:
This has been merged, will close.

@tomakehurst, would you [regen the API docs](http://wiremock.org/docs/api/) when you get a mo? I expect it is just a button push?

PR #548, opened 2016-12-12
Add the shutdown function to the docs (#545)

halfer:
This is a suggested change to the RAML file to document the shutdown command. I am familiar with Swagger, but not RAML unfortunately, so this is untested guesswork! Happy to tweak, or feel free to use this as a base for more changes before merging.

### DIFF
--- a/src/main/resources/raml/wiremock-admin-api.raml
+++ b/src/main/resources/raml/wiremock-admin-api.raml
@@ -294,3 +294,11 @@ schemas:
     responses:
       200:
         description: Settings successfully updated
+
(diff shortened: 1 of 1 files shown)


halfer:
The build error from Travis does not look like it is related to my RAML change. Any ideas?

tomakehurst:
I re-ran it and it's green now. Problem was a flakey Maven repo.

halfer:
I forget where we were with this - can this be merged?

tomakehurst:
Sure

tomakehurst:
Thanks

PR #550, opened 2016-12-13
To make sure Wiremock is compatible with Java 1.7 I enforced the Java version

wojciechbulaty:
I know Gradle is running -source 1.7 -target 1.7 but still, it feels safeer to me to be building on Java 7 as a principle if you wanna be 100% compatible with Java 7.

Let me know what you think.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,8 +26,10 @@ apply plugin: 'idea'
 apply plugin: 'project-report'
 apply plugin: 'com.github.johnrengelman.shadow'
 
-sourceCompatibility = 1.7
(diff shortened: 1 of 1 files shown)


tomakehurst:
The build already does this if you try to release with any other JDK version:
https://github.com/tomakehurst/wiremock/blob/master/build.gradle#L316

TBH, I thought it overkill to enforce it on all compiles, and sometimes it's useful for me to build it with different versions.

wojciechbulaty:
Ok, that makes sense.

PR #553, opened 2016-12-14
When the XML DTD points to an unknown hostname then XMLUnit will get …

wojciechbulaty:
…an UnknownHostException which results in a failed match. A fix for this was to disable fetching DTDs.

Let me know what you think.


tomakehurst:
Thanks for this

PR #556, opened 2016-12-17
Added the option to define ones own mappings repository (instead of j…

wojciechbulaty:
…ust one pre-defined in-memory one). Also, added a File based repository that can be used instead of the default in-memory one.

I have not added documentation yet because I was not sure what you are going to think about this approach.

Let me know what you think.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -51,3 +51,4 @@ tmp*
 
 src/main/resources/swagger
 docs-v2/_docs/wiremock-admin-api.html
+classes
--- a/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
@@ -16,6 +16,7 @@
 package com.github.tomakehurst.wiremock.core;
 
 import com.github.tomakehurst.wiremock.common.*;
+import com.github.tomakehurst.wiremock.core.mappings.StubMappingsFactory;
 import com.github.tomakehurst.wiremock.extension.Extension;
 import com.github.tomakehurst.wiremock.http.CaseInsensitiveKey;
 import com.github.tomakehurst.wiremock.http.HttpServerFactory;
@@ -27,7 +28,6 @@ import java.util.List;
 import java.util.Map;
 
 public interface Options {
-
     int DEFAULT_PORT = 8080;
     int DYNAMIC_PORT = 0;
     int DEFAULT_CONTAINER_THREADS = 10;
@@ -52,4 +52,5 @@ public interface Options {
     HttpServerFactory httpServerFactory();
     <T extends Extension> Map<String, T> extensionsOfType(Class<T> extensionType);
     WiremockNetworkTrafficListener networkTrafficListener();
+    StubMappingsFactory stubMappingsFactory();
 }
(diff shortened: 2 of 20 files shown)


tomakehurst:
It's already possible to swap out the persistence mechanism by setting your own `MappingsSource` impl from the configuration. The default impl `JsonFileMappingsSource` is enabled by default and it's possible to make mappings persistent either by calling the save API or setting the `persistent` flag on mapping you wish to persist automatically.

Does this satisfy the use case you had in mind?

wojciechbulaty:
That makes sense. 

There is one piece that is missing for me. I need to reload a mapping when it has been changed on the disk (for example, somebody used notepad to change the file on the disk manually). 

I cannot do that now as far as I understand because I would have to replace WireMockApp.stubMappings with something that is not "in memory" but reloads from disk every time it is asked for a mapping. Right now `WireMockApp.stubMappings` is hardcoded to `InMemoryStubMappings` in the constructor.

So I had a go at it and I have made `WireMockApp.stubMappings` configurable and written my own `JsonFilesystemStubMappings`. But, I realized there was a lot of code duplicated in  `InMemoryStubMappings` and `FilesystemStubMappings` because both of those classes had to support customMatchers and transformers. 

So, I have refactored it to remove the duplication and separate the responsibilities. I extracted a new class `StubMappingsWithTransformersAndCustomMatchers` (from `InMemoryStubMappings`) that does all the `customMatchers` and `transformers` logic, and another one `SortedConcurrentSetMappingsRepository` that holds mappings in memory. I then created a `JsonFilesystemMappingsRepository` that does what the in memory one does, but on disk not in memory.

Now, after thinking about what you have mentioned in your comment, I can see a challenge, because `SortedConcurrentSetMappingsRepository` is incompatible with the `persistent` flag, it will never persist. Also, `JsonFilesystemMappingsRepository` would persist always (apart from the proxy mapping). So the two new repositories are incompatible with the `persistent` flag.

Possible outcomes:
- Thinking out loud: my first though is to deprecate the `persistent` flag?
- Another option is: I can make `WireMockApp.stubMappings` configurable and extract `customMatchers` and `transformers` logic from InMemoryStubMappings and create a `JsonFilesystemMappings` in my codebase (not in Wiremock). This is very similar to what you see in this changeset, but  `JsonFilesystemMappings` would be in my codebase, so it would not be incompatible with `persistent` flag.

Let me know if I am missing something obvious here and what you think about this reasoning.

tomakehurst:
Given the shortcomings you've pointed out, I think it'd be good to see if the existing MappingsSource abstraction could be made to solve the reloading problem.

In principle I don't think there's anything to stop a MappingsSource implementation from setting up a background thread to watch the filesystem/DB/whatever for updates to stubs and continually push them into the `StubMappings` provided to `loadMappingsInto(...)`. Although this introduces a bit of polling delay, the upside is that stubs continue to be served from memory, with all the performance gains that implies.

Having only just added the persistent flag, I'd be pretty reluctant to deprecate it. The reason for adding it was an observation that some users are running WireMock servers with a mixture of long-term and test-generated stubs. The persistent flag allows you to do a soft reset, only removing the stubs you just created in your test. This is one of the key workflows supported by MockLab.

tomakehurst:
Closing as per the discussion about making MappingsSource work a bit better.

PR #559, opened 2016-12-29
added missing dependency "uri": "^0.1.0"

soggier:
I found the uri package to be missing, when I attempted a local build.


tomakehurst:
Odd, as this works in CI. Could this be something about your node / npm setup?

PR #561, opened 2016-12-29
Ensure persistent non-null notifier for each thread

soggier:
* return a new NullNotifier as the per-thread initial value (avoids extra instances of NullNotifier)
* when the setter is called with a null reference, substitute it with a new NullNotifier


tomakehurst:
Please could you add a test case demonstrating the issue this fixes?

PR #564, opened 2017-01-07
Add additional information about file serving to the documentation

DaniRey:
I added additional information about file serving, especially when using the standard JAR. 

I love this feature for its simplicity and think more people should be aware of it.


MasonM:
Didn't know this. Very useful. 👍

PR #568, opened 2017-01-11
Byte dribble delay support

### LINKED ISSUE #472: Support response dribbling
Support releasing of the HTTP response in small chunks over a configurable time period.

Useful for testing Hystrix.

mkh25:
Allow setting configurable delay after each byte WireMock is sending in the response body.

This resolves #472

Example:

```
stubFor(get(urlEqualTo("/delayedDribble")).willReturn(
    .aResponse()
        .withStatus(200)
        .withBody(new byte[] {1, 2, 3})
        .withByteDribbleDelay(10)));
```

Removed some spurious white spaces as a separate commit because they were inconsistent within the same class file (WireMockHandlerDispatchingServlet) we were working on.

### DIFF
--- a/docs-v2/_docs/simulating-faults.md
+++ b/docs-v2/_docs/simulating-faults.md
@@ -150,6 +150,47 @@ To use, instantiate a `new UniformDistribution(15, 25)`, or via JSON:
 }
 ```
 
+## Chunked Dribble Delay
+
+In addition to fixed and random delays, you can dibble your response back in chunks. 
+This is useful for simulating a slow network and testing deterministic timeouts. 
+
+Use `#withChunkedDribbleDelay` on the stub to pass in the desired chunked response, it takes two parameters:
+
+-   numberOfChunks - how many chunks you want your response body divided up into
+-   totalDuration - the total duration you want the response to take in milliseconds
+
+```java
+stubFor(get(urlEqualTo("/chunked/delayed")).willReturn(
+        aResponse()
+                .withStatus(200)
+                .withBody("Hello world!")
+                .withChunkedDribbleDelay(5, 1000)));
+```
+
+Or set it on the `chunkedDribbleDelay` field via the JSON api:
+
+```json
+{
+    "request": {
+            "method": "GET",
+            "url": "/chunked/delayed"
+    },
+    "response": {
+            "status": 200,
+            "body": "Hello world!",
+            "chunkedDribbleDelay": {
+                    "numberOfChunks": 5,
+                    "totalDuration": 1000
+            }
+
+    }
+}
+```
+
+With the above settings the `Hello world!` response body will be broken into 
+five chunks and returned one at a time with a 200ms gap between each.  
+
 ## Bad responses
 
 It is also possible to create several kinds of corrupted responses:
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
@@ -17,13 +17,7 @@ package com.github.tomakehurst.wiremock.client;
 
 import com.github.tomakehurst.wiremock.common.Json;
 import com.github.tomakehurst.wiremock.extension.Parameters;
-import com.github.tomakehurst.wiremock.http.DelayDistribution;
-import com.github.tomakehurst.wiremock.http.Fault;
-import com.github.tomakehurst.wiremock.http.HttpHeader;
-import com.github.tomakehurst.wiremock.http.HttpHeaders;
-import com.github.tomakehurst.wiremock.http.LogNormal;
-import com.github.tomakehurst.wiremock.http.ResponseDefinition;
-import com.github.tomakehurst.wiremock.http.UniformDistribution;
+import com.github.tomakehurst.wiremock.http.*;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.Lists;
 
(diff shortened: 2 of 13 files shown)


adamdougal:
Hey, is there anything else that needs to be done for this to be merged? It'd be be a great feature to have!

tomakehurst:
There are two things about this I think might improve it:

1) Dribble the whole response, not just the body (more realistic IMO)
2) Be able to specify it in terms of the total response time + number of chunks, rather than delay between individual bytes (thus effectively supporting varying chunk sizes). Alternatively this could be framed as chunk size + inter-chunk delay.

I could live without 1) I think, as it's probably a bit harder to buffer the entire response and have everything still work correctly.

adamdougal:
@tomakehurst Hey, I've changed it to now use total response time and number of chunks, let me know what you think.

I'm not sure with the current setup if it's possible to dribble the whole response. Do you have any ideas for how to do this? Cheers

tomakehurst:
Thanks, I'll try and take a look at this over the next day or two.

The solution I had in mind for dribbling the whole response was to grab the TCP socket (like the fault implementations do), buffer the response then dribble it into the socket.

tomakehurst:
OK, I've had a little play with this. Generally looks good, but I've managed to make it blow up with a particular combination of parameters.

Here's the test case that demonstrates this: https://github.com/tomakehurst/wiremock/blob/sky-uk-byte-dribble-delay/src/test/java/com/github/tomakehurst/wiremock/ResponseDribbleAcceptanceTest.java#L73

I see this error when I run it:

```
java.lang.InterruptedException: sleep interrupted
	at java.lang.Thread.sleep(Native Method)
	at com.github.tomakehurst.wiremock.servlet.WireMockHandlerDispatchingServlet.writeAndTranslateExceptionsWithChunkedDribbleDelay(WireMockHandlerDispatchingServlet.java:195)
	at com.github.tomakehurst.wiremock.servlet.WireMockHandlerDispatchingServlet.applyResponse(WireMockHandlerDispatchingServlet.java:159)
	at com.github.tomakehurst.wiremock.servlet.WireMockHandlerDispatchingServlet$ServletHttpResponder.respond(WireMockHandlerDispatchingServlet.java:125)
	at com.github.tomakehurst.wiremock.http.AbstractRequestHandler.handle(AbstractRequestHandler.java:63)
	at com.github.tomakehurst.wiremock.servlet.WireMockHandlerDispatchingServlet.service(WireMockHandlerDispatchingServlet.java:102)
	at javax.servlet.http.HttpServlet.service(HttpServlet.java:790)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:812)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1669)
	at org.eclipse.jetty.servlets.UserAgentFilter.doFilter(UserAgentFilter.java:83)
	at org.eclipse.jetty.servlets.GzipFilter.doFilter(GzipFilter.java:365)
	at org.eclipse.jetty.servlet.ServletHandler$CachedChain.doFilter(ServletHandler.java:1652)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:585)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1127)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:515)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1061)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.HandlerCollection.handle(HandlerCollection.java:110)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:97)
	at org.eclipse.jetty.server.Server.handle(Server.java:499)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:311)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:258)
	at org.eclipse.jetty.io.AbstractConnection$2.run(AbstractConnection.java:544)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:635)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$3.run(QueuedThreadPool.java:555)
	at java.lang.Thread.run(Thread.java:745)
```

adamdougal:
Hey, @tomakehurst sorry for not getting a chance to look at this sooner.

I think what's happening is you're actually triggering a socket timeout as there are 2 chunks and the total duration is 2 x the socket timeout. 

It's not that nice that an interrupted exception is printed/returned from the Wiremock server though. I don't think there's much we could do other then suppress it/assume that it's been timeout out and log a meaningful message/throw a meaningful exception?

tomakehurst:
OK, I've tweaked my branch to ignore the `InterruptedException` rather than re-throwing it. I can't see any value in it being surfaced as an error in this particular case.

I've also found another oddity, which is that having tweaked the test case I added so that it comes in under the timeout, the total duration is coming out at half what it should be. I've pushed to my branch again if you wouldn't mind taking a look?

adamdougal:
Hmm it looks like the first test `ResponseDribbleAcceptanceTest.requestIsSuccessfulButTakesLongerThanSocketTimeoutWhenDribbleIsEnabled` is actually not really testing the duration correctly as it's only checking that it was longer then the socket timeout rather then the specified total duration.

Moving the start time to before we make the request and changing the assertion in that test to be greater then or equal to the total duration fixes it. Should I make that change or are you okay to do it in your branch?

tomakehurst:
Ah, got it. Quite a trivial change, so I'll make it on the branch.

After that I think we're all set to merge it.

Issue #546, opened 2016-12-12
New stub mapping record/playback feature


tomakehurst:
# Problem
WireMock's record feature seems to be one of the most heavily used, and as such it is quite frequently the subject of feature requests. Rather than apply lots of spot fixes for specific cases, I propose re-writing the record feature from scratch, incorporating general solutions to the types of problems expressed here and on the mailing list.

Based on past questions and requests related to the recorder, I'd like to suggest it attempts to achieve the following things:
* Record and playback without having to restart in between.
* Customise output: file naming, request/response headers/body transformation.
* Filter what gets recorded.
* Statefulness - support a succession of responses for the same request (rather than just ignoring subsequent responses as is currently the case)
* Choose target - in memory, persisted, downloaded
* Potential to import from other formats e.g. HAR, RAML (but not necessarily immediate support)

# Solution
I propose the following solution based on these objectives:

**Snapshot recording** - make some requests (that get captured in the request journal) then hit a "capture snapshot" API. This avoids the stop/start problem. It also means that optional filtering and transformation parameters could be passed into the API call, in addition to a save/download flag.

**Use Scenarios for multiple identical requests** - when more than one instance of a request is seen, capture each under one scenario so that the progression of responses seen originally is played back.

**Recording transformer extension point** - expose a Java interface that (similar to `ResponseDefinitionTransformer`) supports custom transformations of stub mappings as they're recorded.

**Recording adapter extension point** - expose a Java API to the recorder which is sufficiently generic that other formats can be adapted to it e.g. Swagger/RAML or HAR.

scho:
Would this solve #542 as well?

tomakehurst:
Yes it would

selangley:
Looks to be very compatible with the operations I proposed in my pull request last year:

https://github.com/tomakehurst/wiremock/pull/400
Added ability to record a particular request/response repeatedly

so I'm all for this solution.

sabavatbbm:
is **Use Scenarios for multiple identical requests** already implemented? How do I use it?
Also, can we have sub directories in the __files folder?

tomakehurst:
@sabavatbbm yes, scenarios are already implemented. What I'm suggesting here is that the recording mappings make use of this feature to step through multiple states. See: http://wiremock.org/docs/stateful-behaviour/

Yes, subdirectories are fine under `__files` (and `mappings` for that matter).

MasonM:
[...]

PR #573, opened 2017-01-14
I have refactored out and made public the code to add proxy mapping s…

wojciechbulaty:
…o that if you would like to create a recorder programmatically in your test suite you can do it.

Let me know what you think.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/AddProxyMapping.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/AddProxyMapping.java
@@ -1,0 +1,31 @@
+package com.github.tomakehurst.wiremock.standalone;
+
+import com.github.tomakehurst.wiremock.http.ResponseDefinition;
+import com.github.tomakehurst.wiremock.matching.RequestPattern;
+import com.github.tomakehurst.wiremock.stubbing.StubMapping;
+import com.github.tomakehurst.wiremock.stubbing.StubMappings;
+
+import static com.github.tomakehurst.wiremock.client.ResponseDefinitionBuilder.responseDefinition;
+import static com.github.tomakehurst.wiremock.client.WireMock.anyUrl;
+import static com.github.tomakehurst.wiremock.http.RequestMethod.ANY;
+import static com.github.tomakehurst.wiremock.matching.RequestPatternBuilder.newRequestPattern;
+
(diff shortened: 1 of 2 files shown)


tomakehurst:
Have you had a look at https://github.com/tomakehurst/wiremock/issues/546 ?

Rather than adding lots of piecemeal additions to record/playback I'd like to write a new feature and solve the accumulated issues that have come up over time that way.

I haven't explicitly called out the ability to record/playback from within automated tests, so I'll add that to the spec. I believe Betamax has got the right kind of model here. Adding the other stuff listed in this issue should mean it'd be fairly easy to write a rule that behaves how Betamax's does i.e. switchable via runtime config between recording and playback.

wojciechbulaty:
I understand, it does sound like a good strategic approach.

In the meantime, what do you think about having a workaround for today? This changeset is a very small refactor and would help me avoid copying and pasting code.

tomakehurst:
I'm not sure I'm totally clear on how this would be used, and why it's a better approach than a shortcut in the DSL.

Could you elaborate a bit?

wojciechbulaty:
Sure, maybe I overlooked something and this is not necessary.

When you say "a shortcut in the DSL", can you give an example?

I am trying to start up a recorder programmatically, so I am replicating what WireMockServerRunner is doing in the run method:
```java
if (options.recordMappingsEnabled()) {
        wireMockServer.enableRecordMappings(mappingsFileSource, filesFileSource);
}

if (options.specifiesProxyUrl()) {
	addProxyMapping(options.proxyUrl(), wireMockServer);
}
```

Based on the above example, what I have done in my code is:
```java
wireMockServer.enableRecordMappings(mappingsFileSource, filesFileSource);
wireMockServer.loadMappingsUsing(new AddProxyMapping(proxyUrl));
```

tomakehurst:
I suppose I'd want to write something like:

```java
stubFor(proxyAllTo("http://my.target.com"));
```

Where `proxyAllTo` is a static method on `WireMock` that configures a `MappingBuilder`.

One advantage of this is it gives me a way to delete it when I'm finished e.g.

```java
StubMapping proxyMapping = stubFor(proxyAllTo("http://my.target.com"));

// Do things...

removeStub(proxyMapping);
```

tomakehurst:
My solution is in the DSL now, so closing.

PR #577, opened 2017-01-18
In some cases you do not want Wiremock to return "Server" header if y…

wojciechbulaty:
…ou the mappings do not contain it. I have added the possibility to disable sending the server version response header.

Let me know what you think.


wojciechbulaty:
The flaky test PostServeActionExtensionTest went red, will open a pull request again for this changeset so that the builds runs again.

wojciechbulaty:
Opened #579

PR #579, opened 2017-01-18
In some cases you do not want Wiremock to return "Server" header if y…

wojciechbulaty:
…ou the mappings do not contain it. I have added the possibility to disable sending the server version response header.

Let me know what you think.


tomakehurst:
I'm tempted to just drop it completely rather than add the complexity of another option. After all it's easy enough to add back into your stubs if you need it.

Also seems highly unlikely anyone depends on it being there, so I'm not particularly worried about it being a breaking change.

tomakehurst:
Header is no longer sent

PR #586, opened 2017-01-20
In order to allow for easier debugging failed matches while running a…

wojciechbulaty:
… standalone Wiremock (by looking at the log output only) I have added detailed log messages why a given match has failed for a request

Let me know what you think.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java
@@ -41,15 +41,18 @@ public abstract class AbstractRequestHandler implements RequestHandler, RequestE
 
 	@Override
 	public void handle(Request request, HttpResponder httpResponder) {
+		if (logRequests()) {
+			notifier().info("Request received:\n" + formatRequest(request));
+		}
+
 		ServeEvent serveEvent = handleRequest(request);
 		ResponseDefinition responseDefinition = serveEvent.getResponseDefinition();
 		responseDefinition.setOriginalRequest(request);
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java
```diff
@@ -41,15 +41,18 @@ protected void afterResponseSent(ServeEvent serveEvent, Response response) {}
 
 	@Override
 	public void handle(Request request, HttpResponder httpResponder) {
+		if (logRequests()) {
+			notifier().info("Request received:\n" + formatRequest(request));
```
tomakehurst: I'm not sure breaking this off into a separate log message is a good idea. If there are a number of concurrent requests arriving there's no guarantee your request and response would be in adjacent messages. Writing it all out in a single message ensures the response matches the request preceding it.

src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
```diff
@@ -56,14 +57,41 @@
...
+                    methodMatch,
+                    headersMatch,
+                    queryParamsMatch,
+                    cookiesMatch,
+                    bodyPatternsMatch
             );
+            if (!overallMatch.isExactMatch()) {
+                notifier().info(format("Request does not match pattern:\n" +
+                                "%s\n" +
+                                "Match details: url=%s, method=%s, headers=%s, queryParams=%s, cookies=%s, body=%s",
```
tomakehurst: I wonder if it's worth going a bit further and printing the expected and actual values where there's a mismatch?

PR #609, opened 2017-02-17
Update junit-rule.md for JUnit 4.12+

jbohmann:


jbohmann:
Sorry, I submitted this on a bad assumption. I will close it.

For future reference, it seems that even with the fix to allow ClassRule and Rule together in JUnit 4.12 it still does not work given a WireMockClassRule. It will error with "must not be static" because JUnit only checks that WireMockClassRule implements MethodRule.

```
if(member.isStatic() && (isMethodRuleMember || !isClassRuleAnnotated)) {
     String message;
     if(RuleMemberValidator.isMethodRule(member)) {
         message = "must not be static.";
     } else {
         message = "must not be static or it must be annotated with @ClassRule.";
     }

     errors.add(new ValidationError(member, annotation, message));
}
```

Issue #623, opened 2017-03-05
Test failure when attempting to build latest master


hansjoachim:
(Sorry about the contentless initial post)

When attempting to build latest master (633f3cc33bc04a9b38cd503b36a3f571c7756a0d), I get the following error message due to a failing test:
```
:test

com.github.tomakehurst.wiremock.PostServeActionExtensionTest > providesServeEventWithResponseFieldPopulated FAILED
    java.lang.AssertionError: 
    Expected: is <418>
         but: was <0>
        at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:20)
        at org.junit.Assert.assertThat(Assert.java:956)
        at org.junit.Assert.assertThat(Assert.java:923)
        at com.github.tomakehurst.wiremock.PostServeActionExtensionTest.providesServeEventWithResponseFieldPopulated(PostServeActionExtensionTest.java:128)

773 tests completed, 1 failed
:test FAILED
```

I haven't looked into this test in details, but it seems to fail consistently, I'm running OpenJDK `1.8.0_121` on Ubuntu 16.10.

Smedzlatko:
I have the same issue. It seems that tests are not deterministic when runnin `./gradlew clean test
` locally. Sometimes, the build passes and sometimes it fails. I have found two tests with this non-deterministic behavior: 

- PostServeActionExtensionTest > providesServeEventWithResponseFieldPopulated
- PostServeActionExtensionTest > triggersActionWhenAppliedToAStubMapping

tomakehurst:
Thanks. I see this myself occasionally. I have a rough idea why, just haven't had a chance to fix yet.

tkruse:
Also seeing tests sometimes passing, sometimes failing, with 2.5.1 and 2.6.0. In my case multiple test methods run the same GET query one, and verify(1, ...) sometimes results in 
```
Expected exactly X requests matching the following pattern but received X + 1
```

Setup is:
```
@ClassRule
    public static final WireMockClassRule MOCK = new WireMockClassRule(wireMockConfig().dynamicPort().dynamicHttpsPort());

@Before
    public void setUp() {
        MOCK.resetAll();
    }
```

tomakehurst:
@tkruse  - this doesn't look like the same issue as the one described above. Would you mind opening a new one and posting your full test code (or a test case the reliably reproduces the problem)?

tomakehurst:
Not seen this failure for quite a while so I'm assuming it's gone

PR #624, opened 2017-03-05
Upgrade to latest version of zjsonpatch

hansjoachim:
When checking some projects with DependencyCheck, I discovered that wiremock has a vulnerable dependency:

```
One or more dependencies were identified with known vulnerabilities:

commons-collections4-4.0.jar (cpe:/a:apache:commons_collections:4.0, org.apache.commons:commons-collections4:4.0) : CVE-2015-6420
```

(See http://commons.apache.org/proper/commons-collections/release_4_1.html or https://issues.apache.org/jira/browse/COLLECTIONS-580 for more details.)

Turns out this is dependency is added via zjsonpatch, which has just released a new version which is no longer depending on the vulnerable verison of commons-collections4 (https://github.com/flipkart-incubator/zjsonpatch/releases/tag/0.3.0)

Due to #623 I haven't been able to build wiremock successfully, but I didn't see any worse results after changing the version number.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,7 @@ dependencies {
         exclude group: "org.hamcrest", module: "hamcrest-core"
     }
     compile 'org.apache.commons:commons-lang3:3.4'
-    compile 'com.flipkart.zjsonpatch:zjsonpatch:0.2.1'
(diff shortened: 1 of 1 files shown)


tomakehurst:
Thanks

Issue #620, opened 2017-03-02
Jetty HTTP ERROR: 500 with a GET parameter and subsequent requests


Joe-Klauza:
I've been checking out [Couchbase](https://www.couchbase.com/)'s REST API and would like to record and replay some interactions using WireMock. Unfortunately I've encountered an issue where a GET request with a parameter being made multiple times (immediately following the previous) causes Jetty to return a 500 error.

This is the URL I'm `GET`ting, proxied to couchbase via WireMock: `http://localhost:8442/pools/default?waitChange=0`

I'm mostly using the [Insomnia](https://insomnia.rest/) REST client to send the request. I've determined that the following do not affect the result:
- HTTP request headers
- HTTP request auth
- REST client
- `--container-threads` amount
- `--jetty-acceptor-threads` amount
- `--jetty-accept-queue-size` amount

In my tests I've found that removing the parameter `?waitChange=0` does not reproduce the issue (multiple, subsequent requests can be made as quickly as possible with a valid response). I've also found that trying a test parameter (`?test=0`) does not reproduce the issue.

The reason I think WireMock is the culprit (as opposed to Couchbase's REST implementation) is that I can make the same requests directly to Couchbase and get a successful response every time.

Here are the logs and the startup command used to run the server; the problem URI was hit twice for this test, with the second attempt failing:
```
java -jar /path/to/wiremock-standalone-2.5.1.jar --container-threads 200 --print-all-network-traffic --verbose --https-port 8443 --port 8442 --root-dir /path/to/wiremock-recordings/test-record --proxy-all="http://<my-couchbase-host>:8091" --record-mappings --match-headers="Accept,Content-Type,Cache-Control,Authorization"

2017-03-02 11:11:30.149 Verbose logging enabled
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See http://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
2017-03-02 11:11:30.455 Verbose logging enabled
2017-03-02 11:11:30.473 Verbose logging enabled
2017-03-02 11:11:30.512 Recording mappings to /path/to/wiremock-recordings/test-record/mappings
 /$$      /$$ /$$                     /$$      /$$                     /$$
| $$  /$ | $$|__/                    | $$$    /$$$                    | $$
[...]

PR #628, opened 2017-03-10
Add default request config check stale conn

Joe-Klauza:


Joe-Klauza:
Solves Issue #620 , but it might cause some issues for others.

tomakehurst:
Sorry for the delay in getting back to you. I'm fine with the actual code change, but the test case is a bit of a mess.

I get that this is tricky to test properly, but can you think of a way of putting a proper test for this into the suite?

Joe-Klauza:
I agree the test case is a mess; sorry about that. Unfortunately we don't know of a way to reproduce this issue outside of Couchbase, so spinning up a local instance and performing the API calls are necessary steps to reproduce it. As such I'm not really sure how to simplify the test other than removing some of the extraneous logging or perhaps unused imports... I'm okay with removing the test entirely, but like I said I'm not sure if setting the stale connection check would cause issues for others.

tomakehurst:
I can think of some ways it could be done but they're totally overkill. In this case I think just removing the exiting test and living without a proper one is OK.

Joe-Klauza:
Should be all set, @tomakehurst. Thank you!

PR #635, opened 2017-03-28
Check stubMappings before record again

jose1984:
When the record-mapping option is enabled, after restarting the WireMock server or resetting its requests. The first requests that were previously stored would be re-registered, resulting in duplicate mappings. With these changes, we prevent this from happening.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
@@ -15,6 +15,7 @@
  */
 package com.github.tomakehurst.wiremock.stubbing;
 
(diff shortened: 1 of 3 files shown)


tomakehurst:
Could you add a test case that clearly articulates what this is doing?

jose1984:
Of course, the test case is already pushed but I needed to use an external URL to storage the JSON mapping file to check that it does not duplicate after a second client request. So feel you free to suggest me another manner to reproduce it, please.

tomakehurst:
Can I suggest setting up a second WireMock instance as the proxy target as the other record/playback tests do?

I'd prefer not to introduce external dependencies to the unit test suite.

jose1984:
Thank you for your tips! I hope to clarify it with the test case

tomakehurst:
Closing this as there's a new recorder in the works: #674

PR #641, opened 2017-04-02
Add response templating for proxyBaseUrl

MasonM:
This extends the response templating extension to work with proxy URLs. This is very useful when
using the proxy/intercept setup with multiple API providers, especially when the
endpoints aren't known ahead of time. The browser proxy mode is perhaps better suited to this use
case, but it's not an option for HTTPS-only APIs due to the lack of SSL support.

Specific use case: I'm working with a large legacy codebase that has many API tests that I want to stub with Wiremock for performance reasons. I don't know all the API endpoints involved, but I do have the ability to tamper with the request before it's sent. The example I added to the docs is what I want to do: set a header named `X-WM-Proxy-Url` with the API endpoint and template it with `"proxyBaseUrl": "{{request.headers.X-WM-Proxy-Url}}"`.

### DIFF
--- a/docs-v2/_docs/response-templating.md
+++ b/docs-v2/_docs/response-templating.md
@@ -5,7 +5,7 @@ toc_rank: 71
 description: Generating dynamic responses using Handlebars templates
 ---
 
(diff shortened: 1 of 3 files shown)


tomakehurst:
Looks good, thanks.

PR #646, opened 2017-04-10
Suggested API doc change to fix #581

### LINKED ISSUE #581: Documentation does not describe that you can delete requests ?
It is possible to send a http-DELETE to /__admin/requests which will delete all requests. Can you add this to the admin document section please?

halfer:
Not tested, but looks OK to me. I cribbed the response stuff from other delete sections, let me know if any tweaks are required.


tomakehurst:
Thanks

PR #656, opened 2017-04-24
Added hook for extending the template model to ResponseTemplateTransformer

henriropp:


tomakehurst:
Can you describe the use case for this?

Since, as it stands, the additional model data is fixed at the global scope (i.e. same for every request/response and fixed at startup time) I can't see how this provides much of an advantage over hard-coding the values into the stub data directly.

henriropp:
What we are trying to achieve is that for many test scenarios we need to create a response where only a few parameters differ from the other scenarios. To avoid having a lot of almost similar response files, which all need to be modified if the responsible backend system changes it response, I wanted to create template files that could be parameterized.

Currently I am using it to pass the "transformerParameters" from a mapping file so they can be used in a template.

Part of mapping file:

```
"transformers": ["my-response-template-transformer"],
"transformerParameters" : {
   "groupId" : "1234567890",
   "companyName" : "My Company AB",
   "ids": ["10001","10002"]
   }

```
Part of template file:
```
<GroupId>{{parameters.groupId}}</GroupId>
<Name>{{parameters.companyName}}</Name>
<SetOfIds>
  <Id>{{parameters.ids.[0]}}</Id>
</SetOfIds>
<SetOfIds>
  <Id>{{parameters.ids.[1]}}</Id>
</SetOfIds>

```
And the glue between:
```
public class MyResponseTemplateTransformer extends ResponseTemplateTransformer {
   ...

    @Override
    protected void templateModelHook(ImmutableMap.Builder<String, TemplateModel> modelBuilder, Parameters parameters) {
        final MapTemplateModel model = new MapTemplateModel();
        model.putAll(parameters);

        modelBuilder.put("parameters", model);
    }

    public class MapTemplateModel extends HashMap<String, Object> implements TemplateModel { }
```

Another possible use case would be to have a set of preconfigured setups/models (perhaps load from a yaml file) from which one is selected using transformerParameters. The same model could be utilized by different response template files, which only use the parts that are necessary for each template file. 
e.g. 

```
companyOne:
  groupId: 123456
  companyName: Company One
  ids: 1,2,3,4
  user: John
  address: Street 1
  
companyTwo:
  groupId: 654321
  companyName: Company Two
  ids: 5,6,7
  user: Joe
  address: Street 2

```

tomakehurst:
OK, I think this makes sense. I hadn't clocked that the template method you added was a function of the `Parameters` object meaning it can change per-stub. I can see the use for that.

Can I request a few changes?

* Lose the marker interface. Not sure that really serves any purpose.
* Make the default implementation of `templateModelHook` add a `parameters` key to the model, containing the `Parameters` contents (like you say you're doing in the comment thread).
* Rename `templateModelHook` to `addExtraModelElements` for naming tone consistency.

Apologies for the very delayed reply.

Issue #632, opened 2017-03-17
equalToJson(..) with ignoreExtraElements = false should not accept empty arrays or empty objects


zuckel:
### Problem
If I have an expectation of the following form:
```java
verify(anyRequestedFor(anyUrl()).withRequestBody(equalToJson(
    "{'client':'AAA','name':'BBB'}".replace('\'', '"'), false, false)));
```
it will actually accept the following request body:
```javascript
{"client":"AAA", "name":"BBB", "addresses": [ ]}
```

This makes it more difficult than necessary to test if an HTTP client is behaving correctly. My expectation would be that this verification fails because there is a third field "addresses" and I told it to _not_ ignore extra elements.

### Analysis
I think the problem lies in com.github.tomakehurst.wiremock.common.Json#deepSize. It will not count the outmost container as you can see from this passing test.
```java
@Test
public void testDeepSize() throws Exception {
    ArrayNode array0 = new ArrayNode(JsonNodeFactory.instance);
    assertThat(Json.deepSize(array0), is(0));

    ArrayNode arrayBoth = new ArrayNode(JsonNodeFactory.instance).add(array0).add(array0).add(array0);
    assertThat(Json.deepSize(arrayBoth), is(3));
}                                                            
```

### Possible fix:
Start with acc = 1 for the current node. Let child nodes count themselves and call deepSize(child) also for non containers. Sorry for not providing a PR this time.
```java
public static int deepSize(JsonNode node) {
    if (node == null) {
        return 0;
    }
    int acc = 1;
    if (node.isContainerNode()) {
        for (JsonNode child : node) {
            acc += deepSize(child);
        }
    }
    return acc;
}
```

zuckel:
Hi Tom, in the second commit I added two semantic tests for the problem as described in this issue. The PR is now done from my side.

tomakehurst:
@zuckel sorry it's taken so long to look at this, and that I didn't use your patch (I TDD'd my way to the same solution in the end)

PR #660, opened 2017-05-03
fix deepSize so that empty arrays and objects have a size of 1 #632

zuckel:
Some tests had to be adapted wrt. expected sizes or expected distances, but all of these made sense to me.


zuckel:
Hi Tom, I rebased this to your current master.

PR #674, opened 2017-05-30
Issue #546 - New API endpoint for recording stub mappings

MasonM:
[As promised](https://groups.google.com/forum/#!topic/wiremock-user/AxH20Ea-m-U), here's a PR that integrates my [wiremock-snapshot](https://github.com/MasonM/wiremock-snapshot/) extension. It adds a  new API endpoint, `/__admin/snapshot`, for recording stub mappings from the request journal. [The README](https://github.com/MasonM/wiremock-snapshot/) has more details, but here's a summary of the options it accepts:

* `"filters"` - Request patterns to use for determining the requests for which to create stub mappings.
  * Possible values: Identical to `/__admin/requests/find`
  * Default: no filtering.
* `"persist"` - If set to true, persist stub mappings to disk. Otherwise, just output them.
  * Possible values: true, false
  * Default: true
* `"sortFields"` - Array of fields in the request to use for sorting stub mappings, mainly for output.
  * Possible values: "url", "method", or a header name (e.g. "Accept")
  * Default: no sorting.
* `"captureHeaders"` - Header matchers for including headers in the stub mapping. The request is matched against each matcher, and the associated header is added to the stub mapping if there's a match.
  * Possible values: Request matchers for headers.
  * Default: none
* `"outputFormat"` - Determines response body.
  * Possible values: "ids" to return array of stub mapping IDs, "full" to return array of stub mapping objects
  * Default: "ids"

# Design issues
## Filtering
Filtering is currently request-only. I'm not sure if response filtering is desired. If so, we'd probably want to add a new `ResponsePattern` class (analogous to `ResponsePattern`) to encapsulate matchers for the response. This would be best accompanied with some refactoring to make `Response` into a an interface implemented by `LoggedResponse` .

## Body file extraction/naming
Currently, the API does not extract the response bodies to a separate file. While it'd be easy to add a flag to enable this, I think more granularity is needed. Body file extraction is useful for large binary responses, but generally undesirable for small JSON responses. Perhaps we can add a `"bodyFileCritera"` option consisting of matchers that determine whether or not to extract the body. That way, you could use `"headers": { "Content-Type": { "doesNotMatch": ".*json.*" } } }` to only extract non-JSON responses. Again, we'll probably want to add a `ResponsePattern` class for this, and a `"bodySize"` matcher to allow matching by the size of the body.

## Uniqueness/Deduplication
To ensure stub mappings are unique, I have `StubMappingTransformer#apply()` generate a MD5 hash of the `RequestPattern` JSON and set that as the UUID. This is pretty hacky, but it seems to work well. I considered just using `hashCode()`, but that's not appropriate since it has a high collision probability and the result is not guaranteed to be consistent between executions (see [the JavaDoc](https://docs.oracle.com/javase/7/docs/api/java/lang/Object.html#hashCode())).

This makes me wonder why StubMapping IDs are random instead of being a content-based hash. I see this was added in PR #306, but it doesn't seem uniqueness was considered.

## Outputting
The options available in `"outputFormat"` seem sufficient to me. If we add body file extraction, we may want to add another output format to serve everything in a zip file.

## Sorting
I implemented the ability to sort the stub mappings by the request URL, method, or a header. The main use case for this is to make it easier to manually edit the stub mappings. This is arguably better served by extension points, so it might be better to just remove this feature.

The class used for sorting, `RequestFieldsComparator`, is pretty hacky and a violation of the open-closed principle, but I couldn't think of a better implementation. 

## Extension points
An extension point for modifying `StubMapping` objects would be easy to add. Specifically, we could have `WireMockApp#buildAdminRequestHandler()` pass in the stub mapping extensions along with the API extensions to `AdminRoutes`, which could then call

    router.add(POST, "/snapshot", new SnapshotTask(this.stubMappingTransformerExtensions));

instead of 

    router.add(POST, "/snapshot", SnapshotTask.class);

Adding the ability to customize the file names for persisted mappings would be more complex. Perhaps we could use a wrapper class, e.g. `RecordedStubMapping`, that would allow extensions to set the name for the mapping file and (if applicable) the body file.

## Statefulness/Scenarios
It'd be easy to detect duplicate requests and then set the `StubMapping` scenario name, but I'm not quite sure about naming. We could just use something like `UniqueFilenameGenerator` to auto-generate the scenario name, but we'll probably want to have some way of customizing it. Maybe a `"scenarioNameTemplate"` option that accepts a Handlebars template?

edit: Minor clarifications


tomakehurst:
Hi, thanks for raising this, and apologies for taking this long to respond.

I'll address the points you've made in turn:

### Filters
Looks good. Request filtering is virtually free to add so it would seem silly not to include it, but I don't think it's worth attempting to implement response filtering at this point. It's not something I've had any feedback to indicate would be useful.

One additional thing I think is essential here: being able to filter by specifying a list of request IDs.

### Persist parameter
Also looks good.

### Sort fields
Not sure about this. Given that stub insertion order affects precedence i.e. all other things being equal a more recently added stub mapping will be preferred, could sorting have the effect of muddying the water? Also, as per your comment further down, I'm wondering whether adding some kind of configurable/pluggable naming strategy might be better.

### Capture headers
Looks good. I'm wondering whether there's a case for specifying the operator also. I can think of a few cases where `contains` might be better than `equalTo` e.g. the `Accept` header at the time of recording is `application/json` but then changes to `application/json; q=0.9,*/*`.

### Output format
I'd be tempted to default to full here, for consistency with the rest of the API. Given my earlier comment about filtering by ID, it would be great to be able to see the mapping of request ID -> stub mapping ID here.

### Body file extraction
The configuration options I'd probably want here are:

* Size threshold for emitting a body file for binary content 
* Size threshold for emitting a body file for text content

In both cases omitting the parameter means don't write body files for any size.

WireMock already contains some code that can reasonably reliably tell binary from text content. Might be worth reusing that.

### Uniqueness/Deduplication
Is there any reason `StubMapping.equals()` won't work here?

The reason stub IDs are generated randomly is that they're meant to be persistent (serving as a primary key). If a stub is edited, the ID would have to change, making it a harder job to keep track of it over time.

### Outputting
Agree zip output would be nice to have. Perhaps not important enough to require it in the first release though.

### Extension points
An extension point to transform output is really important IMO, especially as it's quite easy to implement.

Regarding file naming, perhaps the answer is that we move to setting the file name to whatever is in the stub's `name` field. That way the transformer can edit that if the filename needs to be altered. More generally, keeping the two the same would make it easier from a user perspective to trace between files and mappings.

### Scenarios
Agree using `UniqueFilenameGenerator` would make sense. Perhaps the scenario gets the name generated by that, and the individual stubs get that plus a sequential number, which would also be the scenario step name.

Also agree it should be customisable. Perhaps given the previous point, the extension should have two methods that can be overridden - one for stub name, one for scenario name.

MasonM:
> One additional thing I think is essential here: being able to filter by specifying a list of request IDs.

Yeah, that'd be useful. Added in 97a767f8cc834ed39499c81af419c87e6ff62241

> Given that stub insertion order affects precedence i.e. all other things being equal a more recently added stub mapping will be preferred, could sorting have the effect of muddying the water? 

I agree, this could be confusing. I removed it in 18d44f5af1c7a3f18a7c177a8b433d16b7dbcdc6

> I'd be tempted to default to full here, for consistency with the rest of the API. Given my earlier comment about filtering by ID, it would be great to be able to see the mapping of request ID -> stub mapping ID here.

You're right, "full" is more consistent. I changed it to default to "full" in 47f4adcacf9afb8e65d091abd208efb2b46b69de. Having that request ID mappings does sound useful; I'll do that this weekend.

> If a stub is edited, the ID would have to change, making it a harder job to keep track of it over time.

Oh, I didn't consider the case of editing a mapping. Yes, that means a surrogate key (in RDBMS terms) makes sense. One of the reasons I had it generate a unique hash is because I figured calling `admin.getStubMapping(stubMapping.getId())` would be a lot faster than `admin.countRequestsMatching(stubMapping.getRequest()`, but looking at `InMemoryRequestJournal` it seems they're both O(n). I changed it to use `admin.countRequestsMatching()` in 7e49cfb21d48d7767ecb186131664d98d5a03efb

>Agree using UniqueFilenameGenerator would make sense. Perhaps the scenario gets the name generated by that, and the individual stubs get that plus a sequential number, which would also be the scenario step name.

That sounds like a good plan for now. I'll work on that this weekend.

For body file extraction and extensions, I entered PR #682 to facilitate that.

tomakehurst:
This is all looking great so far.

I have a couple of small requests:
1) Could we make the route map to `/recordings/snapshot`?
The reason is that I'd like to further extend this at some stage to provide the option of providing a start/stop recording mechanism via `/recordings/start-recording` and `/recordings/stop-recording`. In practice this can probably just track the serve event IDs seen in the time interval between starting and stopping then snapshot on stop, but since this is what most people are used to I still think it aids usability.

2) Could we also have an acceptance test or two?

Again, thanks for all the effort you've put into this!

MasonM:
I've finished adding basic scenario support in 6eaffd841c2b11be7539a7ff8e7c9c0bd515a671. This can be enabled using the new "requestsAsScenarios" flag. The logic for generating the scenario name is at https://github.com/tomakehurst/wiremock/commit/6eaffd841c2b11be7539a7ff8e7c9c0bd515a671#diff-1513fe9344f15d8e88447a9dc21831cfR51

The following acceptance test shows how it works: https://github.com/tomakehurst/wiremock/pull/674/commits/6eaffd841c2b11be7539a7ff8e7c9c0bd515a671#diff-2735104c468d56aed57eb1adbdd3f56eR282

I need to write tests for `SnapshotStubMappingScenarioHandler`, but I think I'll wait until we iron out the details for that.

> Could we make the route map to /recordings/snapshot

Done in ddfbf57

> Could we also have an acceptance test or two?

That's kinda what the `SnapshotTaskTest` class is meant for. It looks like the only difference between those tests and the other Wiremock acceptance tests is that the former instantiate `SnapshotTask` directly, while the latter are full end-to-end tests. I'll go ahead and change `SnapshotTaskTest` to be end-to-end so they're consistent.

MasonM:
I added a bunch of acceptance tests in 194b85142eeaeca4738e43ed769d808741e244bd that were split off from SnapshotTaskTest. They uncovered a few bugs, so that was a good call. I also wrote a shell script at https://github.com/MasonM/wiremock-snapshot/blob/master/test-script.sh to make testing at the command line easier.

I ended up rewriting the whole scenario handling logic because I realized my original implementation was pretty bad. Hopefully this one is a little better.

The extension point for transforming stub mappings is in 44634c5f072ecc5c1bf11de530dd3dd5e38c4097. I added two new options, "transformers" and "transformerParameters", to the endpoint for consistency with response definition transformers. The StubMappingTransformer transformers behace pretty much the  same as response definition transformers.

I'll do body file extraction next.

tomakehurst:
Wow, again this is prolific stuff!

I'll try and review everything you've done by e/o tomorrow at the latest.

MasonM:
FYI: The body file extraction code in 516095a and ce49104 is a WIP. I entered PR #690 to fix an issue I ran into where binary bodies were getting corrupted.

MasonM:
@tomakehurst I should've asked this earlier, but would it help if I rebased these commits into a logical progression, starting with a bare-bones snapshot endpoint and adding on one feature per commit? That seems like the best way of making this easier to review.

src/main/java/com/github/tomakehurst/wiremock/admin/model/SnapshotSpec.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Encapsulates options for generating and outputting StubMappings
+ */
+public class SnapshotSpec {
+    // Whitelist requests to generate StubMappings for
+    private ServeEventRequestFilters filters;
```
MasonM: Sure, done in 331ba0a148ffd8e75f061c788c21ee692b36dd14

src/main/java/com/github/tomakehurst/wiremock/admin/model/SnapshotSpec.java
```diff
@@ -0,0 +1,72 @@
...
+    // Parameters for stub mapping transformers
+    private Parameters transformerParameters;
+
+    @JsonCreator
+    public SnapshotSpec(
+        @JsonProperty("filters") ServeEventRequestFilters filters ,
+        @JsonProperty("captureHeaders") Map<String, MultiValuePattern> captureHeaders,
+        @JsonProperty("extractBodyCriteria") ResponseDefinitionBodyMatcher extractBodyCriteria,
+        @JsonProperty("outputFormat") SnapshotOutputFormat outputFormat,
+        @JsonProperty("persist") JsonNode persistNode,
```
MasonM: You're right, it'd be better if this wasn't coupled to Jackson. Fixed in 331ba0a148ffd8e75f061c788c21ee692b36dd14

src/test/java/com/github/tomakehurst/wiremock/SnapshotAcceptanceTest.java
```diff
@@ -0,0 +1,274 @@
...
+        proxyingTestClient = new WireMockTestClient(proxyingService.port());
+        wireMockServer.stubFor(any(anyUrl()).willReturn(ok()));
+    }
+
+    @After
+    public void proxyServerShutdown() {
+        proxyingService.stop();
+    }
+
+    private static final String DEFAULT_SNAPSHOT_RESPONSE =
```
MasonM: To be clear, you mean having the stub mapping ID as the key and the actual mapping as the value, right? Example:
```json
{
   "mappings" : {
      "365d3bc2-a6c8-49f1-8d59-f6040f44c82b" : {
         "id" : "365d3bc2-a6c8-49f1-8d59-f6040f44c82b",
         "request" : {
            "url" : "/",
            "method" : "GET"
         },
         "response" : { 
            "status" : 200
         },
         "uuid" : "365d3bc2-a6c8-49f1-8d59-f6040f44c82b"
      },
      "aa220624-7f87-4f79-9688-e6ad66117e2f" : {
         "id" : "aa220624-7f87-4f79-9688-e6ad66117e2f",
         "request" : {
            "url" : "/foo",
            "method" : "GET"
         },
         "response" : {
            "status" : 200
         },
         "uuid" : "aa220624-7f87-4f79-9688-e6ad66117e2f"
      }
   }    
}

```

src/test/java/com/github/tomakehurst/wiremock/SnapshotAcceptanceTest.java
```diff
@@ -0,0 +1,274 @@
...
+            "[                                                           \n" +
```
MasonM: Done in fc66370a402947ebf4c0d75057f75b0cb42459b9. For the "full" output format, it wraps the array in a "mappings" key, and for the "ids" output format it wraps it in a "ids" key.

src/test/java/com/github/tomakehurst/wiremock/SnapshotAcceptanceTest.java
```diff
@@ -0,0 +1,274 @@
...
+            "    }                                                       \n" +
+            " ]                                                            ";
+
+    @Test
+    public void returnsRequestsWithDefaultOptions() throws Exception {
+        proxyingTestClient.get("/foo/bar", withHeader("A", "B"));
+        proxyingTestClient.get("/foo/bar/baz", withHeader("A", "B"));
+
+        assertThat(
+            proxyingTestClient.snapshot("{ \"persist\": false }"),
```
MasonM: It will persist the stub mappings to disk, since it persists by default. If there were an acceptance test for that, then there'd have to be logic to delete the saved mappings after the test in order to guarantee test isolation. I noticed that `AcceptanceTestBase` has a `setupServerWithTempFileRoot()` method for that, but I can't use it here because I want to apply it to the `proxyingService` instance, not the backend WiremockServer. Also, it doesn't seem to ensure the files are deleted at the end of each test. Seems to me the easiest fix would be to change `setupServerWithTempFileRoot()` to just return the temporary file root, so it'd be renamed `createTempFileRoot()`, and have a `@After` hook to delete it.

`SnapshotTaskTest` has some a unit test for persistence (`persistsStubMappingWhenPersistSet()`), so at least that's covered.

src/main/java/com/github/tomakehurst/wiremock/http/LoggedResponse.java
```diff
@@ -7,14 +7,17 @@
     private final int status;
     private final HttpHeaders headers;
     private final String body;
+    private final byte[] bodyBytes;
```
MasonM: This is a hack to fix the binary body issues until PR #690 is merged

src/main/java/com/github/tomakehurst/wiremock/admin/model/SnapshotStubMappingPostProcessor.java
```diff
@@ -0,0 +1,49 @@
...
+
+import com.github.tomakehurst.wiremock.stubbing.StubMapping;
+import com.google.common.collect.Lists;
+
+import java.util.List;
+
+/**
+ * Transform stub mappings using any applicable StubMappingTransformers and extract response body when applicable
+ */
+public class SnapshotStubMappingPostProcessor {
```
MasonM: I haven't added unit tests for this class yet. I'm going to wait until we have the details ironed out first.

src/test/java/com/github/tomakehurst/wiremock/SnapshotAcceptanceTest.java
MasonM: Also, one issue with this change is that the order of the stub mappings will be lost, because JSON objects are unordered. This might be an issue is someone wants to do post-processing on the returned mappings in the order the request was received. Probably not a significant concern, but it's something to consider.

MasonM:
p.s. Let me know if there's anything I can do to make this or PR #690 easier to review

tomakehurst:
Sure. I'll be devoting a lot more time to this from Monday (no more client work, woo!), so expect to see some comments then.

src/main/java/com/github/tomakehurst/wiremock/admin/model/SnapshotRepeatedRequestHandler.java
```diff
@@ -0,0 +1,84 @@
...
+ */
+public class SnapshotRepeatedRequestHandler {
+    private final static String SCENARIO_NAME_PREFIX = "scenario";
+    private final boolean shouldRecordRepeatsAsScenarios;
+
+    public SnapshotRepeatedRequestHandler(boolean shouldRecordRepeatsAsScenarios) {
+        this.shouldRecordRepeatsAsScenarios = shouldRecordRepeatsAsScenarios;
+    }
+
+    public void filterOrCreateScenarios(Iterable<StubMapping> stubMappings) {
```
MasonM: The reason this doesn't use `filter()` is because it's stateful (it has to keep track of the previously-seen requests), and predicates are meant to be free of side effects.

src/main/java/com/github/tomakehurst/wiremock/admin/model/SnapshotStubMappingBodyExtractor.java
```diff
@@ -0,0 +1,62 @@
+package com.github.tomakehurst.wiremock.admin.model;
+
+import com.github.tomakehurst.wiremock.client.ResponseDefinitionBuilder;
+import com.github.tomakehurst.wiremock.common.*;
+import com.github.tomakehurst.wiremock.core.WireMockApp;
+import com.github.tomakehurst.wiremock.stubbing.StubMapping;
+
+public class SnapshotStubMappingBodyExtractor {
```
MasonM: This is a WIP since it's dependent on PR #690

tomakehurst:
I tried sorting out those two merge conflicts, but doing so creates a bunch of compile errors elsewhere. Would you mind taking a look?

src/test/java/com/github/tomakehurst/wiremock/SnapshotAcceptanceTest.java
tomakehurst: That should be fixable just by adding this to the After method:

```java
 proxyingService.resetMappings();
```

That will delete persistent mappings (not obvious from the name, I know!).

MasonM: Okay, that worked for deleting the mapping files themselves. It doesn't delete the temporary directories, though, which means a bunch of `/tmp/wiremock-<something>` directories will persist after the test.

That's already happening with `RemoteMappingsLoaderAcceptanceTest`, so I assume it's not a big issue. I went ahead and changed `SnapshotAcceptanceTest` to use a temporary file directory and updated `returnsRequestsWithDefaultOptions()` to check persistence in da6e0e721cce0e1b807e68db8747f55379e516f2

MasonM:
Merge conflicts resolved in 996d723. I also did yet more refactoring for scenario handling/post-processing in 901db8d

tomakehurst:
Great, thanks for this.

I've been spending a bit of time reviewing the original spec, plus past bug tickets etc. and I've compiled a list of what I think we have left to do:

1. Support start recording/stop recording via the API (I think this could be a very simple case of bookmarking the start and end logged requests then just snapshotting as usual)
1. Add all new API calls to the DSL
1. Support truncation of generated filenames to a maximum length (there are a few bug tickets around complaining about very long URLs generating filenames the OS can't handle)
1. Set the name field of recorded stubs to be the same as the filename (without the extension)
1. Add the ability to configure the ignore* flags when JSON body matching

Is there anything you think I've missed, or that you'd like to see in there?

I'm intending to work on 1 and 2, which I think are the biggest but also (hopefully) the least conflicting with what you're doing.

I'll push any changes I make to my remote branch and then post a comment on here so that you can pull them into your branch, if that suits?

tomakehurst:
Two more things:

1. Adding IDs to generated stub mappings (so that they can be addressed later, as discussed)
2. Documentation (I'll start working on this)

src/test/java/com/github/tomakehurst/wiremock/SnapshotAcceptanceTest.java
tomakehurst: OK, I think we can live with that for now, and improve on it if anyone complains.

src/main/java/com/github/tomakehurst/wiremock/admin/model/RequestPatternTransformer.java
```diff
@@ -0,0 +1,72 @@
...
+     */
+    @Override
+    public RequestPatternBuilder apply(Request request) {
+        final RequestPatternBuilder builder = new RequestPatternBuilder(request.getMethod(), urlEqualTo(request.getUrl()));
+
+        if (headers != null && !headers.isEmpty()) {
+            for (Map.Entry<String, MultiValuePattern> header : headers.entrySet()) {
+                String headerName = header.getKey();
+                MultiValuePattern matcher = header.getValue();
+                if (matcher.match(request.header(headerName)).isExactMatch()) {
```
tomakehurst: This behaviour is a bit of a departure from what we do currently e.g. use the value from the incoming request as the thing to match on, rather than a pre-determined value. I think this is probably more useful e.g. if you're recording an API the deals in more than one data content type then the recorder will understand that it needs to record different stubs for each `Accept` header value.

In the current recorder the operator is hard-coded to `equalTo`. What I had in mind here was being able to choose the operator, but still keep the right-hand operand from the request.

MasonM:
@tomakehurst Yeah, that sounds like a good list. I'm fine handling 3-5 and changing the "full" output format to match the example I gave in https://github.com/tomakehurst/wiremock/pull/674#discussion_r125558910 (almost done, just need to update the acceptance tests). I sent an invite to give you write access to https://github.com/MasonM/wiremock/, so you're free to just push to this branch, otherwise I can just merge in your changes.

tomakehurst:
Actually, I think the full output is fine. I hadn't appreciated when reviewing the test case that you'd set it to loose JSON matching and that the IDs are already present in the stubs.

Also, I've started the DSL work and it includes a couple of refactors so please could you pull from here:
https://github.com/tomakehurst/wiremock/tree/MasonM-add-snapshot-admin-task

Thanks again!

tomakehurst:
I'm having trouble pushing to your branch for some reason. Sure it's my setup, but can't figure out what's up. Would you mind pulling from mine for now?

tomakehurst:
My last couple of commits (on my branch still) modify the capture headers behaviour based on my earlier comment.

MasonM:
Okay, I merged and push your changes. Github is saying you haven't accepted the invitation yet, so maybe the e-mail got lost. I think you can visit this link to accept it: https://github.com/MasonM/wiremock/invitations

For the body filename truncation fix, there's one other issue I found: Unicode characters in the path part of a URL or IRI are being included un-encoded. Here's a shell script that reproduces this by recording a request to https://en.wiktionary.org/wiki/%C2%A5:
```bash
#!/bin/bash

java -jar build/libs/wiremock*standalone*.jar &
WIREMOCK_PID=$!
trap "kill $WIREMOCK_PID" exit

until $(curl --output /dev/null --silent --head http://localhost:8080); do
        sleep 1
done                                         

curl -s -d '{ "response": { "proxyBaseUrl": "https://en.wiktionary.org" } }' http://localhost:8080/__admin/mappings > /dev/null

curl -s 'http://localhost:8080/wiki/%C2%A5' > /dev/null

curl -s -d '{ "extractBodyCriteria": { "textSizeThreshold": "1" } }' http://localhost:8080/__admin/recordings/snapshot
```

The Yen symbol [is not allowed in NTFS](https://msdn.microsoft.com/en-us/library/windows/desktop/dd317748(v=vs.85).aspx), so I think this will break on Windows, but I haven't tried it.

We could just percent-encode all Unicode characters, but that'd make the filenames hard to read when recording requests to non-English sites. Providing a`"bodyFileNameTemplate"` option for customizing the filename seems like the best option. Using Handlebars for the template would probably be pretty easy, since we already have a request model for templating. We'd need to add a few properties and helpers, and the default template would look something like:
`{{#truncate 255}body-{{urlSlugify request.url}}-{{uniqueId}}.{{detectExtension request.body}}{{/truncate}}`

tomakehurst:
I think it'd be a good idea to have a default filename generator that can produce safe filenames across all platforms even in the presence of odd unicode chars. If that means producing ugly filenames for non-latin-based languages, then I'd say that's worth it (at least in the short term) if it means we can get it released sooner.

I like your idea for a name template too though. Maybe the template should be for the stub's name field in addition to the body file (so that they're always the same)?

MasonM:
Sorry I haven't been active in the last few days; I'm currently on a business trip. I'll be getting back Thursday and have the weekend free to work on this.

tomakehurst:
No worries, you've contributed a huge amount already!

I think the only significant thing left to do (apart from a mountain of documentation) is the stub and file naming strategy stuff we talked about.

It'd be great to get your feedback on what I've done also though. To summarise (in no particular order):

* Added start/stop recording. Essentially this just bookmarks the starting event when you hit record, then snapshots everything between that and the end event when you hit stop.
* Added a very basic UI web page under `/__admin/recorder` to start and stop recording (since there's no CLI equivalent this time).
* Added everything new in the API to the DSL.
* Tweaked the behaviour of the `persist` flag so that it still adds the stub mapping, but with its `persist` flag set to false (user can always hit reset afterwards if all they wanted was to download the stubs).
* Fixed a bug with the scenario recording (it wasn't setting the state progression up correctly)
* Changed the capture headers behaviour as per my comment earlier (it now uses the captured value).
* Added a JSON matching flags section to the spec
* Moved recording related classes to the `recording` package, as there are now quite a lot of them.

tomakehurst:
OK, I hope you don't mind but I went ahead and wrote a basic naming implementation that hopefully satisfies OS naming restrictions. It now does the following:

* Generates a name for the stub, based on the URL
* Names the file after the stub name + ID, rather than the generic "saved-stub-xxxx"
* Names the body file after the stub name + ID also, but with the correct file extension
* Truncates names to a max of 242 (name 200 + ID 37 + extension 5), since 255 seems to be the predominant OS name length limit

tomakehurst:
I'd still like to see your idea about templating the name implemented, but I think it's worth doing a release on what we've got in the meantime.

tomakehurst:
Right, I'm going to stop changing stuff now.

I'll hold off merging this until you've had a chance to take a look.

MasonM:
Okay, I'm back and I've had a chance to look things over a bit. I haven't reviewed everything, but your changes look great! Just a few comments:

1. Yes, now that I think about it, templating for the body filename is probably overcomplicating things for the initial release. Your code to do least-common-denominator naming in `SafeNames` looks good.
2. With body file extraction, there's still the issue with duplicate files being created for identical requests. This is a problem when recording multiple times (e.g. when running a test suite), and since the extracted body files accumulate until they're manually cleared, it can be pretty annoying. I can think of two solutions to this:
    1. Instead of using the StubMapping ID in the filename, use a hash of the file contents. I did a little research, and [FarmHash](https://github.com/google/farmhash) is a good choice for the hash function, since it's very fast, has few collisions, and is included with Guava. For clarity, we should probably continue using the URL/extension in the name, which means it's still possible for duplicate files to be generated if different requests result in the same response, but that seems like an acceptable tradeoff. [I implemented that in this branch](https://github.com/MasonM/wiremock/compare/add-snapshot-admin-task...MasonM:hash-mapping-body-filename) and tested it out pretty extensively. If it looks good to you, it should merge cleanly. If not, that's fine.
    2. Keep track of extracted response bodies and skip duplicates. Won't work for files that are already on disk and will increase memory usage, but won't be hard to implement.
3. The `JsonMatchingFlags` stuff looks good, but that only deals with the use case of customizing JSON body matching. The proposal in https://github.com/tomakehurst/wiremock/issues/707#issuecomment-312794279 is more general, since it allows overriding the `valuePatternForContentType()` logic for deriving the matcher from the `Content-Type` header. The primary use case for this is with non-standard MIME types for JSON and XML bodies. Many applications send JSON with the MIME type "text/javascript" or "text/plain" (see https://github.com/mislav/faraday-stack/pull/2 for examples), which will cause the current logic to use `equalTo` . Another possible issue is DTDs: [the standard MIME type for DTDs is application/xml-dtd](https://www.ietf.org/rfc/rfc3023), which will cause it to use `equalToXml`. I implemented the proposal [in this branch](https://github.com/MasonM/wiremock/compare/add-snapshot-admin-task...MasonM:body-matcher-work), except I inlined "matchOptions" to keep the deserialization code simple. It it quite a bit more complex than the current code, and I'm not sure how common the cases I mentioned above are, so I'll leave it you to decide if that added complexity is worth it. If so, that branch should merge cleanly.
4. The simplified start/stop recording endpoints are a good idea, but maybe it'd be worth allowing users to specify the whole `ResponseDefinition` instead of just the `targetBaseUrl`. The two use cases for this are passing additional headers via `additionalProxyRequestHeaders` and response templating so the proxy URL is dynamic (which I added in PR #641). I have a working implementation of this [in this branch](https://github.com/MasonM/wiremock/compare/add-snapshot-admin-task...MasonM:accept-proxy-response), but I'd need to update the docs and add more tests. I'm not quite sure if this is worth the added complexity.
5. The parameter list for the snapshot API has gotten pretty long, and may be intimidating for new users. The start/stop recording helps a lot with this, but maybe it'd be worth structuring the parameters a little more. One way of doing that would be wrap the options specific to generating StubMappings in a separate object, called something like `"generatorOptions"`. [Here's a branch implementing that](https://github.com/MasonM/wiremock/compare/add-snapshot-admin-task...MasonM:wrap-generator-options), and here's what a request would look like:

```json
POST /__admin/recordings/snapshot
{
    "filters" : {
        "urlPathPattern" : "/api/.*",
        "method" : "GET",
        "ids" : [ "40a93c4a-d378-4e07-8321-6158d5dbcb29" ]
    },
    "outputFormat" : "FULL",
    "generatorOptions": {
        "persist" : false,
        "captureHeaders" : {
            "Accept" : { },
            "Content-Type" : {
                "caseInsensitive" : true
            }
        },
        "extractBodyCriteria" : {
            "textSizeThreshold" : "2 kb",
            "binarySizeThreshold" : "1 Mb"
        },
        "repeatsAsScenarios" : false,
        "transformers" : [ "modify-response-header" ],
        "transformerParameters" : {
            "headerValue" : "123"
        },
        "jsonMatchingFlags" : {
            "ignoreArrayOrder" : false,
            "ignoreExtraElements" : true
        }
    },   
}
```

I'll try to finish reviewing tomorrow.

MasonM:
Okay, I'm finished reviewing and haven't noticed any major issues. Of the 4 things I brought up in my last comment, I think the issue with duplicate body files is the only major one.

I was thinking a little more about re-structuring the parameter list, and I think it'd be clearer to additionally wrap the request and response options in objects inside `generatorOptions`, e.g.
```json
POST /__admin/recordings/snapshot
{
    "filters" : {
        "urlPathPattern" : "/api/.*",
        "method" : "GET",
        "ids" : [ "40a93c4a-d378-4e07-8321-6158d5dbcb29" ]
    },
    "outputFormat" : "FULL",
    "generatorOptions": {
        "request": {
            "captureHeaders" : {
                "Accept" : { },
                "Content-Type" : {
                    "caseInsensitive" : true
                }
            },
            "jsonMatchingFlags" : {
                "ignoreArrayOrder" : false,
                "ignoreExtraElements" : true
            },
            "repeatsAsScenarios" : false
        },
        "response": {
            "extractBodyCriteria" : {
                "textSizeThreshold" : "2 kb",
                "binarySizeThreshold" : "1 Mb"
            }
        },
        "persist" : false,
        "transformers" : [ "modify-response-header" ],
        "transformerParameters" : {
            "headerValue" : "123"
        }
    }
}
```

The advantage to that is it's immediately obvious that `extractBodyCriteria` applies only to response bodies, `captureHeaders` only applies to requests, and `transformers` applies to the whole StubMapping. This does make calls to the API more verbose, so there is a tradeoff. I implemented this change [in the branch I mentioned previously](https://github.com/MasonM/wiremock/compare/add-snapshot-admin-task...MasonM:wrap-generator-options).

tomakehurst:
1. 👍
2. I'm tempted to live with this for now. There's a bit of a tradeoff - although capturing multiple copies of the same response body isn't very space efficient, it does keep things simple. If I want to edit a stub after recording I can change the body file safe in the knowledge the edit will only affect the intended stub.
3. I like this, let's merge it. I'd like to add matchesJsonPath and matchesXPath support in the near future, and this will enable that nicely.
4. I did consider whether it would be worth giving more control over the proxy mapping when start/stop recording, but TBH for now I think it's enough
just to advise people wanting full control to use snapshotting. My main intention for including start/stop recording was to help beginners who might 
find snapshotting harder to grasp than recording, so it's deliberately a bit simplified.
5. I can see your point, but OTOH an extra level of nesting is a bit painful for folks wanting to construct e.g. a curl request to record/snapshot.
On balance, I think this probably isn't worth the effort so I'd prefer to leave as-is.

tomakehurst:
Great thanks. I think it might be time to release!

MasonM:
>  If I want to edit a stub after recording I can change the body file safe in the knowledge the edit will only affect the intended stub.

Okay, that makes sense. The issue for me is that the mock data I'm working with largely consists of media files, so I have to delete old body files each time I re-record my test suite. That makes Docker-based deployment of Wiremock difficult, since it means I have to map the "__files" directory to the host. 

Would you be open to having `DELETE /mappings` accept a boolean parameter called "removeBodyFiles" that deletes all body files associated with the stub mappings? That'd solve this issue and be easy to add. I can enter a PR for that in the next day or two.

> I like this, let's merge it. 

Okay, done. There are a couple issues with `RequestBodyAutomaticPatternFactory` that I added hacks for:
1. Bodies of type "multipart/form-data" were using `equalTo`, which meant it never matched. To match a m "multipart/form-data" requests requires a new matcher, since it needs to use the boundary specified in the `Content-Type` header. I added a hack to have it use `AnythingPattern` in that case, since that seems better than a stub that never matches.
2. Binary bodies always used `equalTo`, which matched by the UTF-8 decoded body. Since that corrupts non-UTF8 bodies, there needs to be a way to match against the binary body. Adding a `equalToBase64` matcher would work, and I think I can get a PR in for that this weekend. In the meantime, I just added a hack to use `AnythingPattern`

tomakehurst:
OK, understood. I like the idea of being able to specify that deleting a mapping should delete the corresponding body file. If you're happy to PR that, that'd be great.

Form data support isn't great in WM at the moment. I've been meaning to think about adding a matcher for that for a while.

Agree binary body matching also needs sorting out. Again, if you're happy to work on that, it'd be much appreciated.

Overall I don't think these issues are worth holding up a release for. Would you agree?

MasonM:
> Overall I don't think these issues are worth holding up a release for. Would you agree?

Sure, if you're fine with the hacks in `RequestBodyAutomaticPatternFactory`. It's 2AM here, so I'm off to bed. I'll work on adding a `equalToBase64` matcher tomorrow.

tomakehurst:
I'll take a final glance at it. In any case I think I'm going to tweak the DSL a little before finalising.

Thanks again for all your help, sorry if I've kept you up late!

tomakehurst:
And we're live! One more time - thanks for all your help. I think this is the most awesome new feature WireMock has had for a long time, and I'm looking forward to seeing what feedback it gets.

PR #682, opened 2017-06-08
Add getOptions() to Admin interface

MasonM:
For the new recording API endpoint, we'll want to be able to extract response bodies to a file and load stub mapping transformer extensions. This means having access to `Options.filesSource()` so it can write out the extracted bodies, and `Options.extensionsOfType()` for loading the transformer extensions.

I'm not sure if `WireMockServer.getOptions()` and `WireMockApp.getOptions()` are really worth testing, since they're simple getters. I wrote a test for former since there's already a test class for it.

(I'll reply to your comments in PR #674 tomorrow. Just wanted to get this out here)

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -372,6 +372,11 @@ public class WireMockServer implements Container, Stubbing, Admin {
     }
 
     @Override
+    public Options getOptions() {
+        return options;
+    }
+
+    @Override
     public void shutdownServer() {
         shutdown();
     }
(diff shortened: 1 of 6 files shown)


tomakehurst:
I've got no issue with adding this. Can you elaborate why we'd need to load extensions programmatically for the new recorder?

MasonM:
For stub mapping transformer extensions, I'm envisioning having `SnapshotTask` grab those using `admin.getOptions().extensionsOfType(StubMappingTransformer.class)` and apply them to each generated stub mapping. Were you thinking of doing that at a lower level (e.g. `WiremockApp.addStubMapping()`)? The problem I see with that is the transformer wouldn't be able to distinguish recorded stub mappings from manually-created ones.

tomakehurst:
OK, that makes sense.

PR #690, opened 2017-06-20
Fix binary body corruption for logged responses

MasonM:
While working on the snapshot feature, I discovered that binary response bodies were getting corrupted. This is happening because the `LoggedResponse.from()` method sets the body by calling `Response.getBodyAsString()`, which encodes the body in the charset for the content type header (or UTF-8, if that's absent). Any invalid characters in the body get silently replaced with the default replacement character (U+FFFD for Unicode).

I fixed this by just copying what `LoggedRequest` is doing: instantiate the body using a Base64-encoded string and have a `getBodyAsString()` method that does necessary encoding.

I moved the `Response.encodingFromContentTypeHeaderOrUtf8()` method to `ContentTypeHeader.encodingOrUtf8()` so both `Response` and `LoggedResponse` could use it.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Strings.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Strings.java
@@ -15,23 +15,34 @@
  */
 package com.github.tomakehurst.wiremock.common;
 
+import java.nio.charset.Charset;
+
 import static com.google.common.base.Charsets.UTF_8;
 
 public class Strings {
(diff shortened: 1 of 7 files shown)


MasonM:
Build failure looks unrelated

MasonM:
I fixed a small typo ("base64" => "utf-8").

@tomakehurst  These changes will conflict with PR #703, but I don't mind resolving the conflicts if that's merged first.

tomakehurst:
Thanks, looks good!

PR #703, opened 2017-06-29
obeying the charset specified in the content-type header

uevilwizard:
Wiremock was not considering the charset of the submitted data.  The clearest example of this was in the existing unit test requestBodyEncodingRemainsUtf8().  The test posts a body containing utf-8 data, but specifies a content type of ContentType.TEXT_PLAIN, which attaches a HTTP header of "Content-Type" : "text/plain; charset=ISO-8859-1" to the post.   So in this test, the body's charset does not match what is advertised in content-type header, yet the test passed because the request-handling methods ignored the charset value, assuming the submitted content would be utf-8.  

In these changes, I've modified the Wiremock code to use the charset of the submitted data, falling back to utf-8 if the charset is not specified.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -351,7 +351,6 @@ task npmInstall(type: Exec) {
     commandLine 'npm', 'i'
 }
 
-/*
 task generateApiDocs(type: Exec) {
     dependsOn npmInstall
     workingDir 'docs-v2'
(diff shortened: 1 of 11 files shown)


mariohosmar:
When is this going to be merged?, new version release?

build.gradle
```diff
@@ -351,13 +351,15 @@ task npmInstall(type: Exec) {
     commandLine 'npm', 'i'
 }
 
+/*
```
MasonM: I think you forgot to remove this comment

uevilwizard: Oops... thanks

uevilwizard:
@tomakehurst  I've added a test case that uses GB2312.

fenoloco:
We are waiting for this PR. Would be nice if it can be merged 👍 . Just let us know if we can help

PR #715, opened 2017-07-14
Update docs to extending docs to new getName method

skeletonkey:
Seems that that the documentation is behind.  My IDE had me override 'getName' instead of 'name'.  Hope this is correct.


tomakehurst:
Good catch, thanks!

Issue #717, opened 2017-07-14
Allow access to the host/port in response template


samuel-soubeyran:
Templating is an awesome feature but I have a use case that is currently not supported and probably wouldn't require too much work to support.

I am writing test against webservice. I first retrieve the WSDL and then make the SOAP request.
I want wiremock to serve these 2 requests and to avoid port conflict I want to be able to use dynamic port. 

Now the issue is that the wsdl reference the full url with port where the 2 request should be made. So I need to be able to figure out the right port on which wiremock is listening when serving the first request.

to make it more clear, here is an small part of my WSDL where I'd like to template the port:

```
...
<wsdl:port name="WebServiceSoap" binding="tns:WebServiceSoap">
    <soap:address location="http://localhost:{{wiremock_port}}/web/service"/>
</wsdl:port> 
...
```
My understanding is that this could be implemented in RequestTemplateModel, using request.getAbsoluteUrl(). But there might be better ways or more things that other would want to include.
If you can give me some general idea on how you'd envision this, I'd be willing to send a PR.

samuel-soubeyran:
on a side note, I feel that url for the path is slightly misleading, AFAIK url should be the full "absolute" url include scheme, host, port, path and query (https://www.ietf.org/rfc/rfc1738.txt). Currently it seems that in the Request interface, getUrl() return the path, and getAbsoluteUrl() return the url. It's not a big deal but I'm curious about the reasons behind this choice (the logic is in WireMockHttpServletRequestAdapter).

leeturner:
This is a pretty old enhancement request and WireMock has supported `scheme`, `host` and `port` in the request model for a little while now.  As a result I am closing this issue

PR #718, opened 2017-07-14
Add absolute url in the RequestTemplateModel tomakehurst/wiremock#717

samuel-soubeyran:
Looking for comment about this first attempt.
In particular I look to have feedback on the change in the signature of the public constructor. It doesn't seems to cause issues in the project and I don't see why some other codebase would want to reuse this class but arguably I don't have the full picture.

A solution might be to have 2 constructors. Alternatively if nobody is meant to directly instantiate this class (but use the static method `RequestTemplateModel from(final Request request)`) then what about making the constructor private (or protected)?

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
@@ -13,6 +13,7 @@ import java.util.Map;
 
 public class RequestTemplateModel {
 
+    private final String absoluteUrl;
     private final String url;
     private final UrlPath path;
     private final Map<String, ListOrSingle<String>> query;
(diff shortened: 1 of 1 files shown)


tomakehurst:
I think this would be a lot more more useful if the URL was wrapped in a model that exposes the scheme, host, port, path etc. as individual fields rather than just the whole string.

Also some test cases.

samuel-soubeyran:
Got it, make sense.
Concerning the test, I didn't get a clear feeling on where I should put them.
If you don't have a better suggestion, I'll put them in ResponseTemplateTransformerTest.

tomakehurst:
Yeah, that's where I'd have suggested putting the tests.

Thanks.

samuel-soubeyran:
Let me know what you think about this. I am not super happy with the naming. Also I considered actually not wrapping it in a nested object but directly exposing the information directly through RequestTemplateModel.

Issue #704, opened 2017-06-29
Templating XML or XPath from request.body


tatol:
Hello! Your WireMock is very good! I use it standalone.
Now if i have XML or XPath in request body, i can not get any parts of request.body by handlebars.
Is it possible to do this now?
If not, will this be improved?
And can i use some plugins to do it now?
Thank you!

tomakehurst:
You should be able to get the whole request body, but you can't at present extract parts of it via xpath. Is that what you had in mind?

tatol:
I mean, if i have xml in request body like ...<thing>Hello</thing>..., i can not get "Hello" string by {{request.body}} in response. Will this be developed in the future?

tomakehurst:
I'm not exactly sure what you mean, but I am intending to develop handlebars helpers for extracting parts of request XML documents via XPath if that helps.

tatol:
It would help a lot

holomekc:
We had similar needs so I extended wiremock with helpers for xml via xpath and another one especially for soap and one for json via jsonPath. I am currently working in a fork of wiremock but the helpers are not there yet.

The reason is that I created a gui for wiremock which I updated to angular in this new fork. I want to add the helpers in a new branch an then create a pull request. I hope this helps.

PR #730, opened 2017-08-07
added wiremock exclusive handlebars helpers:

holomekc:
Hi,
as I promised in  #704 . Here the three helpers for xml, soap and json content. Xml and soap are plain old Java but Json via JsonPath uses a lib described below. I also added some tests. Here some other points:

1.  Feel free to rename the classes. I was not sure about the naming
2.  I was not sure about the handling of errors if xml/json or xPath/jsonPath are invalid. Currently it fails gracefully and shows an error message in the response. The intention was that it is easier to find the issue.

Best regards,
Christopher

-added com.jayway.jsonpath:json-path:2.2.0 to dependencies which allows the JsonPath implementation

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -66,6 +66,7 @@ dependencies {
     compile 'com.github.jknack:handlebars:4.0.6', {
         exclude group: 'org.mozilla', module: 'rhino'
     }
+    compile 'com.jayway.jsonpath:json-path:2.2.0'
 
     testCompile "org.hamcrest:hamcrest-all:1.3"
     testCompile("org.jmock:jmock:2.5.1") {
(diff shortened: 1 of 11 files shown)


tomakehurst:
This is awesome stuff.

Would you mind adding a few acceptance tests, illustrating the helpers in use in actual response templates?

Thanks!

holomekc:
Done. You can check if this is enough. I needed to extend the json helper, because it was not handling all jsonPath exceptions. 

Still I am not sure about the exception handling. JsonPath behaves differently compared to the xPath implementation. E.g. when path not available in xPath simply no content is returned. JsonPath on the other hand throws the exception which leads to the Error text I wrote. 

What do you think?

tomakehurst:
I made some error handling and name tweaks, and I also added support for selecting fragments of JSON and XML in addition to simple values.

Thanks for contributing this!

PR #738, opened 2017-08-15
Gui for wiremock

holomekc:
Hi guys,

I like wiremock very much and for automatic testing it is really great as it is. I also use it for local deployment. Therefore, all external services are mocked. What I always wanted is a graphical user interface so that I can easily see what is happening or that I can change a mapping pretty easily.

Some time ago I wrote a GUI with Angular js which helps me a lot. Then I thought ok I can rewrite this with the new Angular version and then I though maybe I should share it. So here we go:

Changes in wiremock itself:
--

- Server Side Events added so that the gui can be updated automatically. (Please check this. I tried to implement it in a way so that I did not change too much)
- add new admin Servlet /webapp/* (Redirect rules needed so that every request against webapp/* is forwarded to corresponding index.html. Anchor would be possible as well but I like it with path better.)

The GUI:
--
The GUI consists of three pages:
1. Mappings
2. Matched
3. Unmatched

**Layout:**
All pages have the same layout

**Left side:**
On the left side is a list of items. E.g. for the mappings view it would be a list of mappings. The url, urlPath, urlPattern,... is the highlighted text. Below are some additional information like HTTP method, status or the SOAP method. Above of the list is the pagination. 

**Center:**
The other part of the layout is the center. Which also consists of multiple parts:
1. A search. Which allows to filter the items via regex (every value of the mapping or loggedrequest is searched) or a full text search (used when regex not valid). The checkbox at the end enables case sensitivity
2. Some actions which can be executed on the selected item.
3. The actual content which is also devided.

**The actual content:**
The first tab shows the mapping or loggedrequest as it is. So it is a json string. The string is highlighted with highlight.js and formatted with vkbeautify

The second tab provides a sparated view. (Sorry for the namings. I had no better idea). In this tab the mapping or loggedrequest is devided into their parts. General, Request, Response, ResponseDefinition. This tab comes in handy when the body content is a json or xml string. It also shows the query parameters in a readable way.

It is hard to explain in words so let us start with the views:

Mappings:
--
On the left you can see two mappings. The second one (not selected) has the url /some/thing, method GET and HTTP status 200 for the response. The first one (selected) is a mapping which maybe makes not much sense but shows some features of the GUI. Like the SOAP method GetPrice. Which is extracted from the matchesXPath defined in the mapping itself. This helps to distinguish SOAP requests. The pagination is setup to max 40 items on a page. We currently have only one page.

**Raw:**
mappingraw

**Separated:**
mappingseparated

**Mappings actions:**
_New:_ 
create a new mapping. It shows a basic mapping structure which helps to create a new mapping.
newmapping

- Save&Cancel: I think I do not need to describe this
- Helpers: Add some useful configuration to the mapping like a delay or other stuff. 
helpers

_Edit:_
like new but related to the selected mapping.

_Remove:_
remove the selected mapping

_Mappings:_
- Save Mappings -> /__admin/mappings/save
- Reset Mappings -> /__admin/mappings/reset
- Remove all Mappings -> /__admin/mappings method=DELETE.

_Scenarios:_
-Reset all Scenarios -> /__admin/scenarios/reset

Matched:
--
The layout is the same. Here the few differences:

- Mapping link:
In separated -> General -> stubMapping uuid is a link the the mapping which was used to answer the request. This is done to quickly edit the relevant mapping. Also the stubmapping is not again listed in the separated view although it is part of the json string. This part is still visible in the raw view.

matched

**Matched actions:**
- Reset Journal -> /__admin/requests method=DELETE

Unmatched:
--
Again similar. Let us go directly to the actions.
Btw. I am not sure yet why favicon.ico is requested. I specified to use the wm png file which is also correctly shown.

unmatched

**Unmatched actions:**
- Copy Mapping: Copies a string the the clipboard which contains a basic mapping which fits to the unmatched request.
- Copy Soap Mapping: Only available when a soap request is recognized. If so this button would copy a basic mapping for the soap request. It will look like the one shown in the picture regarding mappings. Although the body is empty.
- Reset Journal: Same as for matched.

Settings:
--
Maybe you already noticed. At the top right corner the GUI has some settings.

settings

- Dark Theme: enables or disables the dark theme. At the end I will post one image of the light theme.
- TabSlide: I use Material Design 2 by google. The tabs have a fancy slide animation. It is kind a cool but also ... well. You know.
- Hide empty values: When you saw the Separated tab. E.g. of the mappings view then only fields with content were shown. With the setting you can toggle between hide empty or show all.
- Reset all -> /__admin/reset

- Start Recording: The new recording. Legacy is not supported by gui. A popup is shown where the url can be posted. If recording is enabled it is shown in the toolbar.

**Enter Url:**
recording

**Recording status:**
recordingindicator

- Stop Recording:
Stops the recording. All recorded mappings will be shown. (There is still a "bug". You need to be on the mappings view... I will add a redirect soon.) It basically creates a regex for all created uuids of the mappings.

recordingstopped

- Snapshot:
/__admin/recordings/snapshot. Same behavior as described in Stop Recording.

- Shutdown:
Shutdown wiremock

**Light theme:**
As promised
lighttheme

Work in progress:
--
Some stuff I had in the old version and I want to add.
- decode basic authorization if available
- proxying switch. Enable or disable proxying on a mapping. I had this in my old version. Currently I am searching for a way to implement it without destorying everything. In the old version I manipulted the mapping. Removed or readded the proxyBaseUrl which I stored in a separate map. 
-reset only selected scenario
- tests for the frontend
- update readme
- cleanup

I need help:
--
- Proxying:
My idea was to add a flag which activates the proxying on a mapping. Why? Well as far I can see I could write a normal mapping and add a proxyBaseUrl and via the flag I could specify If I want to use the proxy or the mapping itself. One use case would be using the external service and when it is down for what ever reason I could disable proxying and continue my work. 
I tried to do so. But I basically broke every junit test. So I reverted this changes. Maybe someone with more experience could help me here.
- Regarding gradle:
I did not commit all my changes here because I would destroy the build. I am also not that familiar with gradle. It looks like you are developing on a linux based system. I had some issues regarding the apiDocs. So I removed this step for me locally. (I need to setup a vm though). I will explain what I configued. Maybe someone with more experience here can explain me how to integrate it in a better way.

_What I did to integrate the fontend is:_
1. remove npmInstall
2. add the following:

```
plugins {
    ...
    id "com.moowork.node" version "1.2.0"
}

...
node {
    version = '7.10.1'
    yarnVersion = '0.27.5'
    download = true
    workDir = file("${project.buildDir}/node/")
    nodeModulesDir = file("${project.projectDir}/src/main/frontend/wiremock/")
}
...
task buildFrontend(type: YarnTask){
    args = ['run', 'buildOld']
}

task devFrontend(type: YarnTask){
    args = ['run', 'start']
}

shadowJar.dependsOn buildFrontend
```

Test it:
--
Normally you could simply build shadowJar and it should work. If you add the stuff I posted above. If it does not work do the following:

_Install:_
- node.js
- npm and then yarn. Well not really needed. But I like yarn more.
- angular-cli

The frontend is located in src/main/frontend/wiremock/.

You can start wiremock as it is on 8080 (or change config in environments/environment.ts) and then execute the following in console:

```
cd <path>/src/main/frontend/wiremock/
yarn install
yarn run start
```

open browser localhost:4200/__admin/webapp/

If you have some issues let me know.

Feedback:
--
The GUI is adjusted to my needs. I tried to make it not too specific. Please tell me what you think and what I miss or what could be added. I know it is a lot of text. I hope the pictures made it more interesting.

Br,
Christopher


tomakehurst:
I think this looks great.

However, it's very large and I think it would be better served existing as a separate project. A major advantage of this is that you'd be free to work on it and release it on your own schedule. I'm already struggling to keep up with WireMock's maintenance in its current shape and a major feature like this will inevitably produce a slew of issues and PRs I'd struggle to prioritise.

I realise that there currently isn't an extension point that can be easily exploited to serve static assets, so I'd like to propose making a change to WireMock to support this. This would also enable me to extract the Swagger UI stuff I put in there a while ago to a separate project (in retrospect I think it was a mistake to put this into the core).

What do you think?

holomekc:
Hi,

sorry for my very late response. I was quite bussy. Yeah I think you are right. Ok so I will keep it in the fork and try to get the on/off switch for proxying running. If there would be a feature to toggle the proxying I would not mind ;). Nevertheless, the merging should be much easier as with my old version of the gui. 

I will close this pull request. Thx a lot

PR #746, opened 2017-09-04
Restored HttpAdminClient with host head and no proxy settings

ggranberry:
Wasn't sure if you wanted tests for the constructors over in HttpAdminClientTest, so I left them out for now.


tomakehurst:
Normally I'd insist on a test, but this is a trivial enough case not to bother IMO.

However, I would suggest only having one constructor that actually does the work and the rest referencing this(....) rather than two with mostly duplicated code. Please could you make that change?

ggranberry:
Amended my commit with the change you requested. Thanks!

tomakehurst:
Great, thanks!

PR #756, opened 2017-09-16
Fixed #567 - Allow duplicate cookies 

### LINKED ISSUE #567: Internal server error when sending duplicate cookie
Hi.

I'm experiencing a strange error in wiremock.

Let's say I have the following mapping:
```json
{
    "request": {
        "method": "GET",
        "url": "/test"
    },
    "response": {
        "status": 200,
        "body": "More content\n"
    }
}
```

If I make a request with two values for one cookie, like this:
```bash
$ curl 'http://localhost:8080/test' -H 'Cookie: cookie1=190283aJDS;cookie1=190283aJDS'
```

I get the following response:
```html
<html>

<head>
    <meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1" />
    <title>Error 500 </title>
</head>

<body>
    <h2>HTTP ERROR: 500</h2>
    <p>Problem accessing /test. Reason:
        <pre>    java.lang.IllegalArgumentException: Multiple entries with same key: cookie1=190283aJDS and cookie1=190283aJDS</pre></p>
    <hr /><i><small>Powered by Jetty://</small></i>
</body>

</html>
```

But if I remove the duplication, it works. For example:
```bash
$ curl 'http://localhost:8080/test' -H 'Cookie: cookie1=190283aJDS'
```

I know that we shouldn't set the cookie value twice, but in my real life case I really can't control that.
What could I do?

Also, I've tried to debug wiremock but I didn't go too far. Is there any debug flag (I'm already using --verbose) that I can see the stacktrace?

mushanga:
Fixed #567

**Issue**
https://github.com/tomakehurst/wiremock/issues/567#issue-199911597
> If I make a request with two values for one cookie, like this:
>$ curl 'http://localhost:8080/test' -H 'Cookie: cookie1=190283aJDS;cookie1=190283aJDS'
> I get the following response:
> 
> <html>
> 
> <head>
>     <meta http-equiv="Content-Type" content="text/html;charset=ISO-8859-1" />
>     <title>Error 500 </title>
> </head>
> </html>
> 
> 
> But if I remove the duplication, it works.

**Reply**
https://github.com/tomakehurst/wiremock/issues/567#issuecomment-272279919
>Cookies are represented by a map internally and the map builder blows up if you try to add more than one with the same key.

**Fix**
- Changes in [Cookie:](https://github.com/mushanga/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/http/Cookie.java) 
  - `String name` property added
  - Json `value` representation replaced with `object`
- Changes in interface [Request:](https://github.com/mushanga/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/http/Request.java) `Map<String, Cookie> getCookies()` method replaced with `List<Cookie> getCookies()`

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
@@ -22,7 +22,6 @@ import com.github.tomakehurst.wiremock.http.QueryParameter;
 import com.github.tomakehurst.wiremock.http.Request;
 import com.google.common.base.Function;
 import com.google.common.collect.Collections2;
-import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.Maps;
 import com.google.common.collect.Maps.EntryTransformer;
 import com.google.common.collect.Multimaps;
@@ -64,7 +63,7 @@ public class RequestTemplateModel {
             }
         });
 
-        ImmutableMap<String, Collection<Cookie>> indexedCookies = Multimaps.index(request.getCookies(), new Function<Cookie, String>() {
+        Map<String, Collection<Cookie>> indexedCookies = Multimaps.index(request.getCookies(), new Function<Cookie, String>() {
             @Override
             public String apply(Cookie input) {
                 return input.getName();
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
@@ -15,23 +15,18 @@
  */
 package com.github.tomakehurst.wiremock.extension.responsetemplating;
 
-import static com.google.common.collect.Lists.newArrayList;
-
 import com.github.tomakehurst.wiremock.common.Urls;
 import com.github.tomakehurst.wiremock.http.Cookie;
 import com.github.tomakehurst.wiremock.http.MultiValue;
 import com.github.tomakehurst.wiremock.http.QueryParameter;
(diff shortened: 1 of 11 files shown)


mushanga:
Is there a problem with `openjdk7` tests? Seems that it's been failing since [7th of September](https://travis-ci.org/tomakehurst/wiremock/builds/273065813)

Please let me know if I should fix something in my pull request.

tomakehurst:
Yeah, sorry about this. It's not your PR - I'm having problems with the Travis CI build image at the moment. Some combination of openjdk7, Gradle 4 and Travis' Linux image. I'm going to try and migrate to Circle CI today or tomorrow.

mushanga:
Great, thanks! :)

tomakehurst:
OK, the build is fixed on master now. If you merge across, it should go green.

Also, please could you make a few amendments to this PR:

The switch from `Map<String, Cookie>` to `List<Cookie>` on the `Request` interface breaks backwards compatibility (since this is an API class), so please could you restore this. I suggest instead that you modify the `Cookie` class so that it can take one or many values. `Cookie` is also an API class, so I suggest that you maintain its current interface but add extra methods to cope with the many values case.

This brings me to my 2nd point, which is that the way multi-valued headers are handled is to present the value as an array rather than a string when JSON serialised. I think it would make a lot of sense for cookies to be consistent with this, and a `Cookie` class that can handle many values would make this a lot easier.

Lastly, please could you add an acceptance test case for response templating with multiple cookie values.

Thanks!

tomakehurst:
I've fixed this in 827e7ccb3edcb15bb78e68add74adee536e921b2

tomakehurst:
I've fixed this in 827e7ccb3edcb15bb78e68add74adee536e921b2 now

PR #769, opened 2017-10-04
Rename setter to reflect intent, it is ambiguous otherwise

wojciechbulaty:


tomakehurst:
Hmmmm...I'm not really keen on breaking backwards compatibility on a public bit of the API. I agree it's not the best named method, but do you think it's really causing any significant confusion?

wojciechbulaty:
Sure. How about I create a new one and deprecate this one but leave it there?

wojciechbulaty:
The filesSource method name suggest to me you will be setting the location of __files directory, that is why I found it confusing.

I have added the old method back, but deprecated it, let me know what you think.

PR #771, opened 2017-10-05
added an endpoint to edit body files

wojciechbulaty:
Let me know what you think please.

I have not added any tests yet cos did not want to waste time if there is a better way of doing this.


tomakehurst:
I think this would be a great thing to have.

A few things to think about:

* What if you're trying to overwrite an existing file?
* Might be handy to have list and delete calls too
* The file write operation should be via the `FileSource` rather than doing it directly in the route handler IMO.

tomakehurst:
One further thought: how easy do you think it'd be to support multipart posts so that more than one file can be added at a time?

wojciechbulaty:
> What if you're trying to overwrite an existing file?

Good point, will test it.

> Might be handy to have list and delete calls too

I might add it soon. Would it be OK to add it as a separate pull request?

> The file write operation should be via the FileSource rather than doing it directly in the route handler IMO.

Will do that.

> One further thought: how easy do you think it'd be to support multipart posts so that more than one file can be added at a time?

Yeah, that seems like a useful thing. I was hoping to keep this changeset minimal if possible. May be somebody else can pick it up?

tomakehurst:
How about a compromise - add the list and delete endpoints in this change and defer multipart for a future PR?

wojciechbulaty:
Hi @tomakehurst , I believe I have implemented what we have discussed. Let me know your thoughts.

PR #772, opened 2017-10-11
Remove deprecated guava API usage as it breaks with Guava version 23.0 and newer

msvab:


tomakehurst:
Eugh, just tried upgrading to 23.1-jre and they seem to have introduced a ton of breaking changes.

tomakehurst:
And they've dropped JDK 7 support in any case...

PR #783, opened 2017-10-22
Fix JsonFileMappingsSource.remove() for subdirectories

MasonM:
The `JsonFileMappingsSource.loadingMappingsInfo()` method looks for JSON files in the mappings directory recursively, which is useful since it allows organizing stubs in a directory hierarchy. However, the `remove()` and `removeAll()` methods will not remove stubs in a subdirectory, because the path components are stripped out of the filename.

This was introduced in 7f93b5eb996722c1919a6e1e627d550f7929ac39, which changed the `getFileName()` to strip out all path components. I updated that method to only do so for non-file URLs.


src/main/java/com/github/tomakehurst/wiremock/standalone/JsonFileMappingsSource.java
```diff
@@ -87,6 +88,14 @@ public void loadMappingsInto(StubMappings stubMappings) {
 	}
 
 	private String getFileName(TextFile mappingFile) {
-		return mappingFile.getUri().toString().replaceAll("^.*/", "");
+		URI mappingFileUri = mappingFile.getUri();
```
MasonM: I considered moving this logic to a method in `FileSource` (called something like `relativePath()`), but `FileSource` is a public interface so I figured we shouldn't break BC for this.

tomakehurst: Adding new methods is fine. It's just renaming, removing or altering the signature of existing ones that causes BC problems. If you think this would be a preferable way to do it, I'd say go ahead.

MasonM:
I think I figured out a better fix: just change `AbstractFileSource#writableFileFor()` to handle file paths. In addition to fixing the `deleteFile()` method, it will also allow writing new stub mappings to subdirectories when calling `writeBinaryFile()` or `writeTextFile()`. 

Since `ClassPathFileSource` does not extend `AbstractFileSource`, it won't use this new logic.

Issue #592, opened 2017-01-26
ResponseDefinitionBuilder.withTransformerParameter changes original transformerParameter variable


viniciusffj:
Hey.

I was writing a wiremock extension that needs to change the transformer parameter and I found something that may be a bug.

While using `ResponseDefinitionBuilder.withTransformerParameter` to build a response definition upon another one, the original value for transformer parameter also changes, so we loose the original parameter value.

One can check this behavior by checking this test I've created:
```Java
    @Test
    public void withTransformerParameterShouldNotChangeOriginalTransformerParametersValue() throws Exception {
        ResponseDefinition originalResponseDefinition = ResponseDefinitionBuilder
                .responseDefinition()
                .withTransformerParameter("name", "value")
                .build();

        ResponseDefinition transformedResponseDefinition = ResponseDefinitionBuilder
                .like(originalResponseDefinition)
                .but()
                    .withTransformerParameter("name", "value2")
                .build();

        assertThat(originalResponseDefinition.getTransformerParameters().getString("name"), is("value"));
        assertThat(transformedResponseDefinition.getTransformerParameters().getString("name"), is("value2"));
    }
```

Is this expected or it is a bug?

tomakehurst:
I'm not sure what other behaviour you'd expect in this case (unless I'm totally misunderstanding you). You're using `name` as the key in both cases so surely you'd expect to to see `value2` in the derived object?

viniciusffj:
So, on the case above, does the `originalResponseDefinition` should have that value changed as well? That's the behavior that I'm sure it is correct.

So, I was writing an extension and I had a problem that the transformer parameters was changing over the requests, and I needed to use the original response definition.

Please, take a look at the [extension code](https://github.com/viniciusffj/wiremock-headers-parameters-transformer/blob/master/src/main/java/com/github/viniciusffj/wiremock/ParametersTransformer.java#L23) to see my solution.

If I remove it, [this test](https://github.com/viniciusffj/wiremock-headers-parameters-transformer/blob/master/src/test/java/com/github/viniciusffj/wiremock/integration/ParametersTransformerIntegrationTest.java#L54) will fail.

tomakehurst:
OK, I see the problem. This bit of code is a bit of a bug magnet. Time to sort it out properly I think.

PR #784, opened 2017-10-22
Issue #592- Fix ResponseDefinitionBuilder.like() to copy transformerParameters

MasonM:
This fixes issue #592. When `ResponseDefinitionBuilder.like()` was called with a ResponseDefinition that had transformer parameters, it just copied the reference. This means subsequent calls to `withTransformerParameter()` would modify the original object.


tomakehurst:
Good spot, thanks!

Issue #776, opened 2017-10-15
WireMock evaluation of all criteria generates unnecessary warnings


SebastianStavar:
**[Description]** 
  _Short version:_ Non-json requests are evaluated against json mappings with body patterns and fill the log with failed json validation warnings.
  _Longer version:_ Because WireMock does an aggregation of the results for all evaluation rules (RequestPattern.match(Request request)) instead of hierarchically allowing a subsequent criterion to be evaluated against only after the higher one was matched, mappings that would immediately be excluded based on things like URL are still evaluated for body pattern matches. This is an issue when json body pattern matching mappings exists because it tries to evaluate ALL requests [that did not previously match] against the json body pattern matching mappings, which fills the log with all kind of json validation warnings.

**[Repro]** 

1. Start WireMock.
2. Train WireMock [use something like Postman and POST to WireMockServer:WireMockPort/__admin/mappings/] with RAW body:

`        {
            "id": "fab6d56b-d66e-4a5a-810a-ab323e485103",
            "request": {
                "urlPath": "/unavailablepath",
                "method": "POST",
                "headers": {
                    "X-SessionId": {
                        "equalTo": "90930D577144AB198B2082CA56E6A0F5",
                        "caseInsensitive": true
                    }
                },
                "bodyPatterns": [
                    {
                        "matchesJsonPath": "$.params[0][?(@.parameterone == 'something')]"
                    }
                ]
            },
            "response": {
                "status": 200,
                "bodyFileName": "a.json",
                "headers": {
                    "Content-Type": "application/json"
                }
            },
            "uuid": "fab6d56b-d66e-4a5a-810a-ab323e485103"
        }`

4. Make sure the _a.json_ file exists. I do not believe its content is relevant, but when I was creating the repro for this bug this is what I put in mine:

`{
  "jsonrpc": "2.0",
  "id": "-1412417872994864924",
  "result": {
    "payload": {
      "timeStamp": "2017-02-17T16:26:28.588"
    }
  }
}`

5. Send a non-admin request to WireMock. I used an empty GET to WireMockServer:WireMockPort/services/none
6. Observe the WireMock console. Result contains something like (actual error differs based on body patterns in mappings and on the requests):

> 20:41:12 INFO  WireMock - Warning: JSON path expression '$.params[0][?(@.parameterone == 'something')]' failed to match document '' because of error 'json string can not be null or empty'
[...]

Issue #652, opened 2017-04-18
Body pattern matching possibly being attempted despite URL not matching


tomakehurst:
See https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/wiremock-user/DOmiAQ83W8E/rlncqKAXBQAJ

philippe-granet:
I fixed this problem with writting stubs using my custom `RequestMatcherExtension ` in my stubs like this:
```json
{
	"request": {
		"customMatcher": {
			"name": "optimized-request-matcher",
			"parameters": {
				"urlPath": "/my-url",
				"method": "POST",
				"bodyPatterns": [
					{
						"matchesXPath": {
							"expression": "//x/text()",
							"contains": "test"
						}
					}
				]
			}
		}
	},
	"response": {
		"status": 200,
		"body": "default test mapping",
		"headers": {
			"Content-Type": "text/plain"
		}
	}
}
```
with this implementation:
```java
public class OptimizedRequestMatcher extends RequestMatcherExtension {

	@Override
	public String getName() {
		return "optimized-request-matcher";
	}

	@Override
	public MatchResult match(Request request, Parameters parameters) {
		RequestPattern originalRequestPattern = Json.getObjectMapper().convertValue(parameters, RequestPattern.class);

		if (!originalRequestPattern.getMethod().match(request.getMethod()).isExactMatch()) {
			return MatchResult.noMatch();
		}
		if (!originalRequestPattern.getUrlMatcher().match(request.getUrl()).isExactMatch()) {
			return MatchResult.noMatch();
		}
		RequestPattern requestPattern = new RequestPattern(
				WireMock.anyUrl(), 
				RequestMethod.ANY,
				originalRequestPattern.getHeaders(), 
				originalRequestPattern.getQueryParameters(),
				originalRequestPattern.getCookies(), 
				originalRequestPattern.getBasicAuthCredentials(),
				originalRequestPattern.getBodyPatterns(), 
				originalRequestPattern.getCustomMatcher(),
				originalRequestPattern.getMultipartPatterns());

		return requestPattern.match(request);
	}
}
```

philippe-granet:
See also my PR https://github.com/tomakehurst/wiremock/pull/1049

TrickyDoodle:
@philippe-granet 
I've tried your implementation, but for some reason received the following error:
```


                                               Request was not matched
                                               =======================

-----------------------------------------------------------------------------------------------------------------------
[...]

PR #791, opened 2017-10-26
Attempt to fix issues #776 and #652 by introducing a flag to disable near miss calculation and delay evaluation of request matching. 

nzCharlie:
This pull request have the following changes: 

- introduce a new flag into wiremock `Options` class `boolean reportNearMissDisabled()`. If the flag is on, we will not try to calculate near misses.
- delays evaluation of default matching in `RequestPattern`, additional matches are not evaluated after previous match fails (if the flag above is set, otherwise, matches are still evaluated when near miss calculation is performed)


nzCharlie:
This pr should resolve issue #652 and issue #776

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -85,6 +85,7 @@
 
     private Authenticator adminAuthenticator = new NoAuthenticator();
     private boolean requireHttpsForAdminApi = false;
+    private boolean reportNearMissDisabled = false;
```
AlixBa: I'm wondering if disabled by default isn't better

AlixBa: (I would also name this reportNearMiss as it's simpler to understand when you're setting its value)

tomakehurst:
Let's back up here a little bit.

Disabling near miss calculation is the wrong thing to do. It doesn't (or shouldn't) happen during matching at all, only when you're asking for near misses so the problems this PR is meant to address won't be helped. If near miss calculation is taking place when it shouldn't, then this is the bug we should be fixing.

The excess of log messages can be fixed in the short term by removing them. The near miss feature provides a better way to figure out why your stub didn't match in any case.

A better way to do this might be to introduce a debug level on the notifier and switch the matcher messages to use this. That way you can keep info level logging enabled without flooding the logs with matcher messages.

nzCharlie:
OK, in this case I will try to set the debug level of the warning message from matcher in question, and see if it will work for us.

PR #793, opened 2017-10-26
Add ability to require absent query parameter.

yinonavraham:
It is sometimes needed to be able to verify that a request was sent without a query parameter. 
This pull request add this ability, similar to the existing support for requiring absent headers.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
@@ -37,6 +37,7 @@ public interface Request {
     Map<String, Cookie> getCookies();
 
     QueryParameter queryParameter(String key);
+    Set<String> getAllQueryParameterKeys();
 
     byte[] getBody();
     String getBodyAsString();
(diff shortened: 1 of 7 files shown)


tomakehurst:
Why doesn't something like this work for you?

```java
@Test
public void matchingOnAbsentQueryParam() {
  stubFor(get(urlPathEqualTo("/absent-query"))
        .withQueryParam("not_this", absent())
         .willReturn(ok()));

  assertThat(testClient.get("/absent-query?not_this=hello").statusCode(), is(404));
  assertThat(testClient.get("/absent-query?this_is_fine=hello").statusCode(), is(200));
}
```

yinonavraham:
Hi @tomakehurst ,
Thanks for your reply. I didn't think about it this way...
Your example made me understand that what I was really looking for can already be done as verification:
```java
@Test(expected = VerificationException.class)
public void throwsVerificationExceptionOnUrlPathEqualsOnExpectedAbsentQueryParam() {
  stubFor(get(urlPathEqualTo("/absent-query")).willReturn(ok()));

  testClient.get("/absent-query?not_this=hello");
  verify(getRequestedFor(urlPathEqualTo("/absent-query")).withQueryParam("not_this", absent()));
}
```
Do you think it is worth adding this test case to `VerificationAcceptanceTest`?

tomakehurst:
Seems like it might be worth adding it as an example in the docs. Fancy raising a PR for that?

tomakehurst:
Closing this, as I don't think there's anything else to be done on it for now.

PR #796, opened 2017-11-07
Switched base64 encoding to native Java library

marsbradley:
Better performing according to this: http://java-performance.info/base64-encoding-and-decoding-performance/

Been profiling stubs with base64 encoded JSON objects and base64->encoding/decoding is by far the hottest method.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Encoding.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Encoding.java
@@ -15,19 +15,19 @@
  */
 package com.github.tomakehurst.wiremock.common;
 
-import com.google.common.io.BaseEncoding;
+import java.util.Base64; 
 
 public class Encoding {
 
     public static byte[] decodeBase64(String base64) {
         return base64 != null ?
-            BaseEncoding.base64().decode(base64) :
+            Base64.getDecoder().decode(base64) :
             null;
     }
 
     public static String encodeBase64(byte[] content) {
         return content != null ?
-            BaseEncoding.base64().encode(content) :
+            Base64.getEncoder().encodeToString(content) :
             null;
     }
 }
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Encoding.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Encoding.java
@@ -15,19 +15,19 @@
  */
 package com.github.tomakehurst.wiremock.common;
 
-import java.util.Base64; 
+import javax.xml.bind.DatatypeConverter;
 
 public class Encoding {
 
     public static byte[] decodeBase64(String base64) {
         return base64 != null ?
-            Base64.getDecoder().decode(base64) :
+            DatatypeConverter.parseBase64Binary(base64) :
             null;
     }
 
     public static String encodeBase64(byte[] content) {
         return content != null ?
-            Base64.getEncoder().encodeToString(content) :
+            DatatypeConverter.printBase64Binary(content) :
             null;
     }
 }


tomakehurst:
Would mind sharing your method for measuring this?

Also, it looks like you're using some Java 8+ APIs. Is there a way to improve this without breaking Java 7 compatibility?

marsbradley:
Used JMC attached to a build of our stubbing framework to do flight recordings of around 20 minutes of typical load. Showed a high percentage (50%) of cycles spent in com.google.io.BaseEncoding. See image for more information.

AFAIK Java 8 is when base encoding was implemented natively...

However javax.xml.bind.DatatypeConverter has a very similar implementation (in terms of speed) and is available in Java 6/7. Could reimplement the PR using that library instead?

tomakehurst:
So for the profile you've shown there, you were using stubs with base64 encoded response bodies?

And do you know if it's encoding/decoding or both that are the bottleneck?

Just wondering if it's running hot because WM in unnecessarily re-encoding/decoding stuff more than once.

If you can get a significant performance improvement with `javax.xml.bind.DatatypeConverter` and keep Java7 compatible, that would be awesome.

Thanks!

marsbradley:
Yes.

Just looked at the flight recordings again, seems the encoding is roughly 2:1 compared to decoding in terms of % time spent.

Yes I believe that is a related issue; there does seem to be a fair amount of unnecessary encoding/decoding on top of Guava's implementation being slow. If you'll look back at the JMC trace, you'll see that the UTF-8 decoder is still being called frequently, which was certainly unexpected from our end.

If wiremock.http.body stored each response body in the format it was provided in, rather than always converting base64/string -> byte[] and then byte[] -> base64/string when it's needed again, it would reduce the overhead of all this encoding/decoding, no?

I will try and get a Java 7 fix for this slowness together soon. I think reducing the number of unnecessary encoding/decoding cycles would also have a big performance impact though.

tomakehurst:
OK, that'd be great. I'll try and find time to look into why excessive encoding/decoding is happening, and whether it's possible to fix without breaking backwards compatibility.

marsbradley:
Pushed a Java 7 compatible fix :).

tomakehurst:
Awesome! How much of a dent did that make on the original performance?

marsbradley:
A small batch of initial test runs showed a speed-up of roughly 1.6x with all bodies Base64 encoded. Base64 encoding/decoding now takes up ~20% of execution time.

Unfortunately it's again slower than just using non-encoded bodies, mainly due to the UTF-8 decoder still being called. Also note that these are very much initial results based off a very small sample size, so I would take them with a grain of salt. I will do more thorough testing once the excessing encoding/decoding has properly been looked into.

Cheers

tomakehurst:
Great, thanks for letting me know. I'll merge this for now, and try to carve out some time to look at the problem more deeply.

marsbradley:
Just to update you, now that I've run some longer/more realistic tests. As expected, initial results were a bit optimistic. Real-world speed-up is still around 1.2x-1.3x with Base64 encoded bodies though.

Before:
image

After:
image

Base encoding still takes up the lion's share of execution time (even using the java.xml library), but this is something that would be better solved through reducing excessive calls.

The time spent decoding has also increased relative to the time spent encoding (now they're roughly even, before far more time was spent encoding), which leads to the interesting possibility that java.xml is faster at encoding, but google.io is faster at decoding. I think the differences would be too small to warrant an investigation though.

tomakehurst:
Thanks for posting this, that's very useful. I'm happy to take a 1.3x speedup!

PR #808, opened 2017-11-14
Added a new API to selectively remove logged requests

msnishan:
Added a new API to selectively remove logged requests

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -324,6 +324,10 @@ public class WireMockServer implements Container, Stubbing, Admin {
         wireMockApp.resetRequests();
     }
 
+    @Override public void resetSelectedRequests(RequestPattern requestPattern) {
+        wireMockApp.resetSelectedRequests(requestPattern);
+    }
+
     @Override
     public void resetToDefaultMappings() {
         wireMockApp.resetToDefaultMappings();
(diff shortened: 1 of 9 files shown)


tomakehurst:
What are the use cases you have in mind for this?

msnishan:
PFB the use case details : 
We have a shared WireMock server and multiple developers run jobs to test the use cases. PFB the steps : 

- Add the behaviors to the WireMock server.
- Then eventually execute the tests (Tests are here service layer tests)
- The LoggedRequest objects are asserted to see whether the services were invoked with correct request objects or not.
- Finally reset the WireMock server

Now imagine a situation where 2 devs are running the job simultaneously. What might happen is, by the time 1st developer sets its behavior, the 2nd developer can reset the whole WM server.

That's why I have thought of a solution where each run would generate one unique header and after every run the dev would reset the LoggesRequests which corresponds to her/his  unique header value.

tomakehurst:
Right OK, I get the problem. And I've been asked about this before, so I think it's one worth tackling.

I'm wondering if there's a more general-purpose way to solve this. It would be nice to provide a solution that doesn't require a modification of the app under test.

One idea that's been kicking around for a while that might help here is stub tagging. One possibility would be to tag the stubs you're about to use from your test code as belonging to that developer/test process. Then you could reset all logged requests that were matched by those stubs. Provided a given stub would only be used by one developer/process at a time, and you weren't making requests for which no stub was present, this should work.

How do you create your stubs during testing? On the fly per test case, or up-front e.g. via JSON files?

msnishan:
Great suggestion. Yes, the stubs that we create are on the fly per test. There are some Ids which are auto-generated. But the we can have a tag created at the start of the test and that tag can be used at the end to reset all the logged requests belonging to that tag. That was the whole motto for this PR. Thanks for taking this matter seriously.

Regards

msnishan:
I will raise one more PR with new approach i.e with the tags. Closing this one.

PR #814, opened 2017-11-21
Fixed typo in _docs/extending-wiremock

hossameldeen:
maxLemgth -> maxLength

Note: I don't know why Github is showing me this as well in the diff:
```
-you can use the ```CollectingNetworkTrafficListener```.
+you can use the ```CollectingNetworkTrafficListener```.
```
But it doesn't show in rich-diff view & made sure they're the same sentences. Conclusion: Safe to ignore.


tomakehurst:
Thanks!

PR #816, opened 2017-11-21
Fix build on windows

anujbhatia:
This should fix the npm related build errors I got on windows


Azork:
Thanks for this!
For some weird reasons I had to install shx manually for this to work...
Not sure why :/

PR #821, opened 2017-12-01
Multipart Request patterns

nxtstep:
I've prepared a multipart request pattern matcher (`MultipartValuePattern`) to allow for verifying MultiPart request bodies and I'd love your feedback on the approach that I took 🙏 

The `MultipartRequestPart` is still quite naively implemented and probably needs to undergo some refactoring, nevertheless I wanted to share my preliminary work.

Clearly the tests for this addition are missing and I'm happy to follow-up or accept help on implementing (and fixing) tests.


tomakehurst:
Hi, thanks for submitting this. I've managed to have a bit of a look, so here are some 1st thoughts:

* Parsing the request body for parts ourselves seems unnecessary to me when the Servlet API provides the means to do this. It might be simpler to add `getParts()` and `isMultiPart()` to `Request` then implement these in the adapter by calling `HttpServletRequest.getParts()`.
* What is the purpose of the `name` field on the builder?
* It would probably make sense to use `MultiValuePattern` to match headers in the part, to be consistent with how headers are currently handled.
* Might also be worth adding the following methods to `RequestPatternBuilder` as shorthands:
  `withAllRequestBodyParts(<matcher>)`,
  `withAnyRequestBodyPart(<matcher>)`

nxtstep:
@tomakehurst Thanks for your response and feedback, much appreciated.

To come back at the points you mentioned:
- Making use of existing `getParts` methods would definitely be a better option, will look into that one.
- The `name` field on the MultipartBuilder is used to match that specific matcher against a specific multipart (request) value
- The `MultipartValuePattern` header matching is indeed done via `MultiValuePattern` (https://github.com/tomakehurst/wiremock/pull/821/files#diff-c01a4c1cde9d18941916596d8febcb3dR56) and the MultipartRequest header matchers are added as `StringValuePatterns` as are the normal Request headers (for consistency indeed) ;)
- I'm not sure if I understand what the All and Any shorthand methods should be doing. I assume something like all multiparts should match or just one of the list of MultipartValuePatterns? I think that deviates a bit from the current setup. As I assumed matchers per part not per request as a whole.

I'm definitely going to cleanup the (multi)parts parsing. Please advise on the other points?

tomakehurst:
What I meant about the header matching is that here you've used `StringValuePattern`: https://github.com/tomakehurst/wiremock/pull/821/files#diff-c01a4c1cde9d18941916596d8febcb3dR31

Regarding `name`, I can't see anywhere it gets read, other than when it's transferred from the builder to the `RequestPattern`. Am I just being dumb and not seeing it? I presume it's intended to aid with debugging?

The `all` and `any` methods would just be a slightly simpler shorthand when all you care about matching one or all of the parts' bodies. For instance, if I know that one part being sent will contain `{ "id": "abc123" }` then I can just say `withAnyRequestBodyPart(equalToJson("{ "id": "abc123" }"))`, avoiding a second level of nesting in the DSL. TBH, I can see the `any` version being more useful than the `all`, but hopefully you get the idea.

tomakehurst:
One further thing to consider when continuing with this: all of the stubbing calls WireMock supports have a JSON representation. Typically I end up writing unit tests for these e.g. the ones in `EqualToJsonTest`.

nxtstep:
Hey @tomakehurst I just worked a bit more on the Multipart pattern matchers. Current status is:

- [x] Using `HttpServletRequest.getParts()` for parsing and getting multiparts
- [x] Name in MultipartBuilder is used here: https://github.com/tomakehurst/wiremock/pull/821/files#diff-5537fd7ceab4d9ff17722e28b172e1bbR155
- [ ] `StringValuePattern` is still used instead of `MultiValuePattern`
- [ ] Create Unit-Tests for `MultipartValuePattern` and Builder
- [ ] Creating shorthand `RequestPatternBuilder` methods for any/all Multipart matchers
- [ ] Potentially extend existing Test-cases

nxtstep:
@tomakehurst I think it's ready 🙏

tomakehurst:
OK, this is generally looking good. I think a few things need tweaking before merge though:

* `javax.servlet.http.Part` should not appear in the `Request` interface, as this couples WireMock's API to the servlet API.
* A couple of acceptance tests would be good, just to show how the whole thing hangs together.
* Please could you add a section to https://github.com/tomakehurst/wiremock/blob/master/docs-v2/_docs/request-matching.md.
* I've pushed some tweaks, formatting and an extra test case to this branch, so please merge this into yours: https://github.com/tomakehurst/wiremock/tree/nxtstep-feature/multipart-post-request

Thanks!

nxtstep:
Sorted ✅

tomakehurst:
Merged, with a few tweaks. Thanks!

PR #823, opened 2017-12-08
Add ThreadPoolFactory to enable custom ThreadPool for Jetty.

mustaine:
Allow setting a ThreadPool for Jetty server.

Useful when needs to use a InstrumentedQueuedThreadPool.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
@@ -28,10 +28,10 @@ import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
-import com.github.tomakehurst.wiremock.admin.AdminTask;
 import com.github.tomakehurst.wiremock.common.*;
 import com.github.tomakehurst.wiremock.extension.ResponseDefinitionTransformer;
 import com.github.tomakehurst.wiremock.extension.responsetemplating.ResponseTemplateTransformer;
+import com.github.tomakehurst.wiremock.http.ThreadPoolFactory;
 import com.github.tomakehurst.wiremock.http.trafficlistener.DoNothingWiremockNetworkTrafficListener;
 import com.github.tomakehurst.wiremock.core.MappingsSaver;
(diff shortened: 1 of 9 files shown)


src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -195,6 +195,19 @@ public HttpServerFactory httpServerFactory() {
         }
     }
 
+    @Override
+    public ThreadPoolFactory threadPoolFactory() {
+        try {
+            ClassLoader loader = Thread.currentThread().getContextClassLoader();
+            Class<?> cls = loader.loadClass(
```
tomakehurst: Why do you need to do it this way, rather than instantiating it directly?

tomakehurst:
This looks good. Please see my one specific comment by the relevant line.

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
mustaine: It can be instantiate directly, I just copied the behaviour as 
```
    @Override
    public HttpServerFactory httpServerFactory() {
        try {
            ClassLoader loader = Thread.currentThread().getContextClassLoader();
            Class<?> cls = loader.loadClass(
                    "com.github.tomakehurst.wiremock.jetty9.JettyHttpServerFactory"
            );
            return (HttpServerFactory) cls.newInstance();
        } catch (Exception e) {
            return throwUnchecked(e, null);
        }
    }
```

because maybe I was missing something.

tomakehurst: OK, might be worth switching to direct if that works. Less likely to get broken by e.g. a package reshuffle.

mustaine:
If is not required, lets change both (httpServerFactory/threadPoolFactory) to be instantiate directly ?

mustaine:
@tomakehurst like that is all good?

tomakehurst:
Yep, that looks fine. Thanks.

mustaine:
Thanks @tomakehurst. It was quick.

PR #830, opened 2017-12-14
configuration for weblogic (swagger-ui not included)

Azork:
Hi, 

Maybe it would be useful for someone.
Here is the exploded war I used for weblogic deployment.

So far it works fine even though I had to use the standalone jar to make it work (wiremock cannot find the truststore without it somehow).

I didn't know where to put this configuration, so I put it in a different folder.

Regards, 

Azork

### DIFF
--- a/wiremock-weblogic/README.md
+++ b/wiremock-weblogic/README.md
@@ -1,0 +1,16 @@
+WireMock - Weblogic 12.1.3.0.0 installation
+======================================================
+
+[![Build Status](https://travis-ci.org/tomakehurst/wiremock.svg?branch=master)](https://travis-ci.org/tomakehurst/wiremock)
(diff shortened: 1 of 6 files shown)


Azork:
Ho, yeah, also, I'm using the exploded war to use the default __files and mappings folder.
I suppose this could be arranged by changing the WireMockFileSourceRoot but I haven't tried that yet.

tomakehurst:
I can't see any reason for this not to be a standalone project.

I'd be happy to signpost it from wiremock.org if you'd like to publish it separately, but I don't think this belongs in the core project.

Azork:
It's only configuration files to help somebody integrating wiremock into weblogic not to loose a couple of hours. 
A bit as the sample war you have.
I won't make a standalone project for this.

PR #844, opened 2018-01-07
add runtime switch for base64 encoder

### LINKED ISSUE #840: DatatypeConverter unavailable on Android
`javax/xml/bind/DatatypeConverter` is a class that is unavailable on Android.
Thus it breaks newer Wiremock versions which use `DatatypeConverter` for Android clients.

I propose to find another solution that does not rely on `DatatypeConverter` by default to make newer Wiremock versions compatible with Android. I've seen other projects also implement runtime checks if these classes exist in the current runtime and otherwise fall back to other implementations.

There might be other classes as well inside the code that are incompatible with Android. But removing `DatatypeConverter` would be a start.

Happy to contribute if you agree with finding a solution for this.

Mauin:
Resolves #840 

As discussed in the Issue this PR adds a runtime check if the `DatatypeConverter` class is available and otherwise falls back to the previous Guava implementation.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Base64Encoder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Base64Encoder.java
@@ -1,0 +1,6 @@
+package com.github.tomakehurst.wiremock.common;
+
+interface Base64Encoder {
+    String encode(byte[] content);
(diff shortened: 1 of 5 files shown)


tomakehurst:
Thanks!

PR #855, opened 2018-01-20
Add ResponseDefinitionBuilder#withCookie

alb-i986:
A convenience method to stub a *response* with a 'Set-Cookie' header.
(a withCookie method for request matching was already present: see RequestPatternBuilder)

PS. I see that you are not using `@since` in javadoc. I think it's a good habit to take, but if you don't like, I can remove it :)


alb-i986:
Hey @tomakehurst,
I tried to address your comments. Please check if it suits you.

PR #881, opened 2018-02-20
Fixed #880 - transformerParameters are applied to template

### LINKED ISSUE #880: TransformerParamerters are ignored
Transformer parameters for `ResponseTemplateTransformer` are not applied.

for example:
```java
    @ClassRule
    public static WireMockRule wireMockRule = new WireMockRule(wireMockConfig()
        .extensions(new ResponseTemplateTransformer(true))
        .dynamicPort());
    .....
    stubFor(get(urlPathMatching("/some/path/(.*)"))
                .willReturn(aResponse()
                        .withStatus(OK.value())
                        .withHeader(CONTENT_TYPE, APPLICATION_JSON)
                        .withBodyFile("some/file.json")
                        .withTransformerParameter("variable", "123456")));
```
```json
{"test": "{{variable}}"}
```

Will not replace `variable` in `file.json`

bduisenov:


tomakehurst:
I like this idea. But I think different groups of attributes should be under a top-level attribute (for the sake of namespacing). Please could you place these under a top-level key called `parameters`?

bduisenov:
@tomakehurst updated PR

tomakehurst:
Thanks!

PR #887, opened 2018-02-24
Support HTTP2 by using Jetty 9.4.8

samuel-soubeyran:
- Increase number of default thread in the pool to 13
- Add a connector for http2
- Use Conscrypt for ALPN Processor

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -35,7 +35,7 @@ def shouldPublishLocally = System.getProperty('LOCAL_PUBLISH')
 
 def versions = [
     jackson: '2.8.9',
-    jetty  : '9.2.22.v20170606',
+    jetty  : '9.4.8.v20171121',
     xmlUnit: '2.3.0'
 ]
 
@@ -48,6 +48,9 @@ dependencies {
     compile "org.eclipse.jetty:jetty-servlet:$versions.jetty"
     compile "org.eclipse.jetty:jetty-servlets:$versions.jetty"
     compile "org.eclipse.jetty:jetty-webapp:$versions.jetty"
+    compile "org.eclipse.jetty:jetty-alpn-conscrypt-server:$versions.jetty"
+    compile "org.eclipse.jetty.http2:http2-server:$versions.jetty"
+
     compile "com.google.guava:guava:20.0"
     compile "com.fasterxml.jackson.core:jackson-core:$versions.jackson",
         "com.fasterxml.jackson.core:jackson-annotations:$versions.jackson",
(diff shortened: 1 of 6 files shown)


tomakehurst:
As has been discussed on previous threads about this, the problem with upgrading beyond Jetty 9.2.x is that this drops support for Java 7, which is problematic both for enterprises with slow upgrade paths and Android devs.

I'd be happy to accept a solution that doesn't drop JDK7 support. Two options I can think of to achieve this are:

* Add another JAR artifact to the build that's Java 8+ only and uses the latest Jetty version. This would probably involve splitting the source tree up into core and version-specific parts, which the build would include/exclude for the various output JARs.
* Add some runtime detection of the environment i.e. are we on Java 8+ and Jetty 9.4+, and load the HTTP/2 code only under these conditions, such that `NoClassDefFoundError`s and the like aren't thrown when running on Java 7.

samuel-soubeyran:
👍 
Do you have any preference ?
The first one seems to be more involved but potentially cleaner. The second one could be pretty straightforward but will clutter the code base a little bit...

tomakehurst:
I'm leaning towards the second at the moment. I made some changes a while back that allow wiremock to work under jetty 9.2 and 9.4 (partly by doing ringtone detention) so it would be an extension of that.

samuel-soubeyran:
Ok that sounds good, I'll get started on that. If you have done something similar in the past, do you have any pointer or commits that I can look at ?

tomakehurst:
This is the code the dynamically detects which Jetty API we're using and acts accordingly, to give you a flavour of what I mean:
https://github.com/tomakehurst/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/jetty9/JettyUtils.java#L53

sunnycmf:
@tomakehurst i believe this concerns can be resolved by dockerlize this? i'm more than happy to contribute after this PR is accepted.

> the problem with upgrading beyond Jetty 9.2.x is that this drops support for Java 7, which is problematic both for enterprises with slow upgrade paths and Android devs.

sunnycmf:
@samuel-soubeyran are you planning to add GRPC support after this PR? perhaps we can work together as well. 👯‍♂️

tomakehurst:
@sunnycmf I'm not sure Docker would be much help in this instance, since it can't be used by Android devs or many people in slower-moving environments like banks.

samuel-soubeyran:
Yes docker wouldn't solve the problem,
 and yes the goal is to be able to use Wiremock to mock services over GRPC.

PR #888, opened 2018-02-26
Updates and fixes for RAML API spec

MasonM:
This makes some updates to the RAML specs, mainly to flesh out the recording API. I also made a few fixes so that running `api-spec-converter -f raml -t openapi_3 tmp/raml/wiremock-admin-api.raml` generates a valid OpenAPI 3.0 (aka Swagger 3) spec file: https://gist.github.com/MasonM/f70fc9feaf0c3cc4b49d337673c3b3f7. This requires upgrading [api-spec-converter](https://github.com/LucyBot-Inc/api-spec-converter) to get OpenAPI 3.0 support.

It's not valid Swagger 2, but that was already the case due to the use of `oneOf` for `delayDistribution`, which is not supported in Swagger 2. There isn't a good workaround for that.

The [latest version of Swagger UI supports OpenAPI 3](https://github.com/swagger-api/swagger-ui#compatibility), so I think it'd be a good idea to upgrade, but I didn't do that here.

### DIFF
--- a/docs-v2/_docs/record-playback.md
+++ b/docs-v2/_docs/record-playback.md
@@ -327,7 +327,7 @@ If you only require the IDs of captured stubs you can specify:
 By default generated stubs will be set to persistent, meaning that they will be saved to the file system
 (or other back-end if you've implemented your own `MappingsSource`) and will survive calls to reset mappings to default.
 
-Setting `persistent` to `false` means that stubs will not be saved and will be deleted on the next reset.
+Setting `persist` to `false` means that stubs will not be saved and will be deleted on the next reset.
 
 
 ### Repeats as scenarios
--- a/src/main/resources/raml/schemas/delay-distribution.schema.json
+++ b/src/main/resources/raml/schemas/delay-distribution.schema.json
@@ -1,0 +1,49 @@
+{
+    "type": "object",
+    "oneOf": [
+        {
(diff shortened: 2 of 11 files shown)


MasonM:
Here's a branch that upgrades to OpenAPI/Swagger 3.0, and updates Swagger-UI (since the current version doesn't support OpenAPI 3): https://github.com/MasonM/wiremock/compare/update-api-specs...MasonM:swagger-3-upgrade

Instead of updating the `src/main/resources/assets/swagger-ui/` files in-place, I deleted them all and had the `package.json` copy over the files from the [swagger-ui-dist](https://www.npmjs.com/package/swagger-ui-dist) module at build time. That should make future upgrades easier, since it'll just require updating `package.json`.

Screenshot:
wiremock

Let me know if you'd like me to enter this as a PR, or include it in this PR.

tomakehurst:
Awesome, thanks for taking the time to contribute these. I think upgrading to OAS 3.0 would be great. Maybe as a separate PR from the RAML stuff if that's not too much hassle?

MasonM:
@tomakehurst Sorry for the delay, but I entered PR #910 with those changes, though most of the changes in this PR are needed before api-spec-converter will successfully convert the RAML to OpenAPI 3.0.

Issue #511, opened 2016-10-10
Query string parameters are matched in order!


RobinKnipe:
I've been using the WireMock record feature to create a mock of a REST API. The generation works pretty well, except that the matching of query strings in URLs seems to be a basic string equality type of match. This is fine for single parameter query strings, but means that URLs with multiple parameters MUST honour the order used in the recording (`mappings`), which isn't much use. The query string should be considered as a list of name/value pairs and matched as such.

The following URLs should be considered identical:
 - `examlpe.com/path?number=1&string=hello&date=01-01-1970`
 - `examlpe.com/path?number=1&date=01-01-1970&string=hello`
 - `examlpe.com/path?string=hello&date=01-01-1970&number=1`
 - `examlpe.com/path?string=hello&number=1&date=01-01-1970`
 - `examlpe.com/path?date=01-01-1970&string=hello&number=1`
 - `examlpe.com/path?date=01-01-1970&number=1&string=hello`

tomakehurst:
You can modify your recordings to use query parameter matching if you require order invariance.

Agree it'd be preferable for the recorder to create stubs this way though.

RobinKnipe:
Thanks for getting back to me Tom. Could you give me an example of what you mean? How do I "modify recordings to use query parameter matching" ? Parameter order invariance is essential really.

tomakehurst:
Basically, hand-edit the JSON mapping files and add query parameter match attributes like this:
```json
"queryParameters" : {
  "search_term" : {
    "equalTo" : "WireMock"
  },
  "limit": {
    "equalTo" : "10"
  }
}
```

See http://wiremock.org/docs/request-matching/

RobinKnipe:
Thanks, that has got parameter invariance working which is good. Is there a way to get a case insensitive match? I could only find examples of `equalTo`, `contains`, `matches` and some JSONy stuff in the matching docs, and there didn't seem to be a way of passing the `i` flag to the regex matcher.

tomakehurst:
No way to do that at the moment unfortunately. Would also be a good addition.

HARIKRISHNAPALEMPATI:
@tomakehurst is `case insensitive match` still open issue?
I am trying to match based on firstname in request xml.

"bodyPatterns" : [
        {
            "matchesXPath" : "//FirstName[text()='TEST']"
         }
        ]

It is working fine if `FirstName` value is `TEST` and not working with `test`.

fraserharris:
[...]

PR #893, opened 2018-03-02
Added stubbing for multiple query parameters

bryantabaird:
I came across difficulties when wanting to create stubs for an unknown amount of query parameters to achieve parameter invariance. This addition will allow stubbing for multiple query parameters with a single method call.

I used existing test cases that had multiple `withQueryParam` as templates to test `withQueryParams`, and they seem to check out fine. Feedback is appreciated.

Somewhat relates to https://github.com/tomakehurst/wiremock/issues/511


tomakehurst:
Merged, not sure why GitHub didn't mark it automatically.

Thanks!

bryantabaird:
Hmm... strange.

Looks like it was merged in though, thanks!

PR #895, opened 2018-03-05
Use a InheritableThreadLocal for WireMock 

jcleezer:
After upgrading from 1.x to 2.x I saw an issue where tests that used a timeout were failing due to the configuration set in a before class being lost. This change moves the WireMock from a ThreadLocal to an InheritableThreadLocal so that configuration is preserved but can still be overriden by new threads.


tomakehurst:
Well, I never knew that even existed. Thanks!

PR #900, opened 2018-03-08
#898 Show the resulting port number instead of 0

### LINKED ISSUE #898: Dynamic port when running via .jar
TL;DR; Would be awesome if there was an ability for selecting a dynamic port when running the standalone jar.

### Why?
- It's required for running concurrent tests.
- It's possible when running the server [via Java](http://wiremock.org/docs/getting-started/#random-port-numbers)

### How?
Possibly with a `--dynamic-port` option, similar to `--port`.
There would need to be a way to feed the chosen port back to the process running the jar file.

### Workaround
Workaround might be to keep trying to start the jar with new ports till it doesn't exit with a non-zero exit code after x time. Doesn't sound very stable to me :stuck_out_tongue_closed_eyes: 

Alternate ideas very much welcome, only started researching wiremock today!

nicoschoenmaker:
Fixes #898

Starting WireMock with `--port 0` will denote the correct port in the output

Also, to prevent further questions like mine, added some documentation of the possibility of running with `--port 0` at
* the [running standalone page](http://wiremock.org/docs/running-standalone/#command-line-options)
* and when running with `--help`.

Feedback always welcome. Wasn't sure about tabs vs spaces, current file contains both... lemme know what you prefer.

### DIFF
--- a/docs-v2/_docs/running-standalone.md
+++ b/docs-v2/_docs/running-standalone.md
@@ -26,7 +26,7 @@ $ java -jar wiremock-standalone-{{ site.wiremock_version }}.jar
 
 The following can optionally be specified on the command line:
 
-`--port`: Set the HTTP port number e.g. `--port 9999`
+`--port`: Set the HTTP port number e.g. `--port 9999`. Use `--port 0` to dynamically determine a port.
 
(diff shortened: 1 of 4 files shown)


tomakehurst:
Thanks!

PR #908, opened 2018-03-22
Added a new API to selectively remove logged request

msnishan:
Added a new API to selectively remove logged request. This is in reference to the one I created some days ago:
https://github.com/tomakehurst/wiremock/pull/808


tomakehurst:
Looks pretty good, apart from the lack of tests.

One change I'd suggest making: rather than getting matching requests then deleting in `WireMockApp`, instead add a `deleteServeEventsMatching(RequestPattern requestPattern)` method to `RequestJournal` and call that. This is consistent with other methods on that class and is more of a "tell, don't ask" approach.

PR #910, opened 2018-03-28
Update to OpenAPI 3.0 and upgrade Swagger UI

MasonM:
**NOTE**: The fixes in PR #922, PR #923, PR #924, and PR #926 all need to be merged first, because api-spec-converter won't compile the current schema to OpenAPI 3.0.

This updates the Swagger JSON generation script in package.json to output the wiremock-admin-api.json in [OpenAPI/Swagger 3.0 format](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.0.md), which is supported in recent versions of [api-spec-converter](https://github.com/LucyBot-Inc/api-spec-converter).

I also updated [Swagger UI](https://github.com/swagger-api/swagger-ui) to the latest version, since the current version doesn't support OpenAPI 3.0. Although I could've extracted the new version in-place and overwritten the old one, I figured it'd be better to handle it in package.json by adding `swagger-ui-dist` as a dependency and copying over the necessary files during build time. This will make upgrades easier in the future.

Screenshot:
wiremock


MasonM:
Oh, and [ReDoc 2.0](https://github.com/Rebilly/ReDoc) supports OpenAPI 3.0, and I think it'd be a good replacement for raml2html. It's currently in alpha, but I tried it out with this branch (along with fixes in pr #888), and it works.

tomakehurst:
Does raml2html still work for OAS3.0 or will a switch be needed?

MasonM:
raml2html should be unaffected because it's called in package.json before the spec is converted to OAS3.0 by api-spec-converter: https://github.com/tomakehurst/wiremock/pull/910/files#diff-f16065c1bc7361e8a654329074f0ff33R49

I think ReDoc has a much better design and is easier to understand, but raml2html will continue to work

docs-v2/package.json
```diff
@@ -44,6 +45,7 @@
     "build:all": "npm run build:css && npm run build:js",
     "preprocess-schemas": "shx rm -rf tmp/raml* && shx mkdir -p tmp/raml && shx cp -rf ../src/main/resources/raml/* tmp/raml && shx rm -rf tmp/raml/schemas/* && node bin/bundle-schemas -i ../src/main/resources/raml/schemas -o tmp/raml/schemas",
     "raml2html": "npm run preprocess-schemas && raml2html -t lib/raml2html-template.nunjucks -i tmp/raml/wiremock-admin-api.raml -o _docs/wiremock-admin-api.html",
-    "apidocs": "npm run raml2html && shx mkdir -p ../src/main/resources/swagger && api-spec-converter -f raml -t swagger_2 tmp/raml/wiremock-admin-api.raml > ../src/main/resources/swagger/wiremock-admin-api.json"
+    "swaggerui": "shx cp -rf node_modules/swagger-ui-dist/ ../src/main/resources/assets/swagger-ui/",
+    "apidocs": "npm run raml2html && npm run swaggerui && shx mkdir -p ../src/main/resources/swagger && api-spec-converter -c -f raml -t openapi_3 tmp/raml/wiremock-admin-api.raml > ../src/main/resources/swagger/wiremock-admin-api.json"
```
MasonM: Note: I added the `-c` option to the call to `api-spec-converter` to have it check the schema is valid after conversion, which will help prevent typos and whatnot

tomakehurst:
Sorry, I'm being daft. I meant to ask whether RAML to Swagger conversion is possible for OAS3.0?

MasonM:
Once PR #922 is merged and I merge this branch with master, the conversion from RAML to OAS3.0 that's done in the `api-spec-converter -c -f raml -t openapi_3 tmp/raml/wiremock-admin-api.raml` call will start working.

If you mean converting from RAML to Swagger 2.0, it'll still be possible to run `api-spec-converter -f raml -t swagger_2 tmp/raml/wiremock-admin-api.raml` like it's currently doing, but it won't produce a valid schema. That's already the case because of the use of `oneOf` in `delay-distribution.schema.json` though.

MasonM:
Okay, I merged with master, updated the docs, and fixed the failing test. I also changed `baseUri: /__admin` to `baseUri: /` in ae146da0c291b9bc8f7188204978bff25372e2b3, because I realized it was confusing for Swagger UI to show each endpoint as `DELETE /<something>` when the actual path includes `/__admin`.

tomakehurst:
Sorry I've left this hanging for so long.

I tried checking it out today and building the docs, but it seems the RAML to Swagger conversion isn't perfect and a couple of problems are occurring:

1) The self-reference in content-pattern.schema.json can't be correctly resolved, leading to a duff `$ref` in the intermediate Swagger 2 doc
2) Even when the self-reference is removed, there are some `$ref` elements (the ones from inside schema files) that point to file names rather than paths under `definitions`.

I'm wondering, given the improvements in OAS3 and its popularity, whether it'd be best to ditch RAML support and make the OAS3 definition the authoritative one.

What do you think?

MasonM:
Yes, switching to OpenAPI 3.0 is probably the best long-term solution, but it raises a few questions:
1. Will there be a deprecation period for the RAML spec? I don't know how many people actually use the RAML spec, but I assume there's enough to warrant keeping it around until the next major version. This would mean keeping both specs updated until then.
2. The docs at http://wiremock.org/docs/api/ are generated using `raml2html`, so there'd need an alternative. I mentioned before that https://github.com/Rebilly/ReDoc is a good option. The branch below has this.
3. Will we the OpenAPI spec be in JSON or YAML? YAML will clearly make it easier to edit, but that means it'll have to be converted to JSON so that the `/__admin/docs/swagger` API endpoint is backwards compatible. The branch below also does this using the "js-yaml" NPM modue.

I made a a branch off master that adds a canonical OpenAPI spec in YAML format: 
https://github.com/tomakehurst/wiremock/compare/master...MasonM:swagger-3-upgrade-with-switch

It needs some fixes, but it's roughly as complete as RAML spec. Also, the branch updates http://wiremock.org/docs/api/ to use ReDoc. Screenshot:
screenshot_2018-10-29_20-56-14

If this approach is okay with you, I can finish it and enter a PR.

tomakehurst:
To answer your questions in turn (and sorry about the delay):

1) I guess in an ideal world that'd be the right thing to do, although I get the impression a lot more folks are using the derived Swagger spec than the RAML one. In any case they're both of limited usefulness at the moment for the reasons posted here, so to preserve scarce engineering time I'm tempted to drop RAML support and see what feedback arrives. It can always be reinstated, and there might even be an automated translation the other way.
2) Agree that would need fixing. If we did auto-generate the RAML from the Swagger then we could keep using `raml2html`. I've not tried ReDoc, but more than happy to give it a go if it's flexible enough to be styled and integrated into other pages. TBH it'd be a good opportunity to get rid of the slightly horrible iframe approach currently used.
3) I think maintaining YAML as the canonical version and automating the conversion to JSON is probably the right way to go.

More than happy for you to proceed with a PR, in fact this would be greatly appreciated!

MasonM:
It looks a lot longer than I expected, but I've finally got the new spec fully fleshed out and have ReDoc in a good state. Visit https://masonm.github.io/wiremock/docs/api/ to see what the new API docs look like.

To make it easier to code review, I've split this into 4 separate commits:

1. 9e80ddbefa5ae9cc311d59e1233da5f8c2085ca8 - Upgrade the Swagger UI instance to 3.0 using `swagger-ui-dist`, as in the original PR
2. 6ea49b5a9582454c63fd3aaae561847b7becf852 - Add the complete OpenAPI 3.0 spec and use it to generate the `swagger.json` file
3. 9e4bd9fdc6b8712c42ab0674599d3ae4e3594874 - Switch from raml2html to ReDoc.
4. 2e1e2f183c4f2255010b32203d4d61236ff3ef32 - Fully remove the RAML spec. If it turns out you want to preserve the RAML spec, this commit can be reverted.

tomakehurst:
Many thanks for contributing this.

I get a 404 from the redoc JS file when I serve the docs locally. Am I missing a step?

MasonM:
The ReDoc JS file is copied over by `npm run redoc`, which is called when `npm run apidocs` is run: https://github.com/tomakehurst/wiremock/pull/910/files#diff-f16065c1bc7361e8a654329074f0ff33R32

So, the following commands should let you view the API docs at http://localhost:4000/docs/api/:
```
git clone https://github.com/MasonM/wiremock.git -b swagger-3-upgrade 
cd wiremock/
./gradlew generateApiDocs
cd docs-v2
./serve.sh
```

tomakehurst:
Great, that works for me now.

Looks awesome!

I might tweak the fonts and spacings a little bit, but otherwise this looks like a goer.

Thanks again!

JesseKPhillips:
Where does this stand? I see travis is upset. If I bring this up-to-date would it be accepted to merge? I'm most interested in the Swagger UI update.

MasonM:
@JesseKPhillips I've been keeping it up date each time there's a conflict. There's no conflict right now, but I pushed a merge commit anyway.

The Travis failures were due to an unrelated build issue that @tomakehurst fixed in 6ecb765ed9a67b3db3ee20c5fe80a60180869a91

tomakehurst:
I've merged this now. I'm not sure why GitHub hasn't marked it as such, so I'm going to close it.

@MasonM apologies it's taken so long to get this onto master. I really appreciate the contribution though - this is hugely valuable!

MasonM:
@tomakehurst Thanks! No need to apologize; this was a huge PR.

PR #916, opened 2018-04-11
Update httpclient to fix hostname verification

jimirocks:
On some cases it's not possible to record using HTTPS, because of BUG
in httpclient, see https://issues.apache.org/jira/browse/HTTPCLIENT-1836

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,7 +52,7 @@ dependencies {
     compile "com.fasterxml.jackson.core:jackson-core:$versions.jackson",
         "com.fasterxml.jackson.core:jackson-annotations:$versions.jackson",
         "com.fasterxml.jackson.core:jackson-databind:$versions.jackson"
-    compile "org.apache.httpcomponents:httpclient:4.5.3"
+    compile "org.apache.httpcomponents:httpclient:4.5.5"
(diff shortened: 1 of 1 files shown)


tomakehurst:
Any reason not to go to 4.5.5 at this point?

jimirocks:
well probably nothing prevent 4.5.5 - commit updated

tomakehurst:
Awesome, thanks!

davidghiurco:
Hi @tomakehurst. Any chance the standalone wiremock jar could shade-relocate its httpclient dependency? I tried using the standalone version when I got a class loading conflict, but found that the conflict still persisted. 

I was about to open an issue on this but the issue guidelines indicate to not open issues for such problems. I was wondering what you might suggest I do.

tomakehurst:
@davidghiurco if it's not relocated already, then I think that's grounds for an issue to be raised.

Having said that I'm looking at the latest standalone JAR builds (2.23.2) and the Apache client classes are all under wiremock. as expected. Can you tell me where you're seeing un-relocated classes?

davidghiurco:
@tomakehurst you're absolutely correct, actually. My bad. 

 It wasn't wiremock's fault, it was unirest-java :) Thank you for confirming though.

PR #926, opened 2018-04-22
Add API schema for /__admin/near-misses/request

MasonM:
Split off from PR #888 to make it easier to review. This adds an API schema for `/__admin/near-misses/request`, which is required for OpenAPI 3 conversion to work.


tomakehurst:
Thanks again for all these!

Issue #475, opened 2016-08-01
TrustStore in classpath not supported


marcodave:
Hi, first of all thanks for WireMock, I think it's a great piece of technology!
I'm setting up an embedded server, started by a Spring Boot application. The configuration is externalized in another jar file, maintained separately. 
The SSL keystore and the truststore is saved into the config jar. 
The problem is that when I try to specify the path to the truststore as a classpath element (using `Resources.getResource("truststore.jks").toString()`), I get the following error
```
Caused by: java.io.FileNotFoundException: jar:file:\my\jar\file\path\config.jar!\truststore.jks (The filename, directory name, or volume label syntax is incorrect)
	at java.io.FileInputStream.open0(Native Method)
	at java.io.FileInputStream.open(FileInputStream.java:195)
	at java.io.FileInputStream.<init>(FileInputStream.java:138)
	at java.io.FileInputStream.<init>(FileInputStream.java:93)
	at com.github.tomakehurst.wiremock.common.KeyStoreSettings.loadStore(KeyStoreSettings.java:48)
	at com.github.tomakehurst.wiremock.http.HttpClientFactory.buildSSLContextWithTrustStore(HttpClientFactory.java:72)
```
I guess that the root issue is that a `FileInputStream` is used regardless of the posiiton of the TrustStore.
There is no such issue with the keystore, as it is also placed in the same jar. 
It would be nice to allow classpath elements also for the truststore.
Cheers

tomakehurst:
Yes, you're correct. At the moment it's not possible to load a trust store from the classpath unfortunately. I'll mark this up as a feature request.

michal-kroliczek:
Quite a bit time since opening this ticket. Any chance for implementation?

tomakehurst:
Do you feel like creating a PR?

marcodave:
Hi @tomakehurst I created a small PR to address the issue, #928

PR #928, opened 2018-04-23
SSL truststore tries to load from classpath

marcodave:
Fix for #475 
Added a fallback in `KeyStoreSettings` to try loading the truststore from classpath if not found as file.
Added relevant test cases


tomakehurst:
Thanks!

PR #936, opened 2018-05-12
Set Response and ResponseDefinition transformers by instance

### LINKED ISSUE #933: Configure stub mapping with transformer instances instead of names - Feature Request
When defining mapping-specific transformers programmatically, it is somewhat cumbersome to have to define them twice: in the extension list of the Wiremock server and in the mapping definition (see #932 for an example). 

It is also easy to forget that the extensions are applied globally by default and end up in confusion like #932 . 

It would be more convenient to set the instance of the transformer directly onto mapping without specifying it in the extension configuration, e.g.: 

```
WireMockServer wireMockServer = new WireMockServer(WireMockConfiguration.options()
            .port(8080));

Stub1 stub1 = new Stub1();
stubFor(any(urlEqualTo("/stub1"))
            .willReturn(aResponse()
                    .withTransformers(stub1)
            )
    );
```

Should such change seem acceptable I can look into contributing it.

mgurov:
Closes #933 

The motivation behind this change is to make in-proc configuration of the transformers less cumbersome. One can simply pass the instances of their transformers to a `ResponseDefinitionBuilder`, e.g.: 

```java
stubFor(any(urlEqualTo("/stub1"))
            .willReturn(aResponse()
                    .withTransformers(new MyResponseOrResponseDefinitionTransformer())
            )
    );
```

#### Benefits

* No need to register the transformations up-front at the configuration of the `WireMockServer`. This makes the composition of the code that sets the transformations easier.  
* Setting up a local stub is much more straightforward, no need to worry about the `AbstractTransformer.applyGlobally` flag. (I've seem and made myself confusions like [here](https://stackoverflow.com/questions/50231111/how-to-create-multiple-response-transformers-on-wiremock))

#### Potential issues

* This would obviously not work with Wiremock running as a standalone process, although could be extended for class names, similarly to the `WireMockConfiguration#extensions()` of the server configuration. 
* It can be potentially confusing that this new stub configuration method accepts global transformers, maybe a check for `false == AbstractTransformer#applyGlobally` would be a better idea.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/ResponseDefinition.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/ResponseDefinition.java
@@ -25,8 +25,10 @@ import com.github.tomakehurst.wiremock.common.Errors;
 import com.github.tomakehurst.wiremock.common.Json;
 import com.github.tomakehurst.wiremock.extension.AbstractTransformer;
 import com.github.tomakehurst.wiremock.extension.Parameters;
-import com.google.common.net.MediaType;
+import com.github.tomakehurst.wiremock.extension.ResponseDefinitionTransformer;
+import com.github.tomakehurst.wiremock.extension.ResponseTransformer;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.Objects;
@@ -50,6 +52,7 @@ public class ResponseDefinition {
     private final Fault fault;
     private final List<String> transformers;
     private final Parameters transformerParameters;
+    private final List<AbstractTransformer> transformerInstances;
 
     private String browserProxyUrl;
     private Boolean wasConfigured = true;
@@ -72,7 +75,26 @@ public class ResponseDefinition {
                               @JsonProperty("transformers") List<String> transformers,
                               @JsonProperty("transformerParameters") Parameters transformerParameters,
                               @JsonProperty("fromConfiguredStub") Boolean wasConfigured) {
-        this(status, statusMessage, Body.fromOneOf(null, body, jsonBody, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, wasConfigured);
+        this(status, statusMessage, Body.fromOneOf(null, body, jsonBody, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, Collections.<AbstractTransformer>emptyList(), wasConfigured);
+    }
+
+    public ResponseDefinition(int status,
+                              String statusMessage,
+                              String body,
+                              String base64Body,
+                              String bodyFileName,
+                              HttpHeaders headers,
+                              HttpHeaders additionalProxyRequestHeaders,
+                              Integer fixedDelayMilliseconds,
+                              DelayDistribution delayDistribution,
+                              ChunkedDribbleDelay chunkedDribbleDelay,
+                              String proxyBaseUrl,
+                              Fault fault,
+                              List<String> transformers,
+                              Parameters transformerParameters,
+                              List<AbstractTransformer> transformerInstances,
+                              Boolean wasConfigured) {
+        this(status, statusMessage, Body.fromOneOf(null, body, null, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, transformerInstances, wasConfigured);
     }
 
     public ResponseDefinition(int status,
@@ -90,8 +112,9 @@ public class ResponseDefinition {
                               Fault fault,
                               List<String> transformers,
                               Parameters transformerParameters,
+                              List<AbstractTransformer> transformerInstances,
                               Boolean wasConfigured) {
-        this(status, statusMessage, Body.fromOneOf(body, null, jsonBody, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, wasConfigured);
+        this(status, statusMessage, Body.fromOneOf(body, null, jsonBody, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, transformerInstances, wasConfigured);
     }
 
     private ResponseDefinition(int status,
@@ -107,6 +130,7 @@ public class ResponseDefinition {
                                Fault fault,
                                List<String> transformers,
                                Parameters transformerParameters,
+                               List<AbstractTransformer> transformerInstances,
                                Boolean wasConfigured) {
         this.status = status > 0 ? status : 200;
         this.statusMessage = statusMessage;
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/tomakehurst/wiremock/http/ResponseDefinition.java
```diff
@@ -72,7 +75,26 @@ public ResponseDefinition(@JsonProperty("status") int status,
                               @JsonProperty("transformers") List<String> transformers,
                               @JsonProperty("transformerParameters") Parameters transformerParameters,
                               @JsonProperty("fromConfiguredStub") Boolean wasConfigured) {
-        this(status, statusMessage, Body.fromOneOf(null, body, jsonBody, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, wasConfigured);
+        this(status, statusMessage, Body.fromOneOf(null, body, jsonBody, base64Body), bodyFileName, headers, additionalProxyRequestHeaders, fixedDelayMilliseconds, delayDistribution, chunkedDribbleDelay, proxyBaseUrl, fault, transformers, transformerParameters, Collections.<AbstractTransformer>emptyList(), wasConfigured);
+    }
+
+    public ResponseDefinition(int status,
```
mgurov: This constructor is a copy with an adaptation of the above as on one hand I need to pass the `transformerInstances` from the existing code usage, but on the other hand I don't want to affect JSON deserialization.

src/main/java/com/github/tomakehurst/wiremock/http/StubResponseRenderer.java
```diff
@@ -49,6 +50,8 @@ public Response render(ResponseDefinition responseDefinition) {
 		}
 
 		Response response = buildResponse(responseDefinition);
+		LinkedList<ResponseTransformer> responseTransformers = new LinkedList<>(this.responseTransformers);
```
mgurov: I use the out-of-fashion `LinkedList` here because of the recursion of the `applyTransformations`.

tomakehurst:
This generally looks great.

A few minor points:

* There probably ought to be some fail-fast behaviour if the user tries to use this when talking to a remote WireMock server. There's a precedent for this with custom matchers, and the strategy if simply to throw an exception if the call is remote. See here: https://github.com/tomakehurst/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java#L129
* The TODO needs to be resolved
* Please add a section to the docs showing how to use this
* `getResponseDefinitionTransformers()` and `getResponseTransformers()` in `ResponseDefinition` could be implemented more succinctly and consistently in a functional style using Guava's `FluentIterable`. There are many examples of this style throughout the codebase.


Thanks!

src/main/java/com/github/tomakehurst/wiremock/http/ResponseDefinition.java
```diff
@@ -334,12 +362,39 @@ public Fault getFault() {
     }
 
     @JsonInclude(NON_EMPTY)
-    public Parameters getTransformerParameters() {
-        return transformerParameters;
+    @JsonProperty("transformerInstances")
+    public List<String> getTransformerInstanceStrings() {
```
mgurov: I assume the JSON serialization is used for the `toString`, hence added this method. Please let me know if there is deserialization that can be negatively impacted by this new field. 

The reason to use `.toString` instead of exposing the transformers directly to JSON is that I prefer to play safe and avoid the situation where the custom transformation for whatever reason isn't nicely serializable, besides out of the box JSON serialization might not be very useful in many cases.

src/main/java/com/github/tomakehurst/wiremock/http/ResponseDefinition.java
```diff
@@ -359,18 +414,26 @@ public boolean equals(Object o) {
             Objects.equals(proxyBaseUrl, that.proxyBaseUrl) &&
             fault == that.fault &&
             Objects.equals(transformers, that.transformers) &&
+            Objects.equals(transformerInstances, that.transformerInstances) &&
```
mgurov: not all the fields are currently included to the `equals`/`hashCode`, but I assume by default a fields should be

src/main/java/com/github/tomakehurst/wiremock/stubbing/InMemoryStubMappings.java
```diff
@@ -67,33 +64,19 @@ public ServeEvent serveFor(Request request) {
...
+        ResponseDefinition response = matchingMapping.getResponse();
+
+		for (ResponseDefinitionTransformer transformer : response.selectApplicableResponseDefinitionTransformers(this.transformers.values())) {
+			response = transformer.transform(request, response, rootFileSource.child(FILES_ROOT), response.getTransformerParameters());
+		}
 
 		return ServeEvent.of(
             LoggedRequest.createFrom(request),
-            copyOf(responseDefinition),
+            copyOf(response),
```
mgurov: one more thing. We copy the `responseDefinition` only at the end of the transformation chain. That means that a not-so-nice transformation can mutate the state of our stub mapping, e.g.: 

```java
    @Test
    public void doesntProtectAgainstStubMutationByTransformers() {
        wm = new WireMockServer(wireMockConfig()
                .dynamicPort()
                .extensions(new ResponseDefinitionTransformer(){
                    @Override
                    public String getName() {
                        return "evil";
                    }

                    @Override
                    public ResponseDefinition transform(Request request, ResponseDefinition responseDefinition, FileSource files, Parameters parameters) {
                        responseDefinition.getTransformerParameters().clear();
                        responseDefinition.getTransformers();
                        return responseDefinition;
                    }
                }));
        wm.start();

        StubMapping stubMapping = wm.stubFor(get(urlEqualTo("/evil-mutation"))
                .willReturn(aResponse()
                .withHeader("MyHeader", "Initial")
                .withTransformerParameter("MyTransformerParam", "value")
                .withStatus(200)
                .withBody("Can you see MyHeader?")));

        assertThat(stubMapping.getResponse().getTransformerParameters().size(), is(1));

        new WireMockTestClient(wm.port()).get("/evil-mutation");

        assertThat(stubMapping.getResponse().getTransformerParameters().size(), is(0));

    }
```

not sure how bad is that though. Let me know if this would better be fixed by moving the copying to the line 67. Would prefer as a separate follow-up change though as this one has grown enough.

docs-v2/_docs/extending-wiremock.md
```diff
@@ -165,6 +165,16 @@ stubFor(get(urlEqualTo("/transform")).willReturn(
...
 
+Another convenience provided by the Java API is ability to add transformers instances, without prior registering them as an extension: 
+
+```java
+new WireMockServer(wireMockConfig()); //.extensions(not needed)
+
+stubFor(get(urlEqualTo("/local-transform")).willReturn(aResponse()
+        .withTransformers(new MyTransformer(), new MyOtherTransformer()));
+```
+Transformers added as instances to stubs will be applied locally. The result of `.applyGlobally()` and `.getName()` methods of such transformers will have no effect.
```
mgurov: Since we don't really care about the `.getName()` and `.applyGlobally()` of the transformers added as instances, maybe it would be a good idea to extract a `Transformer` interface detached from the `AbstractTransformer` and `Extension`?
The difficulty here is that the `ResponseTransformer` and `ResponseDefinitionTransformer` would also have to be detached from the current `AbstractTransformer` which might break existing code with `@Override` on `.getName()`. Leaving the worthness of the effort for others to evaluate.

mgurov:
Thanks for the encouraging feedback, @tomakehurst! 

I have updated the merge request according to it: 

* the `HttpAdminClient` adjusted to fail fast. Please check the exception message makes sence. 
* docs section on Non-global transformations has been extended. Please let me know if needs further adjustement or feel free to fix it up.
* I liked the `FluentIterable` and went on further refactoring of the methods that apply transformation. 
The selection of the applicable transformations has been moved to the `ResponseDefinition`, which is a nicer place for it IMHO. 
This allows making client methods much simpler. Let me know if I went too far. 
* TODO resolved. I am still not 100% sure about the equals/hashCode and toString adjustements though. Please check the inline comments.

I declare the change to be ready for another review round.

mgurov:
@tomakehurst please let me know if I could do anything to help accepting this pull request.

tomakehurst:
Nothing more you can do until I've reviewed it, which will hopefully be next week when my client workload eases up a bit.

mgurov:
Hey, @tomakehurst , this PR seems to have stalled. 

FYI: I'm personally fine with closing it since no progress for almost two years means it probably won't get merged realistically soon.

tomakehurst:
I'm sorry this hasn't progressed. As you can probably see, I've got a lot more plates to spin than hands at the moment.

I'm happy to keep this alive or close it, it's up to you. I think it'd be a useful addition, but it's fighting for bandwidth with a lot of other things that need doing.

PR #949, opened 2018-06-08
Fix build (tests) for users having a non-English default locale

jbretsch:
The tests AdminApiTest and EqualToXmlPatternTest assert English XML parser
error messages. To make those tests succeed for users having a non-English
default locale we set the default locale to English in those tests.

We refrain from setting the system property user.language in build.gradle
because that would only fix the tests when run from the command line, not when
run from some IDE.


tomakehurst:
Thanks!

Issue #950, opened 2018-06-11
Feature request: Support handlebars templates in body file paths


martin-sladecek-o:
Motivation: map requests to pre-defined responses, stored in files.
E.g.
stubFor(get(urlPathMatching("/resource/\\w*")).willReturn(
                ok().withTransformers(ResponseTemplateTransformer.NAME)
                        .withBodyFile("response/{{request.requestLine.path.[2]}}.json")));

dcunited001:
Thanks! i need this. is this in a build of the standalone jar?

tomakehurst:
Not yet, but will be soon. Couple of other things to add before I push a release.

eric1988:
@tomakehurst, do you have an idea when this feature will be released?

tomakehurst:
Fixed in 8a25659f2f854531b673db73937d60785609d5c6

PR #951, opened 2018-06-12
fix for #950 - supporting handlebars templates in body file paths

jainsahab:


tomakehurst:
Cool, thanks for this!

PR #956, opened 2018-06-19
get serve events by stub id

enivid:


enivid:
This is probably not exactly what you want so give me some feedback and I'll update it and then write some unit tests.

enivid:
I bet I could just do this with my own extension?

tomakehurst:
Yes, you could do it that way. This one I don't mind having as a core feature though. I just don't have much review time at the moment.

PR #958, opened 2018-06-21
Health

enivid:
Need a lightweight health check so load balancers etc can check the health of wire mock


tomakehurst:
This is a nice idea, but it doesn't need to be a core feature now that it's possible to add API routes via extensions.

Can I suggest you spin this out into a separate project?

enivid:
Alright. Just got it working using extensions. Thx

Issue #595, opened 2017-01-31
Request match priority


vperfilev:
Hi.
I using WM as standalone for mocking service with 50+ different URLs. I need to to logging requests, so i redirect console output into file. But file is grow extreemly fast (800+ kb/per call). It's because there are too many warnings like:
`2017-01-31 12:15:55.518 Warning: JSON path expression '$.createRequest.contacts.point[?(@ == '')]' failed to match document '{"accountIdentifier":"15"}' because of error 'Missing property in path $['createRequest']'
2017-01-31 12:15:55.545 Warning: failed to parse the XML document. Reason: Content is not allowed in prolog.
XML: {"accountIdentifier":"15"}`

But URLs of this request rules are different with called URL. And if URL pattern is not match why body patterns is checked? Or did I do something wrong?

vperfilev:
Can someone check this error?

tomakehurst:
Are you still seeing this issue?

vperfilev:
I'm sorry, but I can't check it. I've solve it by writing simple pipe filter. And now project is done and I cant restore environment.

tomakehurst:
OK, let's close this off for now then. Happy to re-open it if you see a repetition of the problem.

Bdthomson:
I am coming across this same issue.

tomakehurst:
Can you provide details to reproduce?

Bdthomson:
Let's say I have 10 matchers. 9 of them use `matchesXPath` like so:
```javascript
"bodyPatterns": [
  {
    "matchesXPath": "/x:y [@z='ABC']",
    "xPathNamespaces": {
      "x": "http://www.x.net"
   }
 },
```.

The last one uses `matchesJsonPath` like so:

```javascript
  {
    "matchesJsonPath": {
      "expression": "$.username",
      "contains": "admin"
    }
  },
```.

What I am seeing when I send my only json request is that it tries to match against the XML Paths even though the urls and content-type are different.


```
2018-07-11 16:03:41.283 Warning: failed to parse the XML document. Reason: Content is not allowed in prolog.
XML: {"username": "admin"}
```

Bdthomson:
Any ideas when this issue can be be re-opened or prioritized?

tomakehurst:
Reopened. As for prioritised....depends on all my other priorities unless someone else fancies stepping up to take a look and opening a PR.

Bdthomson:
[...]

Issue #444, opened 2016-06-29
Please add instructions for how to build and install locally


dsyer:
I want to try the latest version, but it's not in a public Maven repo (as far as I know). Happy to build it locally, but "gradle install" doesn't work, and neither does "gradle publish". Something in the README maybe?

tomakehurst:
It will be shortly. Just takes a while to sync across.

You can temporarily use oss.sonatype.org if you want it immediately.

dsyer:
Thanks. What about if I make changes locally. How can I install those and test them?

tomakehurst:
I've added some basic build instructions to the README

tomakehurst:
Looks like install is broken somehow by the presence of two artifacts. I'll add it to the list to fix...

Bdthomson:
Any idea how to get around this and install to local maven repository? Trying to test local changes also.

PR #968, opened 2018-07-17
Adding isJson check for xpath matchers

Bdthomson:
- Adds dependency on json library.
- Adds `isJson` check in Xpath matcher that will prevent a sax parser error getting logged on json data.

Issue #595 

I cannot install locally to test this (I believe because #444 has not been resolved yet and it is not readily apparent how to do so).


tomakehurst:
Sorry it's taken so long to look at this.

I'm not keen on adding another JSON lib to the compile dependencies just for this. As there's already one present (Jackson). Could Jackson's event-based parser be used to do this cheaply?

@philippe-granet I think the issue with your suggestion is that numbers, string literals, booleans etc. are all also valid JSON documents (albeit rarely used at the top level).

philippe-granet:
See #1049 , I simply check if data start with a `<`. If not, it's not XML, so don't match

tomakehurst:
Positively checking for XML rather than negatively for JSON seems like a much better approach.

Bdthomson:
Closing in favor of https://github.com/tomakehurst/wiremock/pull/1049

PR #969, opened 2018-07-18
return unzipped string body when it is zipped

pityq:
I am getting zipped body for a proxied request in a listener. As WireMockHttpServletRequestAdapter is doing it already I thought would be nice to have it in the response as well for consistency.


tomakehurst:
This shouldn't ever happen. The GZip filter should ensure you've always got plain text internally.

Please can you raise a bug if you're still seeing this with details of how to reproduce?

PR #979, opened 2018-08-19
Remove temporal coupling between run() and isRunning()

automatictester:
`isRunning()` no longer throws NPE if executed prior to `run()`.


tomakehurst:
Thanks!

PR #983, opened 2018-08-24
Add base url method to wiremock server

wieczorekm:
I wanted to get url like `http://localhost:34567` but `url(String path)` method allowed only to get url  with a tailing slash (when I passed empty string), so I created method named `baseUrl()` that allows to get wiremock url without a slash at the end.


wieczorekm:
Tests are passing - just false negative due to `com.github.tomakehurst.wiremock.common.FatalStartupException: java.lang.RuntimeException: java.net.BindException: Address already in use`. I can't retigger them AFAIK.

tomakehurst:
Thanks!

PR #991, opened 2018-09-19
Implemented recording of multipart requests

victor-cr:
Also fixed an incorrect import in MockRequestBuilder


tomakehurst:
Thanks, this looks good!

PR #1017, opened 2018-10-24
Add more info to admin request log

zheng-wang:
Since the mock server is shared by multiple people, and enterprise prefers to audit everything, need to add more info to the admin request logs, like client IP address, request method, headers and body.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java
@@ -72,7 +72,7 @@ public abstract class AbstractRequestHandler implements RequestHandler, RequestE
         stopwatch.stop();
 	}
 
-	private static String formatRequest(Request request) {
+	protected String formatRequest(Request request) {
 		StringBuilder sb = new StringBuilder();
 		sb.append(request.getClientIp())
 				.append(" - ")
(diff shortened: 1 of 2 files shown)


zheng-wang:
Closing and opening the PR for a re-build.

tomakehurst:
Please can you add a test case for this

tomakehurst:
Yeah, this has suddenly been failing a lot. Not sure why it's got so much worse recently. I'll put it on my list to look at next time I get some WireMock time.

tomakehurst:
I'm actually planning to move everything to CircleCI at some point, so that might help as it seems a bit faster than Travis.

zheng-wang:
@tomakehurst 
I have added test case. Can you please review and let me know whether anything more is needed? 

By the way, PostServeActionExtensionTest failed again. I guess you have the privilege to re-build without close-reopen the PR. If so, please help with a re-build. Thanks

tomakehurst:
Can I suggest taking a look at other tests that check correct notifications?

Creating your own mock impl of `Notifier` seems overkill and there are a number of others that just use JMock for the same effect. It'd be good for this test to be consistent with those.

zheng-wang:
I have replaced custom impl of Notifier with a jMock created one. Would you think it is ok now, or further improvement is needed?

tomakehurst:
Looks good, thanks!

PR #1021, opened 2018-10-30
Support other keystore types when running from the command line.

kingsfleet:
This is possible with the API; but not when invoking standalone so this just added the extra command line parameters. The default changed in JDK 9 to PKCS12 so having the option to override it is useful.

https://blogs.oracle.com/jtc/jdk9-keytool-transitions-default-keystore-to-pkcs12


kingsfleet:
Not clear how I broken this test on one JDK, this passed locally, any ideas?

tomakehurst:
Sorry, flakey test I haven't got round to squashing. I've re-run it now.

kingsfleet:
As is a dupe of https://github.com/tomakehurst/wiremock/pull/999

kingsfleet:
Both pass now I guess.

PR #1022, opened 2018-10-30
Added ability to capture all headers while recording

aleks-azen:
Added the ability to capture all headers while recording by setting a parameter in the RecordSpec class. Useful when you want to simply see what calls are coming in or for very specific request matching

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/recording/CaptureAllHeadersRequestTransformer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/recording/CaptureAllHeadersRequestTransformer.java
@@ -1,0 +1,30 @@
+package com.github.tomakehurst.wiremock.recording;
+
+import com.github.tomakehurst.wiremock.client.WireMock;
+import com.github.tomakehurst.wiremock.http.Request;
+import com.github.tomakehurst.wiremock.matching.EqualToPattern;
+import com.github.tomakehurst.wiremock.matching.RequestPatternBuilder;
+import com.google.common.base.Function;
+
+public class CaptureAllHeadersRequestTransformer implements Function<Request, RequestPatternBuilder> {
+    private final RequestBodyPatternFactory bodyPatternFactory;
+
+    public CaptureAllHeadersRequestTransformer(RequestBodyPatternFactory bodyPatternFactory) {
+        this.bodyPatternFactory = bodyPatternFactory;
+    }
+
+    @Override
+    public RequestPatternBuilder apply(Request request) {
+        RequestPatternBuilder builder =
+            new RequestPatternBuilder(request.getMethod(), WireMock.urlEqualTo(request.getUrl()));
+
+        for (String header : request.getAllHeaderKeys()) {
+            builder.withHeader(header, new EqualToPattern(request.getHeader(header), true));
+        }
+        byte[] body = request.getBody();
+        if (bodyPatternFactory != null && body != null && body.length > 0) {
+            builder.withRequestBody(bodyPatternFactory.forRequest(request));
+        }
+        return builder;
+    }
+}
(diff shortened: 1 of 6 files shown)


tomakehurst:
This seems like quite a lot of code added for something that doesn't seem all that useful. I can't see why you'd want to constrain a stub on all of the headers a typical UA would send, and if you just want to see which requests you're receiving you can turn verbose logging on.

mrk-han:
Could this be useful for verifying analytic request timing? Screen tracking requests, etc

valb3r:
@tomakehurst In my opinion, recording all request headers to the stub, so that it can be edited later offline is a very useful feature.

I have the following setup:
WireMock user does not know exact requests/responses being done to the remote API (i.e. they contain sensitive information). 

If I could:
1. Deploy WireMock as a proxy to serve requests to some remote server
1. Record complete stubs with all headers
1. Edit the stubs to remove unneeded headers
1. Use edited stubs using an automated process to obfuscate sensitive data (i.e. simple sed on the stub to replace known sensitive fields before sending back stub to developers)
This would solve my problems.

Without this feature, I need to parse WireMock logs for headers obfuscation (that are not that easy to parse compared with JSON stub).

tomakehurst:
Sorry, I still think this is a bad idea.

Based on previous experience, I think it'll lead to a lot of Stackoverflow questions about why recorded stubs aren't matching anything.

Also, how many request headers are really useful as matching constraints? I find it hard to believe any API has so many that fall into this category that the workflow you've described would yield any kind of time saving.

valb3r:
@tomakehurst 
The problem is that I don't know header names before running WireMock or any other kind of request-response capture (and some of them may or may not contain sensitive information).

Typically it is 3-5 header names to be used as constraints (HATEOAS-alike links and IP/Geolocation)
I understand that using this feature without understanding the consequences of the requirement of a complete header set match is a problem, but it is optional setting not enabled by default. 

How about having a regex filter for headers to record?

tomakehurst:
I'd suggest that if you don't know which request headers the target API cares about, then it would be a good idea to learn a bit more about it before hitting record.

I already see questions fairly regularly from people who've hit record without really understanding what its doing and therefore why it's not producing the expected result on playback. This, I think, would just tempt people into another click-it-and-pray situation.

Also, if you're keen to inspect headers before committing to recording, you can already do this by making a load of requests through a proxy stub then snapshotting them after taking a look at the request log.

valb3r:
@tomakehurst 
Ok, thanks for quick replies and great tool!

Just a couple of final questions regarding this issue:
1. I'm going to proceed with the fork of this repository to include the desired functionality. I'll implement header recording into stub (in the fork) by recording all not-blacklisted headers (compared to whitelist functionality that is present in WireMock now)
`--record-headers-without-user-agent` - record headers excluding `Cache-Control`,`Content-Encoding`, etc.
`--record-headers-excluding=<List of headers to exclude>`
Does it make sense to contribute with this functionality in this repository?

1. I see that there is some work done on the plugin system for WireMock (and I could use the plugin system for the pt.1 to solve my problem):
https://github.com/tomakehurst/wiremock/pull/1032
Is there anything I could help with plugin system implementation?

tomakehurst:
You're welcome to do what you wish in your own forks certainly, but I don't want to merge what you're suggesting.

I think the plugin system PR is a long way from being usable. TBH I think the creator and I have some slightly conflicting design goals.

There is already an extension mechanism though - have you looked at whether this is enough for your requirements?

Another thing that might help - the legacy recorder (the one activated via CLI parameters) is just a `RequestListener` impl. You could just subclass/clone and modify that if you want to get the behaviour you've described.

valb3r:
> There is already an extension mechanism though - have you looked at whether this is enough for your requirements?

Thanks, looks like it is a better idea compared with building a fork

> Another thing that might help - the legacy recorder (the one activated via CLI parameters) is just a RequestListener impl. You could just subclass/clone and modify that if you want to get the behaviour you've described.

:+1: Oh, that's great, probably will go with this one

PR #1023, opened 2018-10-31
Support request matcher extensions during verification

modax:
Prior behavior was very confusing. If request matcher extension name was to passed to verify(), that assert would never succeed but WireMock would not say that request matcher extensions are not supported during verification or anything like that ...


modax:
A few notes:

* Tests failing do not seem to be the fault of this PR
* If Parameters contains some non-json friendly value, execution over remote wire will fail. But I'm not sure you intended to support this use case anyway.

tomakehurst:
There was a flakey test causing the build failure, nothing to do with this PR. I've made a couple of improvements so hopefully it won't happen going forward.

modax:
@tomakehurst Thanks. I have rebased my PR branch on the latest master.

Issue #1004, opened 2018-10-04
Load extensions at runtime


elfogre:
Hello Wiremock team,

I'm working with your fantastic tool and due to some requirements I'm thinking about to write a new feature. I firstly want to know your point of view and if this change (well tested) could be merged with your project.

What I need is to load extensions at runtime and be able to reset and load them the same way mappings could be reset (post __admin/mappings/reset). The idea is create a folder in wiremock settings, like 'mappings' and '__files', that could be 'plugins'. Inside this folder you could add your extensions jar's and the idea is to load them at the start and in every reset with URLClassLoader and Reflections to get classes that implement helpers. Once you get them you instatiate and load extensions with a Map with this plugins.

What do you think about this?

tomakehurst:
I like this idea. I've wanted to improve extension loading for some time by doing something like this, so it would be a welcome addition.

What would you propose as a way to tell the extension mechanism which `Extension` class files to load?

I'd thought about a properties file with a well-known name, or possibly something in META-INF, but I've not thought it through all that much. I'm not so keen on classpath scanning where it can be avoided as it can add a lot to startup times.

PR #1032, opened 2018-11-07
Plugin system for Helpers and Extensions

elfogre:
Hello Wiremock team,
Some weeks ago I write to you in https://github.com/tomakehurst/wiremock/issues/1004 talking about a plugin system to load at runtime Extensions in Wiremock.

We've (Tuenti) implemented a first version of this. It lacks of documentation now and need some changes I want to discuss with us, but most code is wrote :)

It searches for a folder "__extensions" (like "__files" does). It expects there a json that defines Helpers and Extension to be load at runtime ([example](https://github.com/tomakehurst/wiremock/compare/master...tuenti:master#diff-c348274057e9b8df95fad5ddbf07ac88) for tests). Extensions can even be parametrized with arguments also defined in json.

Test at [here](https://github.com/tomakehurst/wiremock/compare/master...tuenti:master#diff-d0fdef11f8594fac0f1132d76f39ce27) shows how to load extensions dinamically, just a call to wireMockConfiguration.reloadFileExtensions()

However there are some things to do. What about other Options implementations? Where should be reloadFileExtensions() called (now is only in wiremockApp Options constructor)? Do you like json format of helpers and extensions?

Hope this helps and we can contribute to make even better your project :D


tomakehurst:
This is by no means exhaustive, but here are some initial thoughts and requests:

* Please don't bump the version - I'll do that when I'm ready to release
* Why did you need to unshade handlebars?
* I'm not keen on exposing Handlebars for testing in ResponseTemplateTransformer. I suspect there will be other ways to achieve the same effect without busting encapsulation.
* `Options` is probably the wrong place for `reloadFileExtensions()` to live
* Suggest using `Json.read()` rather than new up an `ObjectMapper` when parsing Json, that way you get a consistently configured, and potentially cached `ObjectMapper` instance.
* Could `Parameters` be reused, rather than introducing a new `Argument` class?
* Could we put extension arguments in WireMock's existing global settings object instead of defining them in yet another place?
* Suggest making value classes immutable unless absolutely necessary - this is the style WireMock is moving towards slowly, despite the fact there are a few exceptions around.
* Avoid checked exceptions (e.g. `PluginInitializationException`)
* A lot more testing needed - maybe a child project is needed, since we want to simulate the situation where both the extension and WireMock are running from their respective JAR files.
* I'm generally not sure about using JSON for extension config in the way you've defined it - it makes the simpler case i.e. I just want to load a couple of JARs a bit painful
* Wondering whether hot reloading is worth the hassle - is there a killer use case for this?

elfogre:
Hello again, thanks for your fast reply. About those points:

- Ok, I've using it in my repo as version 2.20 and I forget to move back again.
- As Helper is an interface from Handlebars, I decided to unshade it to avoid problems related to ClassNotFoundExceptions with plugins and shadowed version of Handlebars.
- As handlebars is not accesible otherwise in ResponseTemplateTransformer, the other way to do this is with reflection.
- I think so. As I don't know as well as you Wiremock architecture, could you please point me where is a good place for this method?
- Yep, better to use Json.read()
- Argument is simpler for this case, and even can used to instantiate params with type from another class, Argument only supports Integer, List, String and Boolean as I see. Maybe Argument is better for this.
- I need you to explain me better this point.
- Ok.
- Something like make PluginInitializationException unchecked (the only one introduced in this PR)? It should be catched somewhere to allow Wiremock load with just a warning saying Extensions can't be loaded.
- Definitely more tests must be done. Now is loading plugins from a jar, but just as a unit test, not integration.
- As all other config like mappings is done in json I use this format to map definitions classes. If you prefer another one just sugest it.
- Really hot reloading is not needed. As we can start it with different helpers and extensions without the need to compile it is enough. 

I'm going to make now the clear points and wait for you to clarify the more difficult ones.

PR #1037, opened 2018-11-13
Add xsd matching support

aleks-azen:
XSD allows for matching of XML body structure instead of exactly matching the XML. For example if there are date fields in a request I may just want to validate that the field was indeed a date but not care what date it was. This change is a pretty simple implementation since XMLUnit already supports this type of validation

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToXmdPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToXmdPattern.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (C) 2011 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


tomakehurst:
**Some Observations**

1) It should probably be named `MatchesXsdPattern` since it's a match, rather than equality operation
2) A few extra test cases are needed - at a minimum an acceptance test case and a test case showing that a sensible diff report is printed when a request doesn't match due to the XSD.
3) I'm currently working on making it possible to use custom matchers in combination with standard ones, so I'd like to weigh the pros/cons of pulling this out into an extension project, rather than adding to the core.

**Considerations for 3):**
Can you see future value in adding configuration to XSD matchers e.g. controlling leniency?

How many users are likely to need this? More one for me to think about, but if it's not that great a % my leaning is generally towards building extensions.

aleks-azen:
pr got a bit messy, closing for now. will reopen with more tests. I do believe that there is a demand for this as XSD is one of the better ways to match dynamic XML without a huge effort from the user

Issue #1042, opened 2018-11-27
Multiple stubs per json file


rpax:
## Feature requests

Would be great to have multiple stubs in the same file (for example, in an array). Something like the following json:
```json
[
  {
    "request": {
      "method": "GET",
      "url": "/some/thing"
    },
    "response": {
      "status": 200,
      "body": "Hello world!",
      "headers": {
        "Content-Type": "text/plain"
      }
    }
  },
  {
    "request": {
      "method": "GET",
      "url": "/some/thing2"
    },
    "response": {
      "status": 200,
      "body": "Hello world! 2",
      "headers": {
        "Content-Type": "text/plain"
      }
    }
  }
]
```
This is particularly useful for grouping stubs, and having a manageable amount of files.

PS: Thanks for the great work!

tomakehurst:
Closing as a duplicate of https://github.com/tomakehurst/wiremock/issues/987

eschulma:
In case anyone runs across this trying to get multiple mapping into one json file as I did, be aware that the suggested format above is not what ended up in the code. There is no requests array, just mappings. See

https://github.com/tomakehurst/wiremock/blob/master/src/test/resources/multi-stub/multi.json

for a working example.

Pons0192:
> In case anyone runs across this trying to get multiple mapping into one json file as I did, be aware that the suggested format above is not what ended up in the code. There is no requests array, just mappings. See
> 
> https://github.com/tomakehurst/wiremock/blob/master/src/test/resources/multi-stub/multi.json
> 
> for a working example.

Thanks this helped me :)

PR #1043, opened 2018-11-27
Multiple stubs per json file

rpax:
Solves #1042


tomakehurst:
Thanks for contributing this.

I was actually about to start working on this myself, as I've got a few specific things I'd like it do to.

These are - detecting whether a file is single or multiple so you can mix both (and maintain backwards compatibility), supporting multiple stub creation via the API and possibly a multiple scenarios format. 

So I'll close this off, but I'll be sure to tag it once my changes are in.

tomakehurst:
Implemented in abfe4e6340fb5ce57a24b7da6a7dd7d00b48cf46

PR #1046, opened 2018-11-28
Admin API: Setting Scenario States

CanYumusak:


tomakehurst:
Hi, thanks for this.

A few things this could probably do with:

* Support in the client class (`WireMock`), both static and instance.
* Acceptance tests that exercise this through the client (again, static and instance)
* Consider using `PUT` rather than `POST`, identifying the scenario in the URL and therefore avoiding the need for the `ScenarioState` class.

PR #1059, opened 2019-01-09
Support XMLUnit placeholders in EqualToXmlPattern

zheng-wang:
When doing stub request body XML equality matching, I need the ability to ignore certain field values. For example, I have two requests

    <message>
      <id>12345</id>
      <content>Hello</content>
    </message>

    <message>
      <id>67890</id>
      <content>Hello</content>
    </message>

I want to define a stub mapping with request body like

    <message>
      <id>#{xmlunit.ignore}</id>
      <content>Hello</content>
    </message>
    
so that both requests match.

Using MatchesXPathPattern is technically doable, but its user experience is not so good as with the placeholder feature. XPath is a logical indicator while the placeholder is a visual indicator, and XPath might involve specifying namespaces while placeholder does not need them.


zheng-wang:
From implementation perspective, I haven't found a way to apply the placeholder settings globally (like via the WireMockConfiguration), so currently can only apply the settings per EqualToXmlPattern instance. It would be great if you know a way to do it globally, and then the StringValuePatternJsonDeserializer and WireMock classes might not need to be changed.

tomakehurst:
I think this is a great idea.

A few points about your implementation:
* The test case in `WireMockTest` should be in `EqualToXmlPatternTest`
* There should be an acceptance test
* It would be nice to pick some sensible default delimiters so you don't have to explicitly specify them
* A section in the documentation (https://github.com/tomakehurst/wiremock/blob/master/docs-v2/_docs/request-matching.md) explaining how to use this would be great.

tomakehurst:
Regarding local vs. global settings, I'll give it some thought.

Ideally, we'd just add some config parameters as usual, but there are possibly some practical difficulties with this (including that you've alluded to with the deserializer).

zheng-wang:
About the implementation

- The test case in WireMockTest has been moved to EqualToXmlPatternTest
- An acceptance test returnsMatchWhenTextNodeIsIgnored_AcceptanceTest has been added to the EqualToXmlPatternTest
- XMLUnit uses default delimiters `${` and `}` if we don't specify custom ones. I have added two *_DefaultDelimiters tests to the EqualToXmlPatternTest to demonstrate this.
- A section in the documentation (https://github.com/tomakehurst/wiremock/blob/master/docs-v2/_docs/request-matching.md) explaining how to use this has been added.

Let me know if you need anything else.

tomakehurst:
Rather than making the `EqualToXmlPatternTest` class into an acceptance test class (by running a WireMock server etc.), please could you move the acceptance test case to an existing acceptance test class (which are all in the `com.github.tomakehurst.wiremock` package). I'd suggest `StubbingAcceptanceTest` as there's already an XML equality test case in there (line 485).

tomakehurst:
A couple more things:
* The acceptance test should demonstrate that the overload of `equalToXml` in the DSL works remotely - i.e. you're using the static DSL off `WireMock` rather than calling something like `wm.stubFor(...)` against the rule (which is a direct method call, and therefore doesn't exercise the full JSON serialization/deserialization path).
* A link in the doc to the XMLUnit placeholders doc page would be really useful.

zheng-wang:
As requested, moved the acceptance test to the StubbingAcceptanceTest class. Exercised stub JSON serialization and deserialization in the two new acceptance tests.

There seems no XMLUnit document on the placeholder feature. Currently only the [unit tests](https://github.com/xmlunit/xmlunit/blob/master/xmlunit-placeholders/src/test/java/org/xmlunit/placeholder/PlaceholderDifferenceEvaluatorTest.java) tell the usages. I'll try to work with XMLUnit author to create the document.

zheng-wang:
A link in the doc to the XMLUnit placeholders doc page has been added. Hopefully it looks better now.

D0rmouse:
Would like this in a new version please 👍

zheng-wang:
Hi @tomakehurst , would you be able to have a check on this PR and see whether it can be accepted?

tomakehurst:
I pulled down your latest changes last night, so I'll see if I can find a few minutes today to take a look.

zheng-wang:
Thanks.

tomakehurst:
Rebased into 7f92d26039398ef22697443a2a059244ac61b41a and merged

PR #1060, opened 2019-01-09
Fix request model documentation in response-templating document

pmihalcin:
Remove .requestLine substring

### DIFF
--- a/docs-v2/_docs/response-templating.md
+++ b/docs-v2/_docs/response-templating.md
@@ -123,23 +123,23 @@ The model of the request is supplied to the header and body templates. The follo
  
 `request.url` - URL path and query
 
-`request.requestLine.path` - URL path
+`request.path` - URL path
 
-`request.requestLine.pathSegments.[<n>]`- URL path segment (zero indexed) e.g. `request.pathSegments.[2]`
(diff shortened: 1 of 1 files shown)


tomakehurst:
Not sure why you want to do this - the `requestLine` part is correct. Am I missing something?

pmihalcin:
@tomakehurst when I was trying to use `requestLine`, it wasn't resolved correctly.

Simply put, 
just compare the `description` and `example` you have in doc:
`request.requestLine.query.<key>` vs `request.query.search` - there is no `requestLine` in the example

The same applies everywhere where `requestLine` is used.

tomakehurst:
Are you on the latest version?

pmihalcin:
I see.. how can I view documentation for `com.github.tomakehurst:wiremock:2.6.0`? sorry for the noise

tomakehurst:
I'd strongly suggest upgrading to 2.20.0 if you can. It's fully backwards compatible.

However, if you absolutely can't, you can view the docs in GH by tag: https://github.com/tomakehurst/wiremock/blob/2.6.0/docs-v2/_docs/response-templating.md#the-request-model

pmihalcin:
Thanks. I upgraded to the latest version. 
Now both `request.requestLine.query.goto` and also `request.query.goto` work

Issue #1051, opened 2018-12-26
Let `WireMockServer#baseUrl()` return an URL with `options.bindAddress()` as host rather than `localhost`


circlespainter:
https://github.com/tomakehurst/wiremock/blob/34d1a911746e5adc1d6eb030bf3d089b3a0c65b7/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java#L205

tomakehurst:
Sounds like a very reasonable idea. Feel like creating a PR?

PR #1063, opened 2019-01-12
Implement FR #1051 "Let `WireMockServer#baseUrl()` return an URL with…

circlespainter:
… `options.bindAddress()` as host rather than `localhost`"


tomakehurst:
Normally you'd want this to be `localhost`. I can see that there are some rare exceptions, but it'd be good to see a test case that confirms that in the absence of the bind address being set, the default behaviour is still to return `localhost` as before.

PR #1066, opened 2019-01-19
Add method to verify zero interactions in the WireMock client

EugeneYushin:
Just a handy method to verify there're no interactions with mock server at all.
It differs from `verify(int count, RequestPatternBuilder requestPatternBuilder)` in the way it doesn't care of request pattern.
For example, this can be useful to verify that client's app doesn't involve any kind of network interaction when it finds data in local cache. 
Inspired by: https://github.com/mockito/mockito/blob/release/2.x/src/main/java/org/mockito/Mockito.java#L2244


oleg-nenashev (commented):
Not sure adding such a static is aligned with the API vision. Otherwise still looks relevant. Will review with @tomakehurst

Issue #456, opened 2016-07-08
parallel request in scenarios (stateful) - race condition 


kzimnicki-zalando:
Hi 
Currently I'm facing some strange behavior: 

https://gist.github.com/kzimnicki-zalando/98ebcf140d578b1bb74222b179d0f69e

It will return: 

> 1
> 1
> 2
> 1
> 2
> 2
> 2
> 2
> 2
> 2
> 2
> 2

My expectations is that first request will change scenario from 1 to 2 and then for other requests it will return only 2. In general there should be only  one "1" in output (order is not expected). Looks like we have race condition here. 

My version: 2.1.4-rc3

tomakehurst:
Thank you for reporting this. I'll investigate.

ryanb93:
@tomakehurst any update with this? We're seeing the same problem. Given a scenario where Request 1 updates the State, when we make 2 asynchronous calls at exactly the same time - on my local machine the tests pass but on a faster CI machine it fails due to the 'second' request still hitting the old return value.

```scala
stubFor(
        requestMatching
          .inScenario(scenario)
          .whenScenarioStateIs("STATE_1")
          .willReturn(aResponse.withBody("1").withStatus(response.status))
          .willSetStateTo("STATE_2")
)

stubFor(
        requestMatching
          .inScenario(scenario)
          .whenScenarioStateIs("STATE_2")
          .willReturn(aResponse.withBody("2").withStatus(response.status))
          .willSetStateTo("STATE_3")
)

List(request1, request2).map(request => callReturningAFuture(request))
```

Both responses contain "1" body.

tomakehurst:
Since they're async WRT to each other, there's no way WireMock (or any other service) could guarantee the state change triggered by one would always be seen by another, short of implementing serializable transactions i.e. locking the Scenario per request.

Do you see similar behaviour if you make the requests in series like the OP of this issue?

gabfssilva:
I'm experiencing the exact same issue.

I have two futures hitting WireMock concurrently, and, that's the exact behavior I wanted to test. For now the workaround is to send the requests in series in order to achieve the wanted behavior.

cvmocanu:
First, let me start by saying thanks for Wiremock. It's a good library, and I used it a lot.

Not to this issue - I just hit it myself: scenarios are not thread-safe.
This severely limits the usefulness of scenarios - basically the first time I needed them, I found out that I can use them.

[...]

PR #1069, opened 2019-01-23
Add synchronized to make state change is thread safe

letylo:
Hello,

This PR is to make the stateful behavior transition thread safe.

Regards,
Leticia López


tomakehurst:
This will produce a major bottleneck in serving stubs, and there's no tests demonstrating that it fixes the problem, so I can't accept it in its current form unfortunately.

letylo:
This test has been created to check if the patch is working.
https://github.com/letylo/WireMockThreadSafe

The test creates two states, the initial state returns "1" and changes to the second, that returns "2".
Then are executed 1000 threads to read from WireMock and the values obtained are added to a list.
At last, it checks that the list contains only one "1".

If there is a race condition, the list will contains more "1", while if not, there will be only one "1", disregarding its position in the list.

In the first part the program executes 10 runs to warm up the JVM, then it starts recording the time for each run. 

The histogram its showing how the difference between the values using WireMock and patch is negligible.

out

lines

Moreover, this patch resolves the unexpected behavior from WireMock, since for a webserver, it is **supposed** to handle correctly multiple parallel requests, independently of the internal threading model implemented and of its performance.

tomakehurst:
I'm still not convinced you're not introducing a bottleneck as you're only creating two stubs in your test case, so the time spent in the `find(...)` method (and therefore inside the synchronized block) is minimal.

I suspect if you added 1000 stubs (and ensured your load was distributed across them), then performance would degrade quite considerably.

letylo:
I reran the test stubbing from 10 to 1000, taking the average of ten runs, and this is the result:

average-runs

As you can see, the patch doesn't had noticeable overhead until 500, and in that case is 10 ms.
This means for common usage ( < 30 stubs) there is practically no difference between the patch one and the standard one.

Lastly, I would stress that before entering a discussion about performance, it's more important to have a software working correctly than a faster one that doesn't.

tomakehurst:
There are two problems with your test that I think render it invalid:

1) You're measuring the time interval to submit a task to an executor, not the execution time of the task. You'd need to move capture of the start/end time (lines 93 and 111) inside the lambda passed to `submit()` to correct this.
2) By not explicitly setting a pool size on the executor, the concurrency level will end up being whatever the executor feels is optimum, which will probably be too low to cause any real contention around the sync block.

I'd also suggest the following for a better test generally:

* The stubs (in addition to their being lots of them) should require some work to match e.g. `POST` requests using `equalToXml` matchers on the request body with a non-trivially sized XML document.
* I'd avoid chaining them into a scenario in the way you have as this isn't typical usage.

letylo:
About the first point, is true that we measured just for the for loop and the we fixed that moving up the latch two lines up, to measure the time of all threads.

This is the updated plot, 

average-runs

As you can see the difference is less that the previous one, thus proving that there isn't a major bottleneck.

About point 2, if you look at the javadoc, the executor start a new thread if there are jobs in the queue, so it creates a lots of threads. Moreover, the threading is enough to exhibit the concurrency issue, as shown by the first test.

However, this PR is about the concurrency problem raised by #456 and not about the performance.
I am more than happy to talk about other race condition problems that aren't resolved by this PR, but I will not follow the debate about possible performance issues, since it's pointless to talk about performance when the problem is a race condition that prevents the **correct** execution of tests.

tomakehurst:
Your latest code doesn't address my first point - it still measures time to enqueue the work, not time to execute it.

tomakehurst:
While I'm not contesting generally that a fix for this would be a good thing, synchronising on a singleton on the read path definitely isn't the right way to achieve this.

I'll give you an example - suppose I have lots of stubs and it takes ~ 100ms on average to find a match when a request arrives (I've seen setups where this is the case). The synchronized block ensures that only one stub can be searched for at a time, so in this scenario the maximum throughput is only 10 requests per second.

A better solution would be to make use of a concurrent data structure such as `AtomicReference` that supports safe parallel modification of state without serial levels of contention.

letylo:
> Your latest code doesn't address my first point - it still measures time to enqueue the work, not time to execute it.

My code is measuring the total time because the `latch` will block until all threads have been executed and called `latch.countDown()`, since it's initialized with the same thread size, and only then will read the end time.

https://github.com/letylo/WireMockThreadSafe/blob/275b7ffdfad59f69b109f81172f6097f42344df0/src/test/java/com/github/letylo/StubForSyncTest.java#L90-L113

letylo:
> While I'm not contesting generally that a fix for this would be a good thing, synchronising on a singleton on the read path definitely isn't the right way to achieve this.
> 
> I'll give you an example - suppose I have lots of stubs and it takes ~ 100ms on average to find a match when a request arrives (I've seen setups where this is the case). The synchronized block ensures that only one stub can be searched for at a time, so in this scenario the maximum throughput is only 10 requests per second.
> 
> A better solution would be to make use of a concurrent data structure such as `AtomicReference` that supports safe parallel modification of state without serial levels of contention.

Here I see two problems:
1) It would be better to not have a singleton over all the request, but a multilevel dispatch to first select which state-machine is needed (if any) and then apply the synchronization only to that the state-machine.
2) If the `find()` takes 100 ms, the bottleneck isn't about the `synchronized` but on the very `find()` method. In that case I think it would require a rewrite of the whole structure in order to be faster on finding the elements it needs, independently of the concurrency problem. Probably it's a wiremock 3.x worth change.

I understand your worries about performance, but also in the case you highlighted (100 ms/request) then, you will need the clients to coordinate between each others and serialize their request in order to get the correct response. So, in the end, you will still be limited to a throughput of 10 req/sec and putting the coordination burden to the clients.

tomakehurst:
To your first point - you're correct, I'd misread the placement of the latch, so it looks like you're measuring the time to n loops.

However the earlier point still stands about your test being non-specific about concurrency. A better setup would be to create a fixed thread pool (i.e. making the concurrency level explicit), rather than letting the executor decide for you. Likewise, measuring throughput in addition to latency is essential.

To your later points:
1. You may be right that this is a better design, and it's something I would consider accepting as a patch.
2. I agree there are likely to be faster (and correspondingly more complex) search algorithms out there. However, I don't agree that's justification to impose an artificial bottleneck, especially while we're still doing a linear search. Regardless, there are always going to be use cases that are heavy on compute e.g. doing full XML equality matches on large documents.

letylo:
I have an idea about the thread-safety in the scenarios.
I see you have an interface called Options with input values. I could add into this interface one more option: as a boolean value to mark if they want to use a thread-safety scenarios implemetation or not.
Then, the option will add a synchronized wrapper around the class InMemoryStubMappings to give thread-safty or not.

In this way whoever wants have the performance can have it, and whoever needs thread-safe scenarios can select the other.

I understand this is an API change in Options, but I think it is worth and it's a minor addition.

If it's OK for you I can write that change for this PR.

PR #1082, opened 2019-02-15
Adds cache support for body response template.

max3163:
Only text or external file body template are cached.

When always cached, it is done synchronously (increase startup time and
REST API call).

Refresh only apply to modification done on external files.

To activate cache, adds "cache" property to mapping response object with
values:

* "Always": cache when mapping is added/replaced
* "OnCall": cache only after first call
* "Never" (default): no cache

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
@@ -46,6 +46,7 @@ public class ResponseDefinitionBuilder {
 	protected List<String> responseTransformerNames;
 	protected Map<String, Object> transformerParameters = newHashMap();
 	protected Boolean wasConfigured = true;
+	protected CacheStrategy cacheStrategy;
 
 	public static ResponseDefinitionBuilder like(ResponseDefinition responseDefinition) {
 		ResponseDefinitionBuilder builder = new ResponseDefinitionBuilder();
@@ -68,6 +69,7 @@ public class ResponseDefinitionBuilder {
 			Parameters.from(responseDefinition.getTransformerParameters()) :
 			Parameters.empty();
 		builder.wasConfigured = responseDefinition.isFromConfiguredStub();
+		builder.cacheStrategy = responseDefinition.getCacheStrategy();
 		return builder;
 	}
 
@@ -148,6 +150,11 @@ public class ResponseDefinitionBuilder {
 	public ResponseDefinitionBuilder withTransformer(String transformerName, String parameterKey, Object parameterValue) {
 		withTransformers(transformerName);
 		withTransformerParameter(parameterKey, parameterValue);
+		return this;
+	}
+
+	public ResponseDefinitionBuilder withCacheStrategy(CacheStrategy cacheStrategy) {
+		this.cacheStrategy = cacheStrategy;
 		return this;
 	}
 
@@ -256,7 +263,8 @@ public class ResponseDefinitionBuilder {
 						fault,
 						responseTransformerNames,
 						transformerParameters,
-                        wasConfigured) :
+                        wasConfigured,
+                        cacheStrategy) :
 				new ResponseDefinition(
 						status,
 						statusMessage,
@@ -273,7 +281,8 @@ public class ResponseDefinitionBuilder {
 						fault,
 						responseTransformerNames,
 						transformerParameters,
-					    wasConfigured
+					    wasConfigured,
+					    cacheStrategy
 				);
 	}
 }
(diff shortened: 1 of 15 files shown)


tomakehurst:
Thanks for this.

A few observations:
* I think the `CacheStrategy` per stub approach adds too much complexity. Setting it overall via configuration might be better.
* I'd suggest putting the listener methods for stub addition/update etc. in their own interface and make that a new extension type, rather than adding them to the transformer interface (the templating transformer can still implement the methods).
* Stub removal should trigger cache eviction (if required). The above interface would therefore also need `onStubRemoval()` and `onReset()`.
* Tests are failing: https://travis-ci.org/tomakehurst/wiremock/jobs/493765113
* Some additional tests are needed covering each of the cache strategies, and ensuring that the cache state is correctly managed when stubs are changed in any way. Important to ensure it won't leak memory over time as stubs are added/updated/deleted.
* Would possibly benefit from some way to limit the cache size via configuration?

loganmzz:
Hi,

Sorry for this late answer, I was a little busy with other stuff:

> I think the `CacheStrategy` per stub approach adds too much complexity. Setting it overall via configuration might be better.

As caching has impact on performance (load time and memory usage), I think its better to have fine grained settings.

> I'd suggest putting the listener methods for stub addition/update etc. in their own interface and make that a new extension type, rather than adding them to the transformer interface (the templating transformer can still implement the methods).

Exactly what I thought when writing code but not wanted to add too much in this PR. It has some impacts on internal architecture and don't know codebase enough to add such features. Listener capapbilities need a dedicated work on whole project not this specific case.

> Stub removal should trigger cache eviction (if required). The above interface would therefore also need `onStubRemoval()` and `onReset()`.

Reset is already implemented in `com.github.tomakehurst.wiremock.stubbing.InMemoryStubMappings#reset` ([code](https://github.com/tomakehurst/wiremock/pull/1082/files#diff-1aa578413fd6757e6181f1cdec45680eR154)). For removal, it's quiet more complex as a single file can be used for many stubs. And don't know how many times it is used. It can be counted but transformers lacks information on how to identify source `StubMapping`. I will take a look on how to modify code to pass `StubMapping` identity to transformers.

> Tests are failing: https://travis-ci.org/tomakehurst/wiremock/jobs/493765113

Sorry, even before development, tests were not passing in our environments. Now, I have a stable one, I will fix tests.

> Some additional tests are needed covering each of the cache strategies, and ensuring that the cache state is correctly managed when stubs are changed in any way. Important to ensure it won't leak memory over time as stubs are added/updated/deleted.

I see idea and already test a lot of cases manually. I understandard needs for automatic test but at user level, it's only an empirical method based on response time measurement. I still try to add some test at different levels.

> Would possibly benefit from some way to limit the cache size via configuration?

According to final goal (having predictable response time), I'm not sure it's a good idea.

loganmzz:
Tests are fixed. I have found some bugs. I'm working on listener + deletion support

loganmzz:
I have added support for cache removal based on stub mapping remove/update

I still need to add some tests regarding body file modification

loganmzz:
Dont understand while Travis is failing. I test locally with:

* Oracle JDK 1.8.0_141-b15 (Linux x64)
* Oracle JDK 1.7.0_80-b15 (Linux x64)

All tests are passing ...

loganmzz:
HI @tomakehurst ! Can you help to complete this PR ? Thank you very much !

tomakehurst:
I've fixed the sources of recent build flakeyness, so a merge from master should help with the test failures.

These points need to be addressed before I'm willing to merge this:

* The cache policy per stub adds too much complexity and more importantly inappropriate coupling (between the core stub model and template extension). This should go, and instead we should have a simple on/off configuration switch plus a parameter to cap the overall cache size. Ideally the cache should implement a LRU eviction strategy.
* If you think the additional listeners work should happen under a separate change, then I'm fine with that, but it should come first and this should build on it. Adding them to the transformer interface changes the public API in way that can't be changed until a major version bump (and even then without annoying the people who maintain extensions).
* `InMemoryStubMappingsCacheTest` is an acceptance test in terms of its design, so it should be named as such and placed in the same package as the other acceptance tests.
* The reset problem needs a solution, otherwise the cache will become a source of memory leakage in setups where stubs are repeatedly created and bulk deleted without stopping WireMock.

loganmzz:
> The cache policy per stub adds too much complexity and more importantly inappropriate coupling (between the core stub model and template extension). This should go, and instead we should have a simple on/off configuration switch plus a parameter to cap the overall cache size. 

As caching has impact on performance (load time and memory usage), I think its better to have fine grained settings.

> Ideally the cache should implement a LRU eviction strategy.

According to final goal (having predictable response time), I'm not sure it's a good idea.

> If you think the additional listeners work should happen under a separate change, then I'm fine with that, but it should come first and this should build on it. Adding them to the transformer interface changes the public API in way that can't be changed until a major version bump (and even then without annoying the people who maintain extensions).

I have already updated this part and added an interface that can be implemented optionally. I had missed the type was part of extension mechanism and perfectly understand the point. I already did the same on Apache JMeter for some needs.

> `InMemoryStubMappingsCacheTest` is an acceptance test in terms of its design, so it should be named as such and placed in the same package as the other acceptance tests.

Ok.

> The reset problem needs a solution, otherwise the cache will become a source of memory leakage in setups where stubs are repeatedly created and bulk deleted without stopping WireMock.

I don't understand the problem. Reset is already implemented since first submision. And I have added removal case since your last comment.

tomakehurst:
Closing, since template caching was added in 2e92e590920a39ab1ba030dffbb8eb197752f1cd

PR #1086, opened 2019-02-21
docs: Fix testCompile -> testImplementation 

fuji-151a:
Hi, I modified `testCompile` to `testImplementation`.
Because `testCompile` is deprecated.
Details: https://docs.gradle.org/4.10/userguide/java_plugin.html

Please check this pull request.
Thank you very much.

### DIFF
--- a/docs-v2/_docs/download-and-installation.md
+++ b/docs-v2/_docs/download-and-installation.md
@@ -69,25 +69,25 @@ Java 8 standalone:
 Java 7:
 
 ```groovy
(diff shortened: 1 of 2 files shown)


tomakehurst:
I agree this needs to change, but I'm wondering if it's a bit premature. I still see a lot of teams using older versions of Gradle that don't support the newer syntax, whereas while `testCompile` is deprecated, at least still works.

rieske:
`testCompile` has been deprecated for quite a while and it is removed altogether in Gradle 7.0 which was released beginning of 2021.

Issue #1077, opened 2019-02-05
Performance issue with templating


max3163:
# Performance issue with templating 

I am encountering a problem of performance when I active the reponse templating on wiremock, especially when the response content is important. ( with or without handlebar function in the content )
By example, for a full text response of 500Kb, I got a response time of 21 second ! ( with the last version of wiremock 2.21 it's better 14sec but still inconceivable )
I've check where we lost so much time, and it's on the handlebars lib. I've ask help on the github of this lib, the answer is to activate a cache. I've check on the code of Wiremock, no cache is implemented. 
But I can't figure out how the process can take so much time to just parse a string of 500Kb.

- [ ] WireMock V2.20 and V2.21
- [ ] Wiremock Standalone with theses options --no-request-journal --container-threads=200 --jetty-acceptor-threads=100 --async-response-enabled --async-response-threads=100 --local-response-templating  --root-dir data
- [ ] Active templating with a response size about 500Kb
- [ ] ![l_3892 tmp](https://user-images.githubusercontent.com/13835366/52265841-3c1c3d80-2935-11e9-9fec-a1093af49e61.PNG)

tomakehurst:
Wow, that is especially bad. I've never seen it take anything like that long.

Any chance you can post the template (or something similar) that caused the problem?

tomakehurst:
Thanks, I'll try and take a look at that next time I'm in the code.

I suspect caching compiled templates might be a good idea, but I need to think a bit about memory usage and ensuring invalidation at the right moments.

max3163:
Hi, 

I implemented the cache in Handlerbars. Result are totaly predicatble and better on second call ( about 60ms ). This solution is not entirely satisfactory because lost 15 sec on the first call in a benchmark test is still not acceptable.

I've implement in wiremock an other lib of templating called pebble which is close to the handlebars one, results are really great !

Response time :

|Size|Handlebars|Pebble|
|----|-------------|---------|
|500kb| 15 sec | 70 ms |
|1Mb| 35 sec | 70 ms |

max3163:
See PR #1082 to implement cache

philippe-granet:
See also PR #1088 to only template datas between first '{{' and last '}}'

tomakehurst:
Closing since the associated PR has been merged

PR #1088, opened 2019-02-21
Only template datas between first '{{' and last '}}'

philippe-granet:
See #1077 and #1082 
See also https://github.com/jknack/handlebars.java/issues/681

Response time with a text file without any {{ }}:

|  Size  | Without this PR | With this PR  |
| ------------- | ------------- | ------------- |
|  500kb  | 15 sec | 50 ms |
|  1Mb | 35 sec | 50 ms |


tomakehurst:
Nice idea!

Maybe worth extending `BaseTemplate` rather than implementing `Template` so that you get some of the method impls for free?

philippe-granet:
All `Template` classes are package-private and in an internal package `com.github.jknack.handlebars.internal`.
Did you prefer implementation in my last commit?

tomakehurst:
Whoops, hadn't spotted it was package private. Clearly I've been looking at Groovy classes too much recently.

PR #1092, opened 2019-02-25
Custom handlebar template standalone

JoePullen:
Add feature to supply custom handlebars helpers to standalone by extending global-response-templating and local-response-templating command line parameters to also be supplied with a comma separated list of classes which implement an enum of helpers, where the name of the enum is also the name of the function. eg

```
      public enum Helpers implements Helper<String> {
        echo {
            private EchoHelper helper = new EchoHelper("echo");
            public Object apply(String value, Options options) throws IOException {
                return helper.apply(String.valueOf(value), options);
            }
        };
 
        @Override
        public Object apply(String value, Options options) throws IOException {
            return value.toString();
        }
    }

    public static class EchoHelper extends HandlebarsHelper<Object> {
        private final String name;
        public EchoHelper(String name){
            this.name = name;
        }

        @Override
        public Object apply(Object value, Options options) throws IOException {
            return value.toString().toUpperCase();
        }
    }
```
Also add a command line parameter to allow the setting of the handlebars TemplateCache.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -206,7 +206,6 @@ subprojects {
         relocate "org.custommonkey", "wiremock.org.custommonkey"
         relocate "com.flipkart", "wiremock.com.flipkart"
         relocate "net.sf", "wiremock.net.sf"
-        relocate "com.github.jknack", "wiremock.com.github.jknack"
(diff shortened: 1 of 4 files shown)


JoePullen:
Trigger rebuild

PR #1096, opened 2019-03-11
Fix XML bad performances on matching XPath

philippe-granet:
With this test case:
```java
        Xml.optimizeFactoriesLoading();
        String mySolarSystemXML = "<solar-system>"
            + "<planet name='Earth' position='3' supportsLife='yes'/>"
            + "<planet name='Venus' position='4'/></solar-system>";
        StringValuePattern pattern = WireMock.matchingXPath("//planet[@name='Earth']");
        for(int i=0;i<100000;i++) {
        	match = pattern.match(mySolarSystemXML);
        }
```
This take about 20s without my PR and about 10s with my PR.  

About 50% of CPU time is wasted in instanciate DocumentBuilderFactory and DocumentBuilder :
cpu_waste
This PR put DocumentBuilderFactory and DocumentBuilder instances in a ThreadLocal (they are not Thread Safe) cache.


tomakehurst:
Thanks!

PR #1098, opened 2019-03-17
Replace usage of internal java.xml class with public abstraction

kr5ture:
This commit fixes problem of using `DocumentBuilderFactoryImpl` from `com.sun.org.apache.xerces.internal.jaxp` package. Problem arises as this package is not public (not exported) since JDK 9. Solution is to replace use of this class with `DocumentBuilderFactory` from exported package `java.xml.parsers`.


tomakehurst:
Marvellous, thanks!

Issue #1075, opened 2019-01-31
[Feature request] New template helpers(e.g. hostname) supported for using in container env


yuanpli:
## Feature requests

Can wiremock add 'hostname' helper to get the hostname that wiremock is running? To make the same mapping configuration can be used in multiple different instances, 'hostname' helper can be used in response template to render different responses.

tomakehurst:
I'm curious what the use case would be for this.

Usually it's `localhost`, but if you've changed the bind address via the config, then surely you already know what hostname to use?

yuanpli:
Sorry I didn't describe the clear use case. What my purpose is to run wiremock in docker env. When multiple containers are setup, it is better to have some keyword to distinguish them.

tomakehurst:
How would this keyword manifest? Do you mean you'd like log messages to be distinguished by it?

yuanpli:
Not like log messages.  It would be better to support new helpers(e.g. hostname, systemenv ...) that can be used in response template.
Take an example to explain my points.
Response Templating configured in docker image
```
{
    "request": {
        "urlPath": "/hostname",
        "method": "GET"
    },
    "response": {
        "body": "{{hostname}}",
        "transformers": ["response-template"]
    }
}
```
Container1 started with this image by
```docker run --hostname wiremock1 -d wiremock```
Container2 started with the same image by
```docker run --hostname wiremock2 -d wiremock```

When client requests http get /hostname on Container1 , it can get the response 'wiremock1', and likewise, when requests on Container2, it can get the response 'wiremock2'.

yuanpli:
@tomakehurst could you kindly share your comments? Thanks!

PR #1085, opened 2019-02-20
Access to system or environment variables in Handlebars templates

philippe-granet:
See #1000

This PR enables system or environment variables to be used in generating the response or in proxy URLs.

Usage in Handlebars templates:
```
{{config.[java.home]}} when you want the value to come from any config source
{{config.env.JAVA_HOME}} when you want to target just the environment
{{config.sys.[java.home}} when you want to target just the system props
```
`config` lookup in the system variables first, followed by environment variables.  
Any . or - in the key will be converted to a _ in `config.env`, as well as uppercased.  
For example, key=wire-mock.rocks will be converted to WIRE_MOCK_ROCKS.  
No such translation will be performed for `config.sys`.

PR #1106, opened 2019-04-08
Added hostname and systemValue helpers. See #1075

yuanpli:
Response Templating stub configured in docker image
```
{
    "request": {
        "urlPath": "/hostname",
        "method": "GET"
    },
    "response": {
        "body": "{{hostname}}",
        "transformers": ["response-template"]
    }
}
```
Container1 started with this image by
```docker run --hostname wiremock1 -d wiremock```
Container2 started with the same image by
```docker run --hostname wiremock2 -d wiremock```

When client requests http get /hostname on Container1 , it can get the response ***wiremock1***, and likewise, when requests on Container2, it can get the response ***wiremock2***.


yuanpli:
Hi @tomakehurst Could you kindly pls review and approve the merge request? Hope it can be merged in next release. Thanks!

yuanpli:
Hello, I am very sad that there is not any response from reviewers for a long time. Could anyone work on this merge request?

yuanpli:
@tomakehurst 
Please review

tomakehurst:
@yuanpli please stop nagging me. I'll review it when I have time to do so and have reviewed all the other stuff that arrived before it.

tomakehurst:
This generally looks good, but I think there are a few things needed before I can merge it:

* Allowing unfiltered access to environment variables via templates is a security risk, since in many setups, environment variables are used for secrets. I suggest adding support for including only permitted variable names.
* Some acceptance tests. At a minimum: one showing resolution of hostname, a system property and an environment variable, and one showing that a forbidden env var cannot be accessed.

tomakehurst:
I'd forgotten that #1085 is also trying to do more or less the same thing. Might be worth you joining forces?

tomakehurst:
Interesting, I hadn't considered using the Java security manager for this.

A couple of issues I see with this:
1) I've no idea how or whether it'll work on Android. Do you have any idea whether Android implements the security manager in the same way as the JRE?
2) There's no easy way to specify which environment variables are permitted when running standalone from the command line.

Perhaps it'd be worth considering avoiding the security manager approach and just supporting a configuration option which is a list of regular expressions for names of permitted env vars?

yuanpli:
> Interesting, I hadn't considered using the Java security manager for this.
> 
> A couple of issues I see with this:
> 
> 1. I've no idea how or whether it'll work on Android. Do you have any idea whether Android implements the security manager in the same way as the JRE?
> 2. There's no easy way to specify which environment variables are permitted when running standalone from the command line.
> 
> Perhaps it'd be worth considering avoiding the security manager approach and just supporting a configuration option which is a list of regular expressions for names of permitted env vars?

Hi, I didn't consider the Android environment.  Do you mean the configuration option should be regarded as one of the wiremock startup parameters?

yuanpli:
BTW, if the configuration option and Response Templating stub both can be configured by user, it seems permission verification doesn't make any sense.

tomakehurst:
WireMock's admin API doesn't have security by default, so without a means for the sysadmin to control which env vars are visible, an unauthenticated user could create a stub via the API that reveals sensitive variables via templating.

yuanpli:
@tomakehurst Done the update. Pls have a review. Thanks!

PR #1107, opened 2019-04-09
WIP: Record stub information from JEE annotations

tomasbjerre:
This is a work in progress. Would be nice to have some opinions on this idea before I spend too much time on it.

The idea is to use the JEE annotations on resources. Have some of the stubbing options configured by giving the `stubFor` an actual method call.

Something like this:

```java
stubFor(invocation(TestResouce.class, (r) -> r.getStuff())
 .willReturn(aResponse().withStatus(202)));
```

I'd like to use lambda here... so **ok to drop Java 7 support now?** =)

**If Java 7 is still needed** then perhaps a pattern like this can be used:
```java
import static se.bjurr.wiremock.builder.*;

...

final MyDTO dto1 = new MyDTO("some value 1");
final MyDTO dto2 = new MyDTO("some value 2");
final List<MyDTO> listOfDtos = Arrays.asList(dto1, dto2);

doReturn(Status.ACCEPTED,listOfDtos) //
  .afterADelayOf(10, TimeUnit.SECONDS) //
  .afterInvocationOf(TestResouce.class) //
  .getListOfStuff();
```

But I cannot really fit that into the current `stubFor` pattern used here.

# TODO
 - headers
 - query parameters
 - path
 - match post content, json equal serialized input in POST/PUT methods.
 - ...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -28,8 +28,8 @@ allprojects {
     group = 'com.github.tomakehurst'
     version = '2.22.0'
 
-    sourceCompatibility = 1.7
-    targetCompatibility = 1.7
+    sourceCompatibility = 1.8
+    targetCompatibility = 1.8
 
     repositories {
         mavenCentral()
@@ -87,6 +87,7 @@ allprojects {
         compile("com.github.jknack:handlebars-helpers:$versions.handlebars") {
             exclude group: 'org.mozilla', module: 'rhino'
         }
+        compile 'javax:javaee-api:8.0'
 
         testCompile "org.hamcrest:hamcrest-all:1.3"
         testCompile("org.jmock:jmock:2.5.1") {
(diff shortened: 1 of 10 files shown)


tomakehurst:
I wouldn't want to couple WireMock to JEE, so I don't think this belongs in the core.

However, I can't see any reason you couldn't implement this in its own project. It shouldn't require any special access to WireMock's internals.

tomasbjerre:
Ok, totally understandable!

tomasbjerre:
@tomakehurst I started fiddling with this in its own repo now. I am planning to name it **Wiremock JAX-RS** and pubish it to Maven Central as: `se.bjurr.wiremock:wiremock-jaxrs:1.0`

Would you be ok with that naming?

Issue #1108, opened 2019-04-10
Allow the WireMock standalone to specify notifier. 


pvanassen:
Allow the WireMock standalone to specify notifier. 

Currently WireMock standalone server always uses 'ConsoleNotifier' as its notifier. A nice improvement would be to make this configurable. This would allow request mismatch logging to slf4j. 

Our specific use-case is, we run WireMock standalone in Mesos and we don't have easy access to those logs but we can forward slf4j to an ELK stack making the logs available for all our devs.

MEZk:
Hello!
I faced the same problem trying to configure WireMock to log in a specific format. For example: if I want to use SL4J to deliver logs to Splunk, or GrayLog. Currently WireMock logs requests journal as a multiline pretty json. These logs cannot be parsed easily by Splunk. It would be nice to have the ability to set custom notifier to transform logs.

pvanassen:
Ready for review

pvanassen:
PR needed merging again after 2 months of nothing. Not interested in only resolving merge conflicts without result.

PR #1109, opened 2019-04-10
Allow the WireMock standalone to specify notifier. 

pvanassen:
A suggested solution for: 
https://github.com/tomakehurst/wiremock/issues/1108


src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -134,6 +135,7 @@ public CommandLineOptions(String... args) {
         optionParser.accepts(ASYNCHRONOUS_RESPONSE_ENABLED, "Enable asynchronous response").withRequiredArg().defaultsTo("false");
         optionParser.accepts(ASYNCHRONOUS_RESPONSE_THREADS, "Number of asynchronous response threads").withRequiredArg().defaultsTo("10");
         optionParser.accepts(USE_CHUNKED_ENCODING, "Whether to use Transfer-Encoding: chunked in responses. Can be set to always, never or body_file.").withRequiredArg().defaultsTo("always");
+        optionParser.accepts(NOTIFIER, "Use a specific notifier. Can be either ConsoleNotifier, Slf4jNotifier or a full identifier").withRequiredArg().defaultsTo("ConsoleNotifier");
```
MEZk: > Slf4jNotifier or a full identifier

Fully qualified name of the class which implements Notifier interface. This will be clear, I gues

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -387,7 +389,23 @@ public MappingsSaver mappingsSaver() {
 
     @Override
     public Notifier notifier() {
-        return new ConsoleNotifier(verboseLoggingEnabled());
+        String notifier = (String) optionSet.valueOf(NOTIFIER);
+        if ("ConsoleNotifier".equals(notifier)) {
```
MEZk: I think that you should use ConsoleNotifier.class.getSimpleName() to avoid changes in future.

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -387,7 +389,23 @@ public MappingsSaver mappingsSaver() {
...
+            return new ConsoleNotifier(verboseLoggingEnabled());
+        }
+        if ("Slf4jNotifier".equals(notifier)) {
```
MEZk: same here. Use Slf4jNotifier.class.getSimpleName()

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -387,7 +389,23 @@ public MappingsSaver mappingsSaver() {
...
+            return new Slf4jNotifier(verboseLoggingEnabled());
+        }
+        else {
+            try {
+                ClassLoader loader = Thread.currentThread().getContextClassLoader();
```
MEZk: it will be better to use well-known and tested approach. See ClassUtils from Apache Commons. https://commons.apache.org/proper/commons-lang/apidocs/org/apache/commons/lang3/ClassUtils.html . The dependency is already in the project.
```
compile 'org.apache.commons:commons-lang3:3.7'
```

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -387,7 +389,23 @@ public MappingsSaver mappingsSaver() {
...
+                Class<?> cls = loader.loadClass(notifier);
+                return (Notifier)cls.newInstance();
+            } catch (Exception e) {
+                return throwUnchecked(e, null);
```
MEZk: In my opinion, It it better to add additional information in the exception. For instance: "Cannot load notifier by its name: full_name_of_the_notifier_class"

MEZk:
@pvanassen 
Thank you for your time and effort. I have the same problem configuring WireMock. Please, see my review comments.

pvanassen:
I will look at your review and resolve them shortly.

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
pvanassen: This part I took from httpServerFactory. I will refactor mine so it can serve as an example for the factory.

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -390,19 +392,19 @@ public MappingsSaver mappingsSaver() {
...
         }
         else {
             try {
-                ClassLoader loader = Thread.currentThread().getContextClassLoader();
-                Class<?> cls = loader.loadClass(notifier);
-                return (Notifier)cls.newInstance();
+                @SuppressWarnings("unchecked")
+                Class<Notifier> notifierClass = (Class<Notifier>)ClassUtils.getClass(notifier);
+                return ConstructorUtils.invokeConstructor(notifierClass);
             } catch (Exception e) {
```
pvanassen: I don't really like the catch all exceptions approach. Do you want to have it more specific?

MEZk:
@tomakehurst 
Hi. Could you review the changes, please? If you need any additional information on the problem described in the issue, let us know.

P.S. Till this fix I have to build the custom version of wiremock from @pvanassen's branch to override default notifier in standalone mode (

MEZk:
@pvanassen 
Please, rebase onto the latest master and resolve conflicts.

pvanassen:
@MEZk Any news if or when this PR will go to master?

MEZk:
@tomakehurst 
Please review

MEZk:
@tomakehurst 
ping

tomakehurst:
@MEZk I'm well aware that this PR exists. You might have noticed it's not the only one I haven't had time to review. Nagging me about it just antagonises me, so please stop.

tomakehurst:
I'm wondering how useful this is without a way to specify an slf4j implementation and config file. Presumably if you're picking slf4j, you want control over the output target, level and log format.

pvanassen:
For now, one could wrap all that in an extension. Not ideal but a step.

tomakehurst:
I'm not sure I see the point in that case. If you run the standalone JAR with slf4j activated at present, then you'll get no logging because there's no implementation in the JAR.

Since you can bake your own fat JAR with `Slf4jNotifier` and whatever impl + config you want, this seems like the sensible way forward.

What I might consider as a design option is including a logging impl in the standalone JAR which is used by default and where you can provide e.g. a `logback.xml` file as a CLI parameter.

pvanassen:
Does no longer compile and I have no interest in fixing it after 1+ month of no communications.

pvanassen:
I am no longer interested in putting effort into this.

PR #1111, opened 2019-04-11
Add support for unix timestamp in seconds

### LINKED ISSUE #1110: Epoch format from `now` helper does not give a UNIX Timestamp
Using `{{now format='epoch'}}` in a templated response gives a timestamp with too much precision. A Unix timestamp should be represented in _seconds_ since 1/1/1970 etc, the representation returned needs to be divided by 1000 it seems.

Right now using `{{now format='epoch'}}` gives `1554989345111` which is 25th August 51245 at 8:45pm... in reality, to be a truly portable timestamp, it the result should be divided by 1000 (and rounded I guess), to give `1554989345` 11th April 2019 @ 1:29pm (the correct value).

Depending on opinions, this could either be a bug (Unix timestamp is output with too much precision), or a feature request (please add support for portable Unix timestamps) :grin: thanks!

asgrim:
Fixes #1110 by adding support for `{{now format='unix'}}`

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RenderableDate.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RenderableDate.java
@@ -22,6 +22,7 @@ import java.util.Date;
 import java.util.TimeZone;
 
 public class RenderableDate {
+    private static final long DIVIDE_MILLISECONDS_TO_SECONDS = 1000L;
 
     private final Date date;
     private final String format;
@@ -41,7 +42,7 @@ public class RenderableDate {
             }
 
             if (format.equals("unix")) {
-                return String.valueOf(date.getTime() / 1000L);
+                return String.valueOf(date.getTime() / DIVIDE_MILLISECONDS_TO_SECONDS);
             }
 
             return formatCustom();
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RenderableDate.java
```diff
@@ -36,9 +36,15 @@ public RenderableDate(Date date, String format, String timezone) {
...
         if (format != null) {
-            return format.equals("epoch") ?
-                String.valueOf(date.getTime()) :
-                formatCustom();
+            if (format.equals("epoch")) {
+                return String.valueOf(date.getTime());
+            }
+
+            if (format.equals("unix")) {
+                return String.valueOf(date.getTime() / 1000L);
```
MEZk: 1000 is a magic number. Please extract to constant or use Duration to transform the time

asgrim: Added `private static final long DIVIDE_MILLISECONDS_TO_SECONDS = 1000L;`

tomakehurst:
Thanks!

PR #1116, opened 2019-04-19
Apply disableBodyFileTemplating only to body content

jtreanor:
The `disableBodyFileTemplating` property provides a useful way to disable templating for body files on a per stub basis. However, it can cause problems with a stub such as this one:

```
{
    "request": {
        "method": "GET",
        "urlPattern": "/asset/.*"
    },
    "response": {
        "status": 200,
        "bodyFileName": "{{request.path.1}}",
        "headers": {
            "Content-Type": "image/jpeg"
        },
        "transformerParameters": {
            "disableBodyFileTemplating": true
        }
    }
}
```

In this case the body file is an image, but the file name is derived through templating. If I don't pass `"disableBodyFileTemplating": true`, WireMock crashes as it tries to apply the handlebar templating to a binary `.jpg` file. However, if I do pass `"disableBodyFileTemplating": true` it also breaks because the file name is not expanded (WireMock tries to find the file at `__files/{{request.path.1}}`).

This PR makes a small change to the templating logic so that even if body file templating is disabled, the file path is still expanded.


tomakehurst:
The use case you've described was actually the intent of this feature, but clearly I managed to balls it up, so thanks for contributing this!

jtreanor:
Thanks a lot for the quick review @tomakehurst!

tomakehurst:
No problem, nothing like a bit of embarrassment as motivation!

Issue #1117, opened 2019-04-20
json body as string in response when adding a mapping with a jsonBody by using the rest api 


JeroenKops:
## Bug reports

- [X] latest version from Github (wiremock-jre8-standalone-2.23.2.jar)
- [X] Starting Wiremock as standalone jar with 
`java -jar wiremock-jre8-standalone-2.23.2.jar`
- [X] Wrote a Postman test and added Junit tests

## Feature requests

When adding a mapping by using the rest API and if the mapping contains a jsonBody, I expect that the response also contains a jsonBody in stead of a body as a string with escaped characters. Also when I try to get the mapping (by it's ID) it also gives the body as a string with escaped characters

I've  added a patch to this issue which solves this problem (with Junit tests). Now when adding a jsonBody in the mapping it also returns a jsonBody in the response.
[patch.zip](https://github.com/tomakehurst/wiremock/files/3100173/patch.zip)

I used Postman to test if it works:
[Postman.zip](https://github.com/tomakehurst/wiremock/files/3100176/Postman.zip)

I also created a pull request for this issue: #1118

tomakehurst:
Closing as associated PR merged

PR #1118, opened 2019-04-20
#1117 json body as string in response when adding a mapping with a js…

JeroenKops:
When adding a mapping by using the rest API and if the mapping contains a jsonBody, I expect that the response also contains a jsonBody in stead of a body as a string with escaped characters. Also when I try to get the mapping (by it's ID) it also gives the body as a string with escaped characters


tomakehurst:
Thanks!

PR #1119, opened 2019-04-22
Updating docs with Wiremock JAX-RS

tomasbjerre:


tomasbjerre:
@tomakehurst Will you also publish this in the gh-pages branch?

PR #1125, opened 2019-05-10
Local response templating in Wiremock War

derari:
Allow to enable local response templating in the `web.xml` using
```
  <context-param>
    <param-name>LocalResponseTemplating</param-name>
    <param-value>true</param-value>
  </context-param>
```


tomakehurst:
Please could you add a test case for this?

mSenussi:
Hi @derari,

thanks for such a useful feature, i've had a workaround to enable registering extensions. 
i'd suggest to enable passing `Map<String, ? extends Extension> extensions` to WarConfiguration. then it'd be possible to add as many extensions. not only ResponseTemplateTransformer.
then override WarConfiguration.extensionsOfType() to return map of extensions grouped by extensionType.

PR #1134, opened 2019-06-05
Add scenario index on record: differentiates per RequestPattern group

adz:
Without this playback gets confused when the same url is
hit (such as is typical with a SOAP API) but groups differently
based on payload or headers.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
@@ -45,14 +45,16 @@ public class ScenarioProcessor {
             }
         });
 
+        int scenarioIndex = 0;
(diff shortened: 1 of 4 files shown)


adz:
@tomakehurst is there something I should do differently in this PR? Not sure if it is in tune with your thoughts of how it should work. It seems repeatsAsScenarios doesn't... as it groups requests correctly (based on method and body as well as url) but assigns scenario names based on url only. This leads to duplicated scenario names and playback can be out of expected order.

I noticed a few people on mailing list with similar problems to myself which triggered the PR. 

Willing to help out fixing this some other way as we're using a fork with this patch at work which I'd rather not be doing.

tomakehurst:
Sorry, I haven't had a chance to review this yet. It's not gone unnoticed.

tomakehurst:
Thanks for contributing (and for your patience)

Issue #1154, opened 2019-07-10
[Feature request] Add ability to modify Request sent to WireMock before proxying it to real server


AndriiFrolov:
Please add possibility to modify request before sending it to proxied real server

Use case:
I have an application that configured to WireMock server. WireMock server configured to proxy requests to real server. 

 wireMock.register(stubFor(any(anyUrl())
                    .willReturn(aResponse().proxiedFrom("http://myrealserver.com")
UI application -> WireMock -> Real server. This scheme is used to stub only several requests.

Would be useful to have possibility to modify requests that sends real application. For instance, it can be used to avoid login procedure via UI. With such feature I would be able to get token via Rest api request and add this token to all requests

bademux:
Duplicate? https://github.com/tomakehurst/wiremock/issues/1147

Issue #1147, opened 2019-07-03
Feature Request: Alter original HTTP Request


bademux:
Hello ,
I have weird problem with SOAP body, server requires to have its url in request body (SOAP AdressingFeature on), I have to rewrite request body by replacing wiremock's url (localhost in my case) to targetServerUrl. More info [here](https://groups.google.com/forum/#!topic/wiremock-user/rU8ZG8MyBJI).

With a bit of pain I found a way to hook the request [here the example](https://gist.github.com/bademux/b2eb7e930207a521aa210d1ebcff21e0), but it will be nice to have a wiremock Extension and ability to alter current request (RequestBuilder#like() method for Request url)

Thanks!

tomakehurst:
Already built :-) Just waiting for a release, which I'll try and do this weekend.

Here's an example: https://github.com/tomakehurst/wiremock/blob/master/src/test/java/com/github/tomakehurst/wiremock/RequestFilterAcceptanceTest.java#L194

bademux:
It's fantastic! Thanks!
Waiting for the release.

bademux:
Works like a charm in [2.24.0](https://github.com/tomakehurst/wiremock/releases)

Issue #745, opened 2017-09-02
Proxy stub mappings - Need proxyUrlPrefixToRemove for proxy context url mapping


kabiliravi:
Hi,

Problem Description:
In Proxy stub mappings, I am trying to map a prefixed request url to a proxy url, but I need that prefix I added to the request call be removed when it is getting dispatched to the proxy host, if it is matched for example to a transformation parameter called **proxyUrlPrefixToRemove**

Acceptance Criteria: 
GIVEN The following code will proxy all GET requests made to http://<host>:<port>/other/service/.* to http://otherservice.com/approot
And proxyUrlPrefixToRemove transformation parameter is set to "/other/service"
WHEN running WireMock locally a request to http://localhost:8080/other/service/doc/123 
THEN it would be forwarded to http://otherservice.com/approot/doc/123

My proposed solution:
Since wiremock code is written in a way that there is no place to override classes to add a feature, I had to customize the wiremock code to do it.

Here is the change I made, 

In com.github.tomakehurst.wiremock.http.ProxyResponseRenderer class:

Current Code:
```
   public static HttpUriRequest getHttpRequestFor(ResponseDefinition response) {
		final RequestMethod method = response.getOriginalRequest().getMethod();
		final String url = response.getProxyUrl();
		return HttpClientFactory.getHttpRequestFor(method, url);
	}
```

Changed Code:
```
protected synchronized HttpUriRequest getHttpRequestFor(ResponseDefinition response) {
		String proxyUrlPrefixToRemove = null;
		if(response.getTransformerParameters() != null) {
			proxyUrlPrefixToRemove = (String) response.getTransformerParameters().get("proxyUrlPrefixToRemove");
		}

		RequestMethod method = response.getOriginalRequest().getMethod();
		String requestContextUrl = response.getOriginalRequest().getUrl();
		String proxyBaseUrl = response.getProxyBaseUrl();
		StringBuilder sb = new StringBuilder();
		sb.append(proxyBaseUrl);
		if (StringUtils.isNoneBlank(proxyUrlPrefixToRemove) && requestContextUrl.startsWith(proxyUrlPrefixToRemove)) {
			requestContextUrl = requestContextUrl.substring(proxyUrlPrefixToRemove.length());
		}
		sb.append(requestContextUrl);
		return HttpClientFactory.getHttpRequestFor(method, sb.toString());
	}
```
As you see, I had to make **getHttpRequestFor** a non-static method so that I can have access to transformation parameter.

[...]

PR #1187, opened 2019-09-13
Use modified requests

jeremydyoung:
I'm submitting this pull request because if you use the extension StubRequestFilter to wrap a request which modifies the URL that will be proxied, the original url gets used for the proxy request.

This change allows for a particular stub to retain the existing functionality of proxying the unmodified request url, or to set the useModifiedRequest flag to true, where the wrapped request will be sent to the ResponseDefinition when rendering the response.  

An example config might contain:

```
{
  "priority": 1,
  "request": {
    "headers" : {
      "X-Destination-Service" : {
        "equalTo" : "foo"
      }
    }
  },
  "response": {
    "proxyBaseUrl" : "https://foo.service/",
    "useModifiedRequest": "true"
  }
}
```

or in a test:

```
        stubFor(
                get(urlMatching('/.*'))
                        .withHeader('X-Destination-Service', containing('foo'))
                        .atPriority(1)
                        .willReturn(aResponse().withStatus(200)
                                .proxiedFrom('https://foo.service')
                                .withModifiedRequests()
                        )
        )
```
The filter I used with this feature is: 

```
public class ServiceNameFilter extends StubRequestFilter {
    @Override
    public RequestFilterAction filter(Request request) {
        String service = "foo";
        Request wrappedRequest = RequestWrapper.create()
                .transformAbsoluteUrl(url -> url.replaceFirst("/" + service, ""))
                .addHeader("X-Destination-Service", service)
                .wrap(request);
        return RequestFilterAction.continueWith(wrappedRequest);
    }

    @Override
    public String getName() {
        return "service-name-filter";
    }
}
```

Where my ServiceNameFilter stripped foo off the front of the url such as https://wiremock.server/foo/status, to make it so I wanted https://foo.service/status as the proxied request.  The filter was also responsible for adding the header.  In our case the reason for this was to disambiguate between two services that had the same url patterns for their requests that we wanted to mock out or proxy on.

Without my changes to ResponseDefinition, the proxied request was https://foo.service/foo/status, even though the filter wrapped the request and made it return /status as the url.

I wanted to make tests that asserted this behavior, but it turns out it wasn't clear where those tests should reside.  I can come back and add tests if someone provides a little guidance on what would fit well with the project.  

The default behavior is intended to be unchanged with this feature.  If there are unintended consequences for this change, then there may need to be adjustments.

I wasn't able to find the code from https://github.com/tomakehurst/wiremock/issues/1147 or seem to make it work with 2.24.1

This feature request seems relevant:
https://github.com/tomakehurst/wiremock/issues/1154

I think this may be a solution to this Feature Request:
https://github.com/tomakehurst/wiremock/issues/745

It might be possible to make the filter configurable, but since the filters seem to only be global, that seems difficult.  I dug into the code a little looking for something like this, but I am not extremely familiar with the code base, so I couldn't find a clear strategy.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
@@ -44,6 +44,7 @@ public class ResponseDefinitionBuilder {
     protected DelayDistribution delayDistribution;
     protected ChunkedDribbleDelay chunkedDribbleDelay;
     protected String proxyBaseUrl;
+    protected boolean useModifiedRequest;
     protected Fault fault;
     protected List<String> responseTransformerNames;
     protected Map<String, Object> transformerParameters = newHashMap();
@@ -63,6 +64,7 @@ public class ResponseDefinitionBuilder {
         builder.delayDistribution = responseDefinition.getDelayDistribution();
         builder.chunkedDribbleDelay = responseDefinition.getChunkedDribbleDelay();
         builder.proxyBaseUrl = responseDefinition.getProxyBaseUrl();
+        builder.useModifiedRequest = responseDefinition.isUseModifiedRequest();
         builder.fault = responseDefinition.getFault();
         builder.responseTransformerNames = responseDefinition.getTransformers();
         builder.transformerParameters = responseDefinition.getTransformerParameters() != null ? Parameters.from(responseDefinition.getTransformerParameters()) : Parameters.empty();
@@ -127,6 +129,11 @@ public class ResponseDefinitionBuilder {
 
     public ResponseDefinitionBuilder withChunkedDribbleDelay(int numberOfChunks, int totalDuration) {
         this.chunkedDribbleDelay = new ChunkedDribbleDelay(numberOfChunks, totalDuration);
+        return this;
+    }
+
+    public ResponseDefinitionBuilder withModifiedRequests() {
+        this.useModifiedRequest = true;
         return this;
     }
 
@@ -196,6 +203,7 @@ public class ResponseDefinitionBuilder {
             this.delayDistribution = from.delayDistribution;
             this.chunkedDribbleDelay = from.chunkedDribbleDelay;
             this.proxyBaseUrl = from.proxyBaseUrl;
+            this.useModifiedRequest = from.useModifiedRequest;
             this.responseTransformerNames = from.responseTransformerNames;
             this.transformerParameters = from.transformerParameters;
         }
@@ -245,6 +253,7 @@ public class ResponseDefinitionBuilder {
                     delayDistribution,
                     chunkedDribbleDelay,
                     proxyBaseUrl,
+                    useModifiedRequest,
                     fault,
                     responseTransformerNames,
                     transformerParameters,
@@ -264,6 +273,7 @@ public class ResponseDefinitionBuilder {
                     delayDistribution,
                     chunkedDribbleDelay,
                     proxyBaseUrl,
+                    useModifiedRequest,
                     fault,
                     responseTransformerNames,
                     transformerParameters,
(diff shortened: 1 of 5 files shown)


kovaku:
I am facing with same issue, although I have implemented the aforementioned `StubRequestFilter` and added an as extension, it is not able to amend the request that is proxied to 3rd party service. It would be great to have this PR merged 👍

DanielScholtz:
+1

tomakehurst:
Sorry for the wait. I decided maintaining an option to keep the original behaviour was too complicated/confusing, so I've switched the behaviour to always use the wrapped request here: f9e64d249889fb346b57bb6ea693b9da7d32de10

nituspl:
Hi, I cannot use this feature. I have wiremock 2.27.2 and when trying to run standalone wiremock I get this error:

```
Exception in thread "main" com.github.tomakehurst.wiremock.standalone.MappingFileException: Error loading file /home/nituspl/wiremock/./mappings/proxy.json:
Unrecognized field "useModifiedRequest" (class com.github.tomakehurst.wiremock.http.ResponseDefinition), not marked as ignorable
	at com.github.tomakehurst.wiremock.standalone.JsonFileMappingsSource.loadMappingsInto(JsonFileMappingsSource.java:121)
	at com.github.tomakehurst.wiremock.core.WireMockApp.loadMappingsUsing(WireMockApp.java:204)
	at com.github.tomakehurst.wiremock.core.WireMockApp.loadDefaultMappings(WireMockApp.java:200)
	at com.github.tomakehurst.wiremock.core.WireMockApp.<init>(WireMockApp.java:103)
	at com.github.tomakehurst.wiremock.WireMockServer.<init>(WireMockServer.java:73)
	at com.github.tomakehurst.wiremock.standalone.WireMockServerRunner.run(WireMockServerRunner.java:65)
	at com.github.tomakehurst.wiremock.standalone.WireMockServerRunner.main(WireMockServerRunner.java:134)

```

andrake1982:
have an error(the same as above) in standalone 2.28 version on
POST .../__admin/mappings
`{
    "request": {
        "urlPattern": "/cds/.*",
        "method": "GET"
    },
    "response": {
        "proxyBaseUrl": "http:cds.cluster.local/cds.service",
        "transformers":["service-name-filter"],
        "useModifiedRequest": "true"
    }
}`

get error
`{
    "errors": [
        {
            "code": 10,
            "source": {
                "pointer": "/response/useModifiedRequest"
            },
            "title": "Error parsing JSON",
            "detail": "Unrecognized field \"useModifiedRequest\" (class com.github.tomakehurst.wiremock.http.ResponseDefinition), not marked as ignorable"
        }
    ]
}`

jeremydyoung:
> 
> 
> have an error(the same as above) in standalone 2.28 version on
> POST .../__admin/mappings
> `{ "request": { "urlPattern": "/cds/.*", "method": "GET" }, "response": { "proxyBaseUrl": "http:cds.cluster.local/cds.service", "transformers":["service-name-filter"], "useModifiedRequest": "true" } }`
> 
> get error
> `{ "errors": [ { "code": 10, "source": { "pointer": "/response/useModifiedRequest" }, "title": "Error parsing JSON", "detail": "Unrecognized field \"useModifiedRequest\" (class com.github.tomakehurst.wiremock.http.ResponseDefinition), not marked as ignorable" } ] }`
> 
> Looks like this is just not implemented

useModifiedRequest was not implemented.  The concept of using the modified request was made default without a flag.

andrake1982:
@jeremydyoung with your solution(with builtin ServiceNameFilter) we would got the solution for [#745](https://github.com/tomakehurst/wiremock/issues/745) "Need proxyUrlPrefixToRemove for proxy" even for standalone Wiremock mode. But with applied solution we don't got builtin tool for mock separated endpoints for different services.
So users to achieve this at standalone wiremock should pack transformer into war, and then run it with extension. It can be not so trivial if wiremock deploing via Helm chart in kubernetes cluster. 

@tomakehurst is this feature at roadmap of wiremock future? really usefull feature. Many users need this.

tomakehurst:
@andrake1982 the feature you're referring to - do you mean #745?

andrake1982:
@tomakehurst 

> @andrake1982 the feature you're referring to - do you mean #745?
Yes

PR #1194, opened 2019-10-01
Fix: tabs to spaces

laflechejonathan:
Indentation is inconsistent across the code base, which makes viewing code in github difficult.

I'm inferring that spaces are the preferred way given:
```bash
❯ git grep --files-with-matches "       " | grep java | wc -l # tabs
      76
❯ git grep --files-with-matches "    " | grep java | wc -l # spaces
     578
```

The transformation I ran was:
```
git ls-files | grep "\.java" | xargs perl -pi -e "s/\t/    /"
```

But I'm happy to do the opposite transformation!


laflechejonathan:
nope - not worth it to ruin git blame + indentation of JSON blocks

PR #1200, opened 2019-10-07
Feature: Add --http-disabled flag

chomatdam:
**Context:**
By default, Wiremock has always an HTTP listener started on 8080 if the flag `port` is not set even if you are setting an HTTPS listener.

**Goal:**
Adding a flag `--http-disabled` allowing the Wiremock server to be reached only from HTTPS.

### DIFF
--- a/docs-v2/_docs/https.md
+++ b/docs-v2/_docs/https.md
@@ -43,6 +43,12 @@ The keystore type defaults to JKS, but this can be changed if you're using anoth
 .keystoreType("BKS")
 ```
 
+To allow only HTTPS requests, disable HTTP by adding:
+```java
+@Rule
+public WireMockRule wireMockRule = new WireMockRule(wireMockConfig().httpsPort(8443).httpDisabled(true));
+```
+
 ## Requiring client certificates
 
 To make WireMock require clients to authenticate via a certificate you
@@ -66,7 +72,6 @@ specify a trust store containing the certificate(s).
 > Version 9.4.15.v20190215 of Jetty (used in the jre8 WireMock build) requires client certificates to contain Subject Alternative Names.
 > See [this script](https://github.com/tomakehurst/wiremock/blob/master/scripts/create-client-cert.sh) for an example of how to build
 > a truststore containing a valid certificate (you'll probably want to edit the client-cert.conf file before running this).
-
 
 ## Common HTTPS issues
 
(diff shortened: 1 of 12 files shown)


tomakehurst:
Looking good!

Please could you add an acceptance test, then I'll merge.

chomatdam:
Cool, I added a few tests!

tomakehurst:
Could you add an acceptance test (that starts WireMock for real and tests via an HTTP client)?

This is probably the right place to put it (and see examples of this kind of test):
https://github.com/tomakehurst/wiremock/blob/master/src/test/java/com/github/tomakehurst/wiremock/HttpsAcceptanceTest.java

chomatdam:
Hi @tomakehurst,
Before your last comment, I had 3 tests:
- one checking the HTTP connector is null
- one checking the command `--http-disabled` is mapped to the object WiremockConfig
- one acceptance test checking HTTPS is still behaving as expected when HTTP is disabled in HttpsAcceptanceTest

On the last one, if I'm testing the HTTP port, I would have to expose the private method `httpPort` and this one would throw an exception saying the server can't start with the option `--http-disabled`. 
I could test this exception `IllegalStateException` but it would be more a unit test.

tomakehurst:
Urgh, sorry. Don't know how I glossed over that. Guess I shouldn't review PRs on a Monday morning before my 1st coffee! I'll take another look on the train home, and probably merge.

tomakehurst:
One more small thing - I'm not keen on changing the behaviour of `WireMock.port()` depending on the whether HTTP is enabled or not. Can you revert that so that it's always explicitly `WireMock.httpsPort()` that needs to be called?

tomakehurst:
Would also be good to have an assertion in the acceptance test confirming that there is nothing listening on the HTTP port.

chomatdam:
Hey @tomakehurst,

I reverted the part around `port()` returning HTTPS or HTTP port based on the options set. 

The HTTP connector being null, I'm asserting the `IllegalStateException` and message thrown when we are calling `port()`. 

If we recap, we have:

- one test validating the command option is well applied to the Wiremock configuration
- one test validating the HTTP connector is null when this option is passed
- one test validating HTTPS returns a stub and throws an exception when we are calling `port()`

chomatdam:
@tomakehurst 🙂

PR #1210, opened 2019-10-18
Added unauthorizedJson method to support json body

sachingorade:
In some cases we get JSON content in return for unauthorized status code which clients might want to parse, so added unauthorizedJson method which returns 401 status code with specified body in response with application/json content type.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -493,6 +493,10 @@ public class WireMock {
         return aResponse().withStatus(401);
     }
 
+    public static ResponseDefinitionBuilder unauthorizedJson(String body) {
+	    return aResponse().withStatus(401).withHeader(CONTENT_TYPE, "application/json").withBody(body);
+    }
+
     public static ResponseDefinitionBuilder forbidden() {
         return aResponse().withStatus(403);
     }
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
```diff
@@ -493,6 +493,10 @@ public static ResponseDefinitionBuilder unauthorized() {
         return aResponse().withStatus(401);
     }
 
+    public static ResponseDefinitionBuilder unauthorizedJson(String body) {
+	    return aResponse().withStatus(401).withHeader(CONTENT_TYPE, "application/json").withBody(body);
```
conbon: This could be useful for other HttpStatus' too. 

Would it be possible to extract the ```withHeader().withBody()``` out into another method to allow for fluent composition with other error codes? For example unauthorized().withJsonBody() / badRequest().withJsonBody()

sachingorade: Yes, sounds better. Updated code with the suggestion.

sachingorade:
@conbon Please check the updated diff, is there anything that should be updated?

LEQADA:
@tomakehurst any chance to progress here?

conbon:
A bit late @sachingorade but looks good. @tomakehurst, can this be merged?

PR #1215, opened 2019-10-28
Exclude transitive jackson dependencies

xalvarez:
`zjsonpatch` adds older `jackson` dependencies than the ones that are already being added by wiremock itself.


tomakehurst:
Not sure what purpose this serves, given that WireMock declares its own (up to date) versions of these dependencies?

tomakehurst:
Hmmm....this isn't the first time Maven Enforcer has been mentioned. It seems a bit over-weening to me, given that Maven will also upgrade the dependency, but I guess if it's a common problem it's worth appealing to it.

PR #1218, opened 2019-10-31
WIP: added support to enable slf4j notifier for standalone mode

sumitanvekar:
This is only a WIP branch. We need to discuss how we can integrate with agreed upon logging framework.

@tomakehurst continuing from the email thread in wiremock-users (Subject: Slf4jNotifier cannot be enabled when running Wiremock in Standalone mode), I was wondering if we really need to tie it down to a particular framework or would it be preferable to keep wiremock independent of logging framework.

If we can make SLF4J work just by adding any desired framework and config into class-path at runtime (I need to verify if something like that would work), would that be better?

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
@@ -92,9 +92,8 @@ public class CommandLineOptions implements Options {
     private static final String USE_CHUNKED_ENCODING = "use-chunked-encoding";
     private static final String MAX_TEMPLATE_CACHE_ENTRIES = "max-template-cache-entries";
     private static final String PERMITTED_SYSTEM_KEYS = "permitted-system-keys";
-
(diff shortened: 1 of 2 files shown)


tomakehurst:
This a good start, but it'll need some things added in order to do anything useful:

* An actual logging implementation.
* A shading rule for the logging implementation + whatever fixes this requires

Also, I'd suggest forming the CLI parameter like this (assuming we used logback, which I'm not necessarily advocating):
`--use-logback-config=/path/to/wiremock-logback.xml`

Also, I'd suggest just setting the verbose flag to true in this instance, since you can filter messages by severity using the logging config.

sumitanvekar:
Sorry, I was not able to contribute to this pr sooner.

I was thinking about this solution and was wondering if there is any harm in using logback and utilising its `-Dlogback.configurationFile` parameter to provide configuration path instead of creating a new config parameter.
One of the problems I see with a new parameter is all the logs up until the code where logback config from the given path is read will be in default format which is not very nice.

PR #1234, opened 2019-12-10
EqualToJsonPattern simplified

lukas-krecan:


tomakehurst:
Ah nice, this is a big improvement! This looks like it might be good enough to fully integrate and merge.

Thanks!

tomakehurst:
One question - it seems like the conversions to/from `JsonNode` are using a conversion function that does a tree walk, rather than simply wrapping/unwrapping, which (although better than parsing) will presumably still have a performance penalty.

Am I interpreting that correctly, or is there a shortcut somewhere that prevents a full conversion from happening when the type is already correct?

lukas-krecan:
Conversion from JsonNode can be done directly (currently is not, but it's a one-liner). The opposite direction will indeed require some tree walking. 

I prefer to pay a small performance penalty than to maintain universal JSON node abstraction.

lukas-krecan:
https://github.com/lukas-krecan/JsonUnit/commit/caa3d3056a2f80254e12be0c7579fa5945181e1d

tomakehurst:
Implemented in 7ead912283b9c89d010eaf523e877ff5a5a5a728. Thanks again @lukas-krecan!

PR #1273, opened 2020-02-25
Make response template caching works

philippe-granet:
Actually, cache doesn't work because of incorrect HttpHeaders.hashCode() implementation.
This pull request fix this bug 😃


tomakehurst:
Thanks for this Philippe. Good spot!

PR #1287, opened 2020-03-18
Add request details to VerificationException

prathik:
The intent behind the exception message is that request must be printed but it was not being done due to the lack of formatter in `String.format`, this PR fixes that issue and modifies the tests to assert the same.


prathik:
@tomakehurst would appreciate your feedback on this.

PR #1297, opened 2020-04-13
wrong RequestSpec parsing fix by using reference equality instead of class equality

vananos:
It seems that it's safer to use reference equality when trying to get RequestSpecification by AdminTask instance because using class equality produces the wrong behaviour when a task is decorated by decorator (or lambda), and it also safer than instance.equals(), because we are looking for spec that returned this concrete instance.


vananos:
@tomakehurst take a look at fix please)

PR #1309, opened 2020-05-15
Doc update to "Response Templating" adding `parameters.` prefix

poobury:
It took me a while to work out why my parameters weren't showing up in my template, and it was because I didn't realise I needed a `parameters.` prefix to select them from the model (I ended up debugging it to work out why it wasn't working). Thought I'd update the documentation to hopefully prevent others making the same mistake with this great tool!


tomakehurst:
Thanks!

PR #1310, opened 2020-05-15
Update stubbing example

ashishkujoy:
Explicitly mentioned that query parameter will not work in ```WireMock.urlPathEqualTo()``` and in not suggested to use ```. withQueryParam```.


tomakehurst:
This isn't correct - the example you've modified uses `urlEqualTo` not `urlPathEqualTo` so it **does** include the query.

ashishkujoy:
My bad. Closing it.

Issue #1313, opened 2020-05-23
Response Templating: request.headers.<key> is case sensitive


tkrieger:
## Problem Description
The Response Templating request model expressions `request.headers.<key>` and `request.headers.[<key>]` are not case insensitive for header keys. E.g.:
- `{{request.headers.X-CorrelationID}}` could not replaced if only a header with the key `x-correlationid` is present.

Request matching is working perfectly case insensitive instead.

## Impact
We have to use WireMock behind a Node.JS proxy in our automated test setup. The Node.JS proxy converts all HTTP headers keys to lowercase. 
(Which is completely okay due RFC 2616 https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2). 
Which means that or response templating based on header values is not working in this setup. 
We can not simply use lowercase in the mapping definition because this would break the local development setup.
Unfortunately this bug stops us from fully automate all our tests right now.

As work around I tried to add a request filter which corrects the case for required headers. However this is not working because here it seems that a case insensitive check is done and the replacement of the key itself is not executed for any reason.
## WireMock Setup
WireMock Version: `2.26.3`
WireMock Standalone Command: `java -jar wiremock-standalone-jre8.jar  --verbose --global-response-templating`

### Sample Mapping and Request
Sample mapping using X-CorrelationID and tries to access request.headers.<key> with different cases.
```
{
  "id": "733ee0bd-7cbb-4483-8179-213dabc6d4c5",
  "name": "case-sensitive-headers",
  "request": {
    "url": "/test",
    "method": "GET",
    "headers": {
      "x-correlationid": {
        "equalTo": "test-id"
      }
    }
  },
  "response": {
    "status": 200,
    "body": "{\"X-CorrelationID\": \"{{request.headers.X-CorrelationID}}\", \"x-correlationid\": \"{{request.headers.x-correlationid}}\"}"
  },
  "uuid": "733ee0bd-7cbb-4483-8179-213dabc6d4c5",
  "persistent": true,
  "insertionIndex": 15
}
```
Request used to test:
```
curl --location --request GET 'http://localhost:8080/test' --header 'X-CorrelationID: test-id'
```
Received response:
```
{"X-CorrelationID": "test-id", "x-correlationid": ""}
```
Expected behavior: both attributes should contain `test-id`.
### JUnit Test to reproduce the issue
I added following JUnit test to reproduce the issue into the _ResponseTemplateTransformerTest.java_:
```
@Test
[...]

PR #1314, opened 2020-05-23
Response Templating: Case-Insensitive Headers

tkrieger:
This changes adaptedHeaders in `RequestTemplateModel.from(..)` to TreeMap with CASE_INSENSITIVE_ORDER to ignore case for headers in response templating.

E.g. allows that `{{request.headers.X-CorrelationID}}` also matches `x-correlationid`.

Resolves issue https://github.com/tomakehurst/wiremock/issues/1313


tomakehurst:
Great stuff, thanks!

Issue #401, opened 2016-05-07
Support forward (browser) proxying to an HTTPS target


tomakehurst:
Currently, it's not possible to forward proxy onto an HTTPS site with WireMock, since HTTPS proxying is done via the HTTP CONNECT method, which essentially makes the proxy act as a TCP proxy. The reason this is necessary is that the stream can be simply forwarded to the destination IP without being decrypted, meaning it can be done securely (the proxy isn't effectively a man-in-the-middle attack).

I believe it might be possible to work around this, with the caveat that the caller would still have to either trust a self-signed cert, or ignore the cert entirely. If WireMock, on receipt of the CONNECT call, forwarded the stream to its own HTTPS port rather than to the actual destination, it could then handle it as normal, albeit serving a different SSL cert to the one on the destination server.

selangley:
You might look for some implementation hints in these two projects that claim to do this - both Java and Apache-licensed:

https://github.com/jamesdbloom/mockserver
https://github.com/ganskef/LittleProxy-parent

tomakehurst:
Thanks for the tip. I had a little dig around already in mockserver, but couldn't find anything obviously useful.

I hadn't encountered the other one before so I'll definitely take a look.

selangley:
Yesterday, I found a couple of more projects:

Zed Attack Proxy (Java and Apache-licensed)
https://www.owasp.org/index.php/OWASP_Zed_Attack_Proxy_Project
https://github.com/zaproxy/zaproxy

mitmproxy (Python and Apache-licensed)
http://docs.mitmproxy.org/en/stable/howmitmproxy.html
https://github.com/mitmproxy/mitmproxy

tomakehurst:
Again, thanks for the pointers.

LittleProxy and MockServer are both Netty based, and it looks like this does the heavy lifting. Unfortunately, I think taking advantage of that would mean a wholesale shift of WireMock onto Netty, which would be a big job.

james-boswell:
@tomakehurst I am trying to use standalone wiremock (version 2.1.7) to record requests/responses to a service using HTTPS. I am currently getting a "SunCertPathBuilderException: unable to find valid certification path..." exception.

Given what you have said on this issue, are you saying it is currently not possible to proxy a HTTPS request?

tomakehurst:
[...]

PR #1316, opened 2020-05-25
Stub & record forward proxy https traffic

Mahoney:
This allows WireMock to act as a forward (browser) proxy for HTTPS as well
as HTTP origins whilst stubbing & recording.

Step towards implementing #401.


Mahoney:
Opened https://github.com/tomakehurst/wiremock/pull/1317 instead.

PR #1317, opened 2020-05-25
Stub & record forward proxy https traffic

Mahoney:
This allows WireMock to act as a forward (browser) proxy for HTTPS as well
as HTTP origins whilst stubbing & recording.

Step towards implementing #401.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,3 +54,19 @@ To build both JARs (thin and standalone):
 
 The built JAR will be placed under ``java8/build/libs``.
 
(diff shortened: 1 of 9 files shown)


SamD:
Just found this searching through some WireMock HTTPS related issues

Is the goal here to support in conjunction with something like https://github.com/ganskef/LittleProxy-mitm ?

I was attempting to setup WireMock today so that it would `proxy-via` the littleproxy-mitm instance as well as recording (--record-mappings) which didn't work

PR #1321, opened 2020-05-29
Validate certs by default on forward proxy

Mahoney:
Provide an option `--trust-all` to allow turning this behaviour off.


tomakehurst:
Looks good. Happy for this to be merged now.

PR #1336, opened 2020-06-19
Better error message on communication failure when proxying

Mahoney:


src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
```diff
@@ -105,11 +105,21 @@ public Response render(ServeEvent serveEvent) {
...
+        } catch (SSLException e) {
+            return proxyResponseError("SSL", httpRequest, e);
+        } catch (IOException e) {
+            return proxyResponseError("Network", httpRequest, e);
+        }
 	}
 
+
+    private Response proxyResponseError(String type, HttpUriRequest request, Exception e) {
+        return response()
```
tomakehurst, line 117: How about we log the error here too, so that the full stack trace is available if you've got access

Mahoney: Pushed a change; I've logged it as info, as it's normal behaviour to make a request to an endpoint that doesn't exist / is down / has no valid cert

PR #1367, opened 2020-08-20
Persist request

yashaswini9440:


tomakehurst:
We usually ask that for non-trivial PRs, you first raise an issue so that the problem and solution design can be properly discussed.

I suggest you pause coding and open an issue explaining clearly what problem you're trying to solve here, and the design you're aiming for. Aside from anything else, request journal persistence is something I've been considering for a while, so I'm not likely to merge any changes that conflict with the design goals I have in mind.

tomakehurst:
Closing this off as it looks like you've opened a newer PR on the same topic.

PR #1376, opened 2020-09-11
Add support for jsonpath defaults

ncjones:
This PR includes two related changes to the JsonPath template helper:

1. output nothing (empty string) when no path matched instead of error message
2. allow overriding the default value when no path matched

The default value is static only. There may be a more generic solution to apply defaults for example by adding a "default" helper.

Gradle build didn't work for me on Oracle JDK 14 so I just ran a subset of the unit tests in Intellij.

Let me know if you want docs updated.

### DIFF
--- a/docs-v2/_docs/response-templating.md
+++ b/docs-v2/_docs/response-templating.md
@@ -374,6 +374,14 @@ And for the same JSON the following will render `{ "inner": "Stuff" }`:
 ```
 {% endraw %}
 
+Default value can be specified if the path evaluates to null or undefined:
+
+{% raw %}
+```
+{{jsonPath request.body '$.size' default='M'}}
+```
+{% endraw %}
+
 
 ## Date and time helpers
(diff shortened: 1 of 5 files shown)


ncjones:
@tomakehurst Do you have any feedback on this? 

The original discussion was at https://groups.google.com/g/wiremock-user/c/vg3PWVzmc1Q/m/gvOMzN3cBAAJ.

tomakehurst:
Sorry for the delay getting to this.

I'm going to merge but make a small change, which is to keep the original behaviour when there's no default specified and the JSONPath doesn't match anything. This is to avoid a breaking change. It'll still be possible to pass an empty string as the default if required.

tomakehurst:
Actually, scratch that. The current default behaviour is pretty misleading. I'll keep it your way.

ncjones:
Marvelous! Thank you :)

PR #1380, opened 2020-09-16
persist Dynamo mappings

yashaswini9440:


tomakehurst:
The title of this implies your intention is to bake support for Dynamo into WireMock itself. Is that the case?

yashaswini9440:
> The title of this implies your intention is to bake support for Dynamo into WireMock itself. Is that the case?

Yes @tomakehurst  added an extension to persist into any database

yashaswini9440:
Addressed all the comments

yashaswini9440:
> Add space before options.

DrMattFaulkner:
> > The title of this implies your intention is to bake support for Dynamo into WireMock itself. Is that the case?
> 
> Yes @tomakehurst added an extension to persist into any database

I suspect the title of the PR is just wrong. This looks like it's adding an extension for the journal and stub services.

tomakehurst:
Closing this as I'm going to support this requirement in due course, but need to take a step back and think carefully about the design.

PR #1397, opened 2020-10-15
MetricsPublisher Interface as an Extension for publishing metrics

nitinPaypay:
**Problem** :  Existing Wiremock doesn't support metrics publishing and if Wiremock is being used as a service for third-party mocks then it is very difficult to track no of requests served by Wiremock and latency corresponding to the requests.

**Solution** : Added an interface MetricsPublisher as an extension which can be implemented by Service who is using Wiremock as a service. 
This interface is having 3 methods for publishing metrics corresponding to each request.
1. Summary 
2. Counter
3. Gauge

These are the default meters which are used by most of the instrumentation libraries. 
Consumer can implement MetricsPublisher interface for publishing metrics based on their Instrumentation library.

As of now MetricsPublisher can be used for publishing latency information and no of requests served by Wiremock.

By Default we have added 3 tags - 
1. uri - (Normalized form)
2. method - (GET/POST)
3. status code - (200/404)

Which will help to identify served requests details based on these tags.

Implementer of this interface can be passed as an extension while using Standalone like - 
        `  --extensions jp.ne.paypay.ee.transformer.NewRelicMetricsPublisher `


tomakehurst:
It's actually quite straightforward to publish metrics from WireMock via the `PostServeAction` extension point. In fact we use this internally in MockLab to push metrics via Micrometer.

I can see that you're adding some value here by normalising URLs etc. so I suggest you look at publishing this as a separate library that can be used via this extension point.

nitinPaypay:
Thank you so much Tom for reviewing this PR and providing information about PostServeAction. 
So looks like now we can use PostServeAction for publishing metrics so no need of this PR.
I will mark it closed.

Issue #1343, opened 2020-06-30
Support multiple matches in "regexExtract"


bashmish:
**tl;dr**: I need the `g` flag or so-called Global match (at least in the JavaScript terminology)

Currently I'm limited in the "regexExtract" if I want to extract multiple matches.

`{{ regexExtract '/path/1/path/2' '\d+' 'parts' }}{{ parts.0 }} {{ parts.1 }}`

Expected result: `1 2`
Actual result: `1`

I know that `parts` is currently reserved only for match groups. And multiple matches as such are simple not implemented in https://github.com/tomakehurst/wiremock/blob/2.26.3/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RegexExtractHelper.java#L33

We can also think about an alternative helper for that, e.g. 

`{{ regexExtractAll '/path/1/path/2' '(\s+)\\(\d+)' 'parts' }}{{ parts.0.0 }} {{ parts.0.1 }} {{ parts.1.0 }} {{ parts.1.1 }}`
Expected result: `path 1 path 2`

PS: sorry if I made some mistakes in the regexpes above, not my strongest point, however I hope that the idea is clear

kwoding:
I just created a pull request to support this.

bashmish:
thanks for a quick response!

I looked into your unit tests here https://github.com/tomakehurst/wiremock/pull/1400/files

it's a bit uncommon to me to see comma separated values in the path rather than query params, but technically speaking it's kinda the same, so I feel like this is a work in the right direction

my use case is a bit more than just that and it's my bad I didn't explain it in detail in this issue

I also need uniformity in the way how single value or a comma separated list of values are parsed, let me illustrate it with the code based on your unit tests

```java
// ...
mockRequest()
  .url("/items?ids=1"),
aResponse()
  .withBody("{\"test\": \"{{regexExtract request.query.ids '(\d+)' 'parts'}}{{#each parts}}{{this}} {{/each}}\"}").build(),
// ...
assertThat(responseDefinition.getBody(), is("{\"test\": \"1 \"}"));
```

```java
// ...
mockRequest()
  .url("/items?ids=1,2,3"),
aResponse()
  .withBody("{\"test\": \"{{regexExtract request.query.ids '(\d+)' 'parts'}}{{#each parts}}{{this}} {{/each}}\"}").build(),
// ...
assertThat(responseDefinition.getBody(), is("{\"test\": \"1 2 3 \"}"));
```

[...]

PR #1400, opened 2020-10-27
Add #1343 support multiple matches in regexextract

kwoding:


kwoding:
Added unit tests to cover this. Let me know whether that is also sufficient for your use cases.

kwoding:
@tomakehurst could you also check this PR? Thanks in advance.

PR #1474, opened 2021-04-29
Resolve sonatype issues

tacotown:
commons-io upgraded
netty-all upgraded
log4j upgraded to log4j2

Also, would not build due to node-sass 4.12.0 no longer found at location.  Upgraded to 4.14.1 to resolve the build issue.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -76,7 +76,11 @@ dependencies {
     }
     compile "net.minidev:json-smart:2.4.6" // Pinning this above the transitive version from json-path to get CVE fix
     compile "org.ow2.asm:asm:9.1"
-    compile "org.slf4j:slf4j-api:1.7.30"
+    compile "org.slf4j:slf4j-api:1.7.30", {
+        exclude group: 'log4j',module:'log4j'
+    }
+    compile "org.apache.logging.log4j:log4j-api:2.13.3"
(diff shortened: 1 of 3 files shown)


tomakehurst:
This is quite a major dependency reshuffle so please could you share the specific issues Sonatype is raising and how these changes resolve them?

build.gradle
```diff
@@ -77,6 +77,9 @@ dependencies {
     compile "net.minidev:json-smart:2.4.6" // Pinning this above the transitive version from json-path to get CVE fix
     compile "org.ow2.asm:asm:9.1"
     compile "org.slf4j:slf4j-api:1.7.30"
+    compile "org.apache.logging.log4j:log4j-slf4j-impl:2.13.2"
```
tomakehurst, line 80: log4j shouldn't be made a production dependency. At present it's only used for testing. Is there a reason you're adding it to the compile scope in addition to test?

tomakehurst (changes requested):

tomakehurst:
Please can you also drop the changes to docs-v2/package.json and package-lock.json?

These are just used to generate the doc site, not in WireMock itself. They're not tested in CI and often just upgrading dependencies results in the site looking broken, so a full visual regression test is needed.

tomakehurst (approved):

PR #1490, opened 2021-05-13
Fix broken link to script on "Proxying" page

MasonM:
http://wiremock.org/docs/proxying/ has a broken link to a script to create a self-signed root certificate. This link was added in 396e4c870775f37e4ff2efe80c8e6af91ca7ef8e, and I'm pretty sure it was intended to link to https://github.com/tomakehurst/wiremock/blob/master/scripts/create-ca-keystore.sh


tomakehurst:
Good spot, thanks!

PR #1493, opened 2021-05-18
Fix broken link

irof:
The standalone-jar link in getting-started was broken, so I made it the same as download-and-installation and running-standalone.


tomakehurst:
Thanks!

Issue #1420, opened 2021-01-20
Wiremock polutes system variables, which causes incompatibilities in multi-module-builds


henrykuijpers:
- [ 2.27.2 ] Which version of WireMock you're using
- [ Unit tests (JUnit) ] How you're starting and configuring WireMock, including configuration or CLI the command line
- [ ] A failing test case that demonstrates the problem

Issue is in: com.github.tomakehurst.wiremock.common.xml.Xml, where it sets System-wide JVM properties that define which TransformerFactory will be used.

WireMock should get rid of poluting system variables, by removing these statements, or at least make this configurable.
Otherwise, if they are intended to have a certain functionality in the command line tool, at least do not set them in the unit testing context.

Right now, this is failing our build in another module, with: 
```
Exception in thread "main" javax.xml.transform.TransformerFactoryConfigurationError: Provider org.apache.xalan.processor.TransformerFactoryImpl not found
    at javax.xml.transform.FactoryFinder.newInstance(FactoryFinder.java:181)
    at javax.xml.transform.FactoryFinder.find(FactoryFinder.java:216)
    at javax.xml.transform.TransformerFactory.newInstance(TransformerFactory.java:106)
    at sun.util.xml.PlatformXmlPropertiesProvider.emitDocument(PlatformXmlPropertiesProvider.java:174)
    at sun.util.xml.PlatformXmlPropertiesProvider.store(PlatformXmlPropertiesProvider.java:168)
    at java.util.Properties$XmlSupport.save(Properties.java:1208)
    at java.util.Properties.storeToXML(Properties.java:953)
    at java.util.Properties.storeToXML(Properties.java:908)
    at org.apache.jackrabbit.filevault.maven.packaging.GenerateMetadataMojo.execute(GenerateMetadataMojo.java:497)
    at org.apache.maven.plugin.DefaultBuildPluginManager.executeMojo(DefaultBuildPluginManager.java:137)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:210)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:156)
    at org.apache.maven.lifecycle.internal.MojoExecutor.execute(MojoExecutor.java:148)
    at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject(LifecycleModuleBuilder.java:117)
[...]

PR #1496, opened 2021-05-18
#1420 Timeout & disable optimize xml factories loading

henrykuijpers:
Add configuration options for the global default timeout, as well as the disabling of optimize factories loading

Leave optimize factories loading enabled by default, for backwards compatibility
Remove no longer used org.mortbay.log.class system property setting

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/jetty9/JettyHttpServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/jetty9/JettyHttpServer.java
@@ -31,6 +31,7 @@ import org.apache.commons.lang3.ArrayUtils;
 import org.eclipse.jetty.http.MimeTypes;
 import org.eclipse.jetty.io.NetworkTrafficListener;
 import org.eclipse.jetty.server.*;
+import org.eclipse.jetty.server.handler.AbstractHandler;
 import org.eclipse.jetty.server.handler.HandlerCollection;
 import org.eclipse.jetty.server.handler.HandlerWrapper;
 import org.eclipse.jetty.servlet.DefaultServlet;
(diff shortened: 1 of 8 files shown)


henrykuijpers:
I see that now a `org.eclipse.jetty.http.HttpGenerator.STRICT` system property has appeared, @tomakehurst  :(

tomakehurst:
Yes, unfortunately I couldn't find any other way to fix that particular bug. Jetty only permits that value to be set from a system property. I find it pretty irritating that library authors still do that in this day and age 😞

tomakehurst:
I suppose we could move that under a config toggle too.

henrykuijpers:
I just added an additional commit, to also enable/disable strict HTTP headers in Jetty, @tomakehurst . 

Any chance you can do a review? :)

henrykuijpers:
@tomakehurst What do you think about the changes?

tomakehurst:
This is really two PRs in one. Please can you split out the timeout change into its own?

src/main/java/com/github/tomakehurst/wiremock/jetty9/JettyHttpServer.java
```diff
@@ -102,7 +107,13 @@ public JettyHttpServer(
 
         applyAdditionalServerConfiguration(jettyServer, options);
 
-        jettyServer.setHandler(createHandler(options, adminRequestHandler, stubRequestHandler));
+        final HandlerCollection handlers = createHandler(options, adminRequestHandler, stubRequestHandler);
+        jettyServer.setHandler(new HandlerCollection(ArrayUtils.insert(0, handlers.getHandlers(), new AbstractHandler() {
```
tomakehurst, line 111: How about adding this inside the HandlerCollection created in `createHandler`? Might be a bit neater than accepting a collection then re-packing it into another.

tomakehurst (changes requested):
Generally looks good. I've made a few minor suggestions.

PR #1533, opened 2021-06-15
Update handlebars to 4.7.7 to resolve CVE-2021-23369

t-patt:


PRAgarawal (approved):
I don't think my approval will help because Github is saying you need a maintainer to approve, buuut... :shipit:

tomakehurst:
There's no version 4.7.7 of the Java Handlebars port, so this won't work.

But since this CVE applies to the JS version, I don't think that's a problem.

Issue #1545, opened 2021-06-30
NPE thrown when null JSON passed to parseJson helper


tomakehurst:
We should either return an empty object or report the error.

PR #1549, opened 2021-07-04
Fix 1547 parse json helper

### LINKED ISSUE #1547: parseJson helper won't parse JSON that has an array at the top level

TuxedoFish:
fixes #1547

- Adds a ParseJsonHelperTest.java test file.
- Adds 5 test cases including 2 where the array is at the top level.
- Checks whether the json is enclosed by array tags.
   - If it is then a List <Object> is returned instead of a Map <String, Object>.
   - This way it does not interfere with other test cases.

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/ParseJsonHelperTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/ParseJsonHelperTest.java
@@ -24,6 +24,8 @@ import org.junit.Test;
 
 import java.io.IOException;
 import java.util.Map;
+import java.util.Arrays;
+import java.util.ArrayList;
 
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.*;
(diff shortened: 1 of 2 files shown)


src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/ParseJsonHelperTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @Test
+    public void parsesAJsonObjectContainingArray() throws Exception {
+        String inputJson = "{\"arr\": [\"one\", \"two\", \"three\"]}";
+        Object output = render(inputJson, new Object[]{}, TagType.VAR);
+
+        assertThat(output, instanceOf(Map.class));
+        Map<String, Object> result = (Map<String, Object>) output;
+        assertThat(result, aMapWithSize(1));
+        assertThat(result, hasKey("arr"));
+        assertThat(result.get("arr"), instanceOf(ArrayList.class));
```
tomakehurst, line 61: I suggest checking for a `List` here (and in the subsequent tests) rather than `ArrayList` so that this won't break if for some reason the underlying implementation changes.

tomakehurst (changes requested):
Looks good, just one minor point about the class we assert for in the tests.

src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/ParseJsonHelperTest.java
TuxedoFish: Thanks 😃, that makes sense, I have updated that, let me know if there is anything else @tomakehurst, I'll have a look at #1545 next otherwise!

PR #1556, opened 2021-07-08
#745 Need proxyUrlPrefixToRemove for proxy context url mapping

SatsG:
Fix for #745. The original request mentions using a transformation parameter, but this PR follows a similar pattern as the additionalProxyRequestHeaders option in the response definition.


SatsG:
@tomakehurst Raised: https://github.com/wiremock/wiremock/pull/1696

PR #1563, opened 2021-07-13
Enforce consistent code style with Spotless

### LINKED ISSUE #1559: Enforce consistent code format
I've noticed that there are a few places in the codebase (i.e. [`FileSource`](https://github.com/tomakehurst/wiremock/blob/d62c457e71e6f66a5120139d9cb93d35d2622cc5/src/main/java/com/github/tomakehurst/wiremock/common/FileSource.java#L24-L35)) which use inconsistent code style.

Would you be happy with an enforced code style, perhaps via the [Spotless Gradle plugin](https://github.com/diffplug/spotless/tree/HEAD/plugin-gradle), which allows folks to easily apply code style changes through `./gradlew spotlessApply`, instead of needing any other local setup?

If so, are there any preferences you would have? I.e. are you happy with Google Java code style, or would another style be preferable?

jamietanna:
As a first step towards having consistent, enforced code style across
the project, to simplify diffs, and to reduce the effort by
contributors, we can use Spotless.

By default, we'll enable Java, Gradle and JSON configurations, and have
the option of adding other source types in the future.

We need to exclude the `sample.json` file as it's not valid JSON,
otherwise Spotless fails to run.

This also documents how to run Spotless locally, and documents how to
have it running automagically on compiles via https://www.jvt.me/posts/2020/05/15/gradle-spotless/

Finally, we can manage our license headers using Spotless, too.

Closes #1559.


gradle/spotless.java.license
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright (C) 2011 Thomas Akehurst
```
jamietanna, line 2: We can also make this so it updates? So either:

```suggestion
 * Copyright (C) $YEAR Thomas Akehurst
```

Or:


```suggestion
 * Copyright (C) 2011-$YEAR Thomas Akehurst
```

PR #1565, opened 2021-07-15
Enable support for this codebase's to write JUnit5 tests

jamietanna:
JUnit5 is arguably a better development experience, and we should allow
contributors to start writing unit tests using it, instead of JUnit4,
where possible.

However, until we've migrated the codebase between versions, we can use
the vintage-engine to run both JUnit4 and JUnit5 tests at the same time!

This also includes the platform-launcher, which will allow us to launch
the tests from an IDE.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -94,7 +94,10 @@ dependencies {
     }
     compile "commons-io:commons-io:2.10.0"
 
-    testCompile "junit:junit:4.13"
+    testImplementation platform("org.junit:junit-bom:5.7.2")
+    testImplementation "org.junit.jupiter:junit-jupiter"
+    testRuntimeOnly "org.junit.vintage:junit-vintage-engine"
+    testRuntimeOnly "org.junit.platform:junit-platform-launcher"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -94,6 +94,10 @@ dependencies {
     }
     compile "commons-io:commons-io:2.10.0"
 
+    testImplementation platform("org.junit:junit-bom:5.7.2")
+    testImplementation "org.junit.jupiter:junit-jupiter"
+    testRuntimeOnly "org.junit.vintage:junit-vintage-engine"
+    testRuntimeOnly "org.junit.platform:junit-platform-launcher"
     testCompile "junit:junit:4.13"
```
tomakehurst, line 101: What's the effect of keeping this here?

jamietanna: Hmm, so I thought that without this, JUnit4 would fail to run, but as it's brought in as a transitive dependency of `junit-vintage-engine`, it's not required - I'll amend :+1:

jamietanna:
Closing as it looks like this has been solved already through 79907e474b718ad9266d51708cf5460a98b7976e

Issue #1476, opened 2021-05-01
Document running Wiremock without HTTP Server


jamietanna:
I've recently [discovered and blogged about](https://www.jvt.me/posts/2021/04/29/wiremock-serverless/) how to run Wiremock without an HTTP server.

This is something that is on the autocomplete when searching google for `wiremock without`, so I feel this is something that's quite likely to be wanted by users.

At the very least, I've found it to be a requirement to put Wiremock into a serverless function, and found this solution to work really nicely.

I'd be happy to raise a PR to add this to the documentation, as well as adding a regression test to the codebase to ensure that this functionality is still available in the future.

tomakehurst:
I think this would be a valuable addition. I'm slightly hesitant about documenting the way you've blogged about though, as it would make `WireMockApp` part of the public interface, whereas at the moment it's purely an implementation class. This will become more of a problem when WireMock starts properly supporting Java modules.

I'm wondering if it might be worth adding a new class specifically that provides a request/response interface based around the classes/interfaces that are already part of the public API, then documenting how to use this. That would certainly be easier to support in the future without having to make difficult choices around backwards compatibility.

jamietanna:
That makes sense - do you have any initial thoughts about how this would best be served? And would this be something you'd prefer to see contributed, or would it be something you'd look to get to when you had time, especially to reduce the number of folks relying on internal APIs, as you say :sweat_smile:

jamietanna:
Hey Tom, was wondering if you had any thoughts about this - I'd like to propose something, and maybe even get a PR raised, but would like to see what you're thinking about it before I dig into it.

I'm seeing 10s of hits/week on my blog post for this, so although I've made it clear on my post it's going to be unstable to rely on this, it may still be happening.

tomakehurst:
I think the way to do this would be to add alternative implementations of `HttpServer` and `HttpServerFactory` that simply expose admin and stub request handlers via the `RequestHandler` interface. This avoids exposing internal classes as we discussed.

The slightly tricky part of this is that `RequestHandler` is designed to support async operation - it takes an `HttpResponder`, which is essentially a callback as a parameter. Since you're likely to favour synchronous calling when running without a web server it might make sense to wrap futures around it e.g.:

```java
public class DirectCallHttpServer implements HttpServer {

    private final RequestHandler adminHandler;
    private final RequestHandler stubHandler;

   // Called from the corresponding HttpServerFactory impl
[...]

PR #1566, opened 2021-07-15
Refactor creation of `RequestHandler` classes

jamietanna:
As part of work towards #1476, we need to refactor this logic out of the
core `WireMockApp`, so we can then re-use it as part of the Direct HTTP
client setup.

This further simplifies the setup of the `WireMockApp`, too.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
@@ -129,45 +129,13 @@ public class WireMockApp implements StubServer, Admin {
     }
 
     public AdminRequestHandler buildAdminRequestHandler() {
-        AdminRoutes adminRoutes = AdminRoutes.defaultsPlus(
-            options.extensionsOfType(AdminApiExtension.class).values(),
-            options.getNotMatchedRenderer()
-        );
-        return new AdminRequestHandler(
-            adminRoutes,
-            this,
-            new BasicResponseRenderer(),
-            options.getAdminAuthenticator(),
-            options.getHttpsRequiredForAdminApi(),
-            getAdminRequestFilters()
-        );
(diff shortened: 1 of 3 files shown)


tomakehurst:
I don't think this is necessary as you can just pass them in via the factory:

```java
public class DirectCallHttpServerFactory implements HttpServerFactory {

    @Override
    public HttpServer buildHttpServer(Options options, AdminRequestHandler adminRequestHandler, StubRequestHandler stubRequestHandler) {
        return new DirectCallHttpServer(adminRequestHandler, stubRequestHandler);
    }
}
```

jamietanna:
So the thought here was that we need to create them somewhere, though, so we can then pass them into the factory - I was planning to have this code called in the constructor of the `DirectCallHttpServer`, similar to what we're doing with the `WiremockServer`.

This was primarily to reduce the duplication of writing the same code across the classes, and then there's only one place to add test coverage for in the future.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
```diff
@@ -129,45 +129,13 @@ public WireMockApp(
...
-        );
-        return new AdminRequestHandler(
-            adminRoutes,
-            this,
-            new BasicResponseRenderer(),
-            options.getAdminAuthenticator(),
-            options.getHttpsRequiredForAdminApi(),
-            getAdminRequestFilters()
-        );
+        return new AdminRequestHandlerFactory(). // TODO: field? Static?
```
jamietanna, line 132: I guess this would go into `Options`, following how we've done it in `WireMockServer`

tomakehurst:
You won't need to explicitly create them as this already happens in `WireMockServer` (which is agnostic of the specific HTTP server you're using).

jamietanna:
Gotcha, I'll close this then and read up on the difference between `WireMockApp` and `WireMockServer`

PR #1572, opened 2021-07-30
Allow running Wiremock without HTTP Server

### LINKED ISSUE #1476: Document running Wiremock without HTTP Server

jamietanna:
As noted in #1476, a common search online is to run Wiremock without
needing an HTTP server, for instance if using a serverless deployment
model.

It is currently possible to do this using the `WireMockApp`, which
relies on internal APIs, which will break when we move to Java Modules.

To avoid this, we can implement the `DirectCallHttpServer` as a first
class citizen, which, when created, will allow consumers to retrieve the
mapped `Response` (whatever it may be) for Stub and Admin requests.

We can use the existing Jetty `stopTimeout` configuration to control the
configuration for this server's async timeouts, so consumers can
override it in cases they need delays on responses.

Because we're using the `HttpResponder`s as callbacks, we need to use
Mockito's `doAnswer` to invoke the real `HttpResponder` from our mock.

We also add the Mockito Jupiter bindings to more easily write JUnit
Jupiter tests using Mockito.

Finally, we want to add an example test to show consumers how they can
use it for themselves.

Closes #1476.

---

Depends on #1565, but otherwise ready for reviews


tomakehurst:
Hey, please could you rebase this against master?

jamietanna:
Thanks Tom, have done :+1:

PR #1592, opened 2021-09-12
Allow standalone runner to fetch mappings from classpath

### LINKED ISSUE #1502: Allow the Standalone runner to fetch mappings from the classpath
Is it possible to make it so the standalone runner picks up anything on the classpath?

For instance I've shaded the standalone app with a set of stubs, and a custom extension, and I'd like to have standalone not needing a second set of stub mappings on disk as well.

It looks like this isn't quite possible with `--root-dir` provided as an argument to the standalone runner

jamietanna:
To allow consumers to shade their standalone JAR with the stubs they
want to produce, for instance to allow there to be a single artefact
with all their stubs and code, we should allow this being specified to
be part of the classpath.

This introduces a new flag, `--root-dir-under-classpath`, to allow
configuring both the `__files` and `mappings` directories, as it's more
likely that they'll be configured together, especially as the underlying
`FileSource` is shared.

Closes #1502.

### DIFF
--- a/docs-v2/_docs/running-standalone.md
+++ b/docs-v2/_docs/running-standalone.md
@@ -298,6 +298,29 @@ matching the URL exists. For example if a file exists
 `/things/myfile.html` then hitting
 `http://<host>:<port>/things/myfile.html` will serve the file.
 
+## Packaging the stubs into a standalone JAR
+
+If you want to package your stubs into the standalone JAR, so you can distribute an executable JAR with all the stubs intact, you can do this using the `--load-resources-from-classpath` option.
+
+For example, let's say have the following directory structure:
+
+```
+src/main/resources
+src/main/resources/wiremock-stuff
+src/main/resources/wiremock-stuff/__files
+src/main/resources/wiremock-stuff/mappings
+```
+
+You could then run the packaged JAR as:
+
+```
+java -jar custom-wiremock.jar --load-resources-from-classpath 'wiremock-stuff'
+```
+
+Which will load your files and mappings from the packaged JAR.
+
+Note that it is not currently possible to load from the root of the classpath.
+
 ### Shutting Down
 
 To shutdown the server, either call `WireMock.shutdownServer()` or post
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
@@ -113,6 +113,7 @@ public class CommandLineOptions implements Options {
     private static final String HTTPS_CA_KEYSTORE = "ca-keystore";
     private static final String HTTPS_CA_KEYSTORE_PASSWORD = "ca-keystore-password";
     private static final String HTTPS_CA_KEYSTORE_TYPE = "ca-keystore-type";
+    private static final String LOAD_RESOURCES_FROM_CLASSPATH = "load-resources-from-classpath";
 
     private final OptionSet optionSet;
     private final FileSource fileSource;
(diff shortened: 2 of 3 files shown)


src/test/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptionsTest.java
```diff
@@ -598,6 +601,42 @@ public void returnsTheSameInstanceOfTemplatingExtensionForEveryInterfaceImplemen
...
+        assertThat(fileSource, instanceOf(ClasspathFileSource.class));
+    }
+
+    @Test
+    public void rootDirUnderClasspathSetsMappingsSourceToUseClasspath() {
+        CommandLineOptions options = new CommandLineOptions("--root-dir-under-classpath");
+
+        MappingsSaver mappingsSaver = options.mappingsSaver();
+
+        assertThat(mappingsSaver, instanceOf(JsonFileMappingsSource.class));
```
jamietanna, line 637: :question: I'm not sure this is good enough as either way it's returned - thoughts?

jamietanna (commented):
Hmm, this doesn't seem to have quite worked - I'll try and dig through it, and put this into Draft for now

jamietanna:
This works when running:

```
java -jar ./build/libs/wiremock-jre8-standalone-2.31.0.jar --root-dir-under-classpath --root-dir 'wiremock'
```

With stubs under `/wiremock/mappings/` on the classpath

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -113,6 +113,7 @@
     private static final String HTTPS_CA_KEYSTORE = "ca-keystore";
     private static final String HTTPS_CA_KEYSTORE_PASSWORD = "ca-keystore-password";
     private static final String HTTPS_CA_KEYSTORE_TYPE = "ca-keystore-type";
+    private static final String ROOT_DIR_UNDER_CLASSPATH = "root-dir-under-classpath";
```
tomakehurst, line 116: Maybe `load-mappings-from-classpath` would be a better config key for this?

src/test/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptionsTest.java
tomakehurst: Is there anything stopping you calling `CommandLineOptions.filesRoot()` and checking the runtime class of the result?

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
jamietanna: That's fair - I guess this will be both `__files` and `mappings` - is `load-resources-from-classpath` OK to cover both?

src/test/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptionsTest.java
jamietanna: If you see above, we do have that tested - would you be happy that this is "good enough", as we're pretty confident that the mappings are sourced by the underlying `filesRoot`?

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
tomakehurst: Yes, I guess "resources" is better since it's both. Let's go with that.

tomakehurst:
One further request - please could you add a section to the standalone doc page explaining how to make use of this i.e. bundling mappings/files into a fat JAR?

jamietanna:
Certainly, will get on that shortly!

jamietanna:
Let me know what you think - when you're happy I'll fix up the commit history :+1:

docs-v2/_docs/running-standalone.md
```diff
@@ -298,6 +298,29 @@ matching the URL exists. For example if a file exists
...
+src/main/resources
+src/main/resources/wiremock-stuff
+src/main/resources/wiremock-stuff/__files
+src/main/resources/wiremock-stuff/mappings
+```
+
+You could then run the packaged JAR as:
+
+```
+java -jar custom-wiremock.jar --load-resources-from-classpath --root-dir 'wiremock-stuff'
```
tomakehurst, line 317: I feel like these two parameters are (or ought to be) mutually exclusive - either I'm loading from the classpath or file system. What happens if I specify both?

jamietanna: Ah so root dir is used to inform the directory on the classpath at that point

tomakehurst: OK, I see.

As a way of avoiding some confusion, perhaps it'd be worth considering `load-resources-from-classpath` taking the root path as its parameter e.g. `--load-resources-from-classpath=wiremock-stuff`?

jamietanna: Good thinking - amended :+1:

jamietanna:
Hey Tom, any idea when we may be looking at getting a release out with this? :raised_hands:

Issue #1612, opened 2021-10-01
Improve ObjectMapper configuration API


Adven27:
## Feature requests

Default ObjectMapper configured with the following options disabled:
```
DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS
JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN
```

Which results in applying scientific number notation or rounding to big numbers, e.g. `1264534222.24` in a stub file became `1.26453422218E9` in response. That may lead to false-positive or false-negative tests.

I've failed to find a straightforward and bullet-proof way to manually tune ObjectMapper configuration because different instances are created on a per-thread basis and `Json.getObjectMapper().configure(...)` should be called in each thread what is not always possible.

Is something from the following possible:
1)  to add this options to the default config
2) to add a more straightforward API for ObjectMapper configuration, for example as part of the regular `WireMockServer(...)` or `WireMock(...)` configuration process

Thanks in advance! And Thank You for the awesome project!

tomakehurst:
Presumably this only happens if you're using `jsonBody` rather than specifying your response body as a string?

I feel like this should probably be the default setting as you should always get back what you put in. Can you see any potential breaking changes from this (I can't off the top of my head)?

Adven27:
Yep, I have an issue with `jsonBody` currently, not sure about other cases...

Regarding breaking changes I don't see any too.

tomakehurst:
Do you feel like creating a PR to add these to the defaults?

PR #1613, opened 2021-10-01
#1612 prevent applying scientific notation and rounding to big numbers

Adven27:
prevent applying scientific notation and rounding to big numbers by ObjectMapper


tomakehurst:
Thanks!

PR #1616, opened 2021-10-01
Draft: Migrate project tests to JUnit 5 from JUnit 4

timtebeek:
This MR converts the project tests from JUnit 4 to JUnit 5, except where we're explicitly testing JUnit 4 compatibility.

Conversion was started using https://docs.openrewrite.org/tutorials/migrate-from-junit-4-to-junit-5;
Expanded by hand mostly to convert WireMockRule usage to WireMockExtension.

Not ready yet, thought to gather feedback on whether this has a chance of getting merged before I continue.

Any feedback appreciated!

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/admin/model/GetScenariosResult.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/admin/model/GetScenariosResult.java
@@ -20,7 +20,6 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 import com.github.tomakehurst.wiremock.stubbing.Scenario;
 
 import java.util.List;
-import java.util.Map;
 
 public class GetScenariosResult {
 
--- a/src/main/java/com/github/tomakehurst/wiremock/client/BasicMappingBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/BasicMappingBuilder.java
@@ -32,7 +32,6 @@ import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
 import static com.google.common.base.MoreObjects.firstNonNull;
 import static com.google.common.base.Preconditions.checkArgument;
 import static com.google.common.collect.Lists.newArrayList;
-import static com.google.common.collect.Maps.newLinkedHashMap;
 
 class BasicMappingBuilder implements ScenarioMappingBuilder {
 
--- a/src/main/java/com/github/tomakehurst/wiremock/common/DateTimeOffset.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/DateTimeOffset.java
@@ -17,10 +17,7 @@ package com.github.tomakehurst.wiremock.common;
 
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
-import java.util.Calendar;
 import java.util.Date;
-import java.util.TimeZone;
-
 import static com.github.tomakehurst.wiremock.common.DateTimeUnit.SECONDS;
 
 public class DateTimeOffset {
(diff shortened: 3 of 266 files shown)


timtebeek:
There's two somewhat puzzling test failures left over:
```
HandlebarsXPathHelperTest. rendersNamespacedElement()
PlainTextDiffRendererTest. wrapsLargeXmlBodiesAppropriately
```
Both fail on whitespace when comparing expected & actual; not sure if that's platform related, or if we can just update the expectation or matching.

timtebeek:
Also notable that we appear to still use [JMock 2.5.1 from 2008](https://github.com/jmock-developers/jmock-library/releases/tag/2.5.1)(!), which needs an upgrade to at least [2.10 from 2019](https://github.com/jmock-developers/jmock-library) to support JUnit Jupiter.

For now I had left the JMock tests on JUnit4, and suggest those are upgraded separately from this MR.

timtebeek:
@tomakehurst / @alexsapps would you mind taking a look at this MR? I understand it's sizeable, but I figured the best way to expand on the JUnit Jupiter use is to dog-food it in the project itself.

I mostly went for [programmatic usage](https://github.com/wiremock/wiremock/blob/master/docs-v2/_docs/junit-jupiter.md#advanced-usage---programmatic) as it seemed the most direct replacement for the previous Rule usage.

alexsapps:
Hi @timtebeek I am not a maintainer of wiremock; I just made a small edit to the docs recently. Perhaps someone else can take a look.

tomakehurst:
@timtebeek can I suggest we pause for a moment on this one?

This PR is scarily big and is going to create merge hell for all the other open PRs.

Also, I'm really not sure whether the value of porting all the tests to Jupiter is justified. Perhaps just switching to the Vintage engine and ditching Junit 4 as a dependency would make sense? That would (I think) also avoid us having to worry about jmock.

timtebeek:
Yeah sure; understand your concerns about the size of the merge request, even if most of it is imports.

I think there's value in adopting the latest version of JUnit, even while JUnit 4 maintenance & bug fix releases are still coming out. It does well to showcase both that this project supports JUnit 5, as well as keeps the code base future ready. But I won't push hard on this issue; was just hoping to be helpful and conversion was a nice test to play around with what can and can not be achieved automatically already.

As for JMock, I've pulled that [into a separate merge request](https://github.com/wiremock/wiremock/pull/1617), as the 2008 dependency there is potentially problematic. Even when run through the vintage engine that still ties you to a quite old and infrequently updated mock engine.

Would there be value in doing a partial/gradual upgrade through more easily digested per-package upgrades for instance? Or do you want to hold of on a JUnit5 upgrade completely for now?

tomakehurst:
Taking a ratcheting approach over time might be a good idea i.e. have a policy that all new tests should be Jupiter and that when you touch an old test class you port it to Jupiter.

tomakehurst:
I get that the JMock version being very old isn't ideal. If it was a production dependency I'd want to upgrade it with some urgency but since it's just in the test scope I'm a bit more relaxed until it actually blocks us from achieving something.

I tend not to do much object mocking now anyway, leaning more heavily on acceptance tests.

Again, perhaps a ratcheting policy for a gradual migration to mockito might be an approach we could take.

tomakehurst:
BTW, if you're keen to work on changes that would be high value to WireMock's users there's plenty that needs attention. Happy to point you at the important stuff if you'd like.

timtebeek:
Yeah happy to help; upgrading the tests was a good way to get some exposure to the code base, which I hope to apply on more "exciting" stuff.
Noticed there aren't yet any help-wanted or good-first-issue labels on the issue backlog; that might be helpful not just for me but others as well. :)

tomakehurst:
I keep meaning to add some help wanted labels, so this is a good reminder!

timtebeek:
Just in case it's useful in the future: [OpenRewrite now supports converting WireMockRules to WireMockExtension](https://github.com/openrewrite/rewrite-testing-frameworks/issues/170). :tada:

PR #1618, opened 2021-10-02
Delegate WireMockExtension.url(String) to WireMockServer

timtebeek:
(Final unsolicited MR, I promise :heart: )

In migrating both my own projects and [all project tests](https://github.com/wiremock/wiremock/pull/1616), I found that for [programmatic use](http://wiremock.org/docs/junit-jupiter/) there's a frequent need to compose URLs to instances of `WireMockExtension`. Since `WireMockExtension` does not expose the underlying `WireMockServer`, one is then forced to compose URLs by calling `baseUrl()` and appending/formatting the suffix. For such a common use case I think there's value in creating a direct delegate method to do the same.

As seen in `JUnitJupiterExtensionFailOnUnmatchedTest` this simplifies the URL composition, and reduces the need for `baseUrl()`. I understand it's not desirable to have delegates to all methods in `WireMockServer`, but for such a common use case I think it's warranted to make the development experience easier.

If you feel any different no problem to close the MR; I figured it's easier to get across the point in an MR rather than starting a separate issue for the discussion first.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
@@ -26,8 +26,6 @@ import org.junit.platform.commons.support.AnnotationSupport;
 
 import java.util.Optional;
 
-import static com.google.common.base.MoreObjects.firstNonNull;
-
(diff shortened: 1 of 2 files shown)


tomakehurst:
Thanks!

Issue #1418, opened 2021-01-19
Unable to validate Date-Time field is greater than or Less than in mappings json


PRABAREPO:
**My Request Body:**
{
  "username": "string",
  "Time": "2021-05-01T08:30:52.123Z",
  "Reference": "string"
}

**My Mappings JSON:**
{
    "request": {
        "method": "POST",
        "urlPattern": "/v1/test",
        "bodyPatterns": [
            {"matchesJsonPath" : "$[?(@.Time <= '2021-19-01T08:30:52.123Z')]"}  ]
    },
    "response": {
        "headers": {
            "Content-Type": "application/json"
        },
        "status": 200,
        "jsonBody":{
        "success": "Time is in Past"
        }
    }
}

Here I want to validate the Time attribute in request whether its less than than or equal to given time in mappings. 
But Its not working as that time field considered as String. 

Could you please help me on how to configure/validate that time attribute in Mappings

tomakehurst:
Please post questions to the mailing list

PR #1619, opened 2021-10-02
Draft: Http client 5.x

### LINKED ISSUE #627: HTTP DELETE body content is not forwarded in Proxy mode
I have a third party REST endpoint that has a DELETE resource that **requires** specific body content. However, WireMock in proxy mode does NOT pass on this content. In fact, it also strips out the Content-Type and Content-Length headers.

To properly pass-through requests to the proxy target, WireMock should not be stripping any contents of the original request.

1. Run WireMock standalone 2.5.1 with proxy target to an active listener.
2. Use a client to make a DELETE request with body data on the WireMock port.
3. Use the same client to make the same request directly to the listener.
4. Compare the outputs: There will be body content and content headers in the directly received request, but none in the WM-proxied request.

### LINKED ISSUE #1481: Upgrade to HTTP client 5.x
Track the latest and greatest now that it's GA.

Also, fix #627

timtebeek:
Fixes #1481 

At present still contains a number of TODO items related to DNS & SSL handing, as well as pointers where closer inspection for review might be warranted. These tests are still failing locally:
```
HttpsAcceptanceTest. randomDataOnSocketFault
HttpsBrowserProxyAcceptanceTest. *
HttpsBrowserProxyClientAuthAcceptanceTest. canDoClientAuthEndToEndWhenProxying
RecordingDslAcceptanceTest. generatesStubNameFromUrlPath // Name suffix dropped
ResponseDelayAcceptanceTest. requestTimesOutWhenDelayIsLongerThanSocketTimeout
ProxyResponseRendererTest. rejectsSelfSignedCertificateForForwardProxyingByDefault
HandlebarsXPathHelperTest. rendersNamespacedElement // whitespace
PlainTextDiffRendererTest. wrapsLargeXmlBodiesAppropriately // whitespace
```

I'm guessing it's pretty close to completion, but I'm not sure if I have the time or skills to complete it fully on short term. @tomakehurst would you be willing to take a look?

As noted in #1418 I guess it's best to start with an initial conversion to fix #627, and only in a second instance fix the usage of closeable resources, if that's OK.


timtebeek:
The `RecordingDslAcceptanceTest. generatesStubNameFromUrlPath` failure is quite puzzling; 
We do a request for
`/record-this/with$!/safe/ŃaMe?ignore=this`
which some how ends up as
`/record-this/with$!/safe/?aMe?ignore=this`
which is then mapped incorrectly to the stub name:
```
Expected: is "record-this_with_safe_name"
     but: was "record-this_with_safe"
```

Can't really pinpoint the exact cause of the change; possibly somewhere in the HttpClient call chain. Help needed. :)

tomakehurst:
This is a pretty large change to make in one go, and since it's changing the client for both tests and production code I feel like there's a bit of risk involved.

Can I suggest breaking this into two PRs - 1) production code upgrade to 5.x, keeping tests exactly the same, 2) upgrade test code to 5.x.

PR #1623, opened 2021-10-06
Return an Bad Request when invalid `since` parameter is provided

### LINKED ISSUE #1595: 500 error is thrown when invalid date supplied to since parameter on request log
Present in WireMock 2.31.0.

e.g. `/__admin/requests?since=2021-09-10T11222:11:10Z` causes an exception to be thrown (and thus 500 returned).

Should return a 400 with some context.

jamietanna:
As noted in #1595, we currently cause an Internal Server Error when an
invalid date is provided to the admin `/requests` API.

Instead of using the runtime exception `IllegalArgumentException`, which
is less nice to try-catch, we can use a checked exception
`ValidationException` which can then be mapped to a
`ValidationErrorResult` that can be serialised and returned to the
caller as a Bad Request.

Closes #1595.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/admin/Conversions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/admin/Conversions.java
@@ -16,6 +16,8 @@
 package com.github.tomakehurst.wiremock.admin;
 
 import com.fasterxml.jackson.databind.util.ISO8601DateFormat;
+import com.github.tomakehurst.wiremock.common.Errors;
+import com.github.tomakehurst.wiremock.common.InvalidInputException;
 import com.github.tomakehurst.wiremock.http.QueryParameter;
 
 import java.text.ParseException;
@@ -35,7 +37,7 @@ public class Conversions {
                 new ISO8601DateFormat().parse(parameter.firstValue()) :
                 null;
         } catch (ParseException e) {
-            throw new IllegalArgumentException(parameter.firstValue() + " is not a valid ISO8601 date");
+            throw new InvalidInputException(Errors.validation(parameter.key(), parameter.firstValue() + " is not a valid ISO8601 date"));
         }
     }
 }
--- a/src/main/java/com/github/tomakehurst/wiremock/admin/LimitAndSinceDatePaginator.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/admin/LimitAndSinceDatePaginator.java
@@ -15,6 +15,7 @@
  */
 package com.github.tomakehurst.wiremock.admin;
 
+import com.github.tomakehurst.wiremock.common.InvalidInputException;
 import com.github.tomakehurst.wiremock.http.Request;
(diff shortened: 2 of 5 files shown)


timtebeek:
Looks like you're running up against the recently added ArchUnit test that restricts adding new JUnit 4 tests.
This test was added to gently nudge new contributions to use JUnit Jupiter where reasonably possible, as per https://github.com/wiremock/wiremock/pull/1635.

```
JUnit4DetectionTest > junit4PackageShouldNotBeUsed FAILED
    java.lang.AssertionError: Architecture Violation [Priority: MEDIUM] - Rule 'org.junit should not be used, because we want to migrate to JUnit Jupiter' was violated (2 times):
    Method <com.github.tomakehurst.wiremock.AdminApiTest.getLoggedRequestsWithInvalidSinceDateReturnsBadRequest()> has annotation member of type <org.junit.Test$None> in (AdminApiTest.java:0)
    Method <com.github.tomakehurst.wiremock.AdminApiTest.getLoggedRequestsWithInvalidSinceDateReturnsBadRequest()> is annotated with <org.junit.Test> in (AdminApiTest.java:0)
```

We maintain a record of, for now, accepted violations in the code base, as stored in violation store.
You can lookup the accepted violations per rule in [src/test/resources/frozen/stored.rules](https://github.com/wiremock/wiremock/blob/master/src/test/resources/frozen/stored.rules).
In this case the associated violation store file is [src/test/resources/frozen/do-not-use-junit4](https://github.com/wiremock/wiremock/blob/master/src/test/resources/frozen/do-not-use-junit4).

There's roughly two ways to resolve this test failure:
1. Either change the test (and class) to use JUnit Jupiter instead of JUnit 4, aiding the migration.
2. Or add the new violations to the corresponding violation store; in this case `src/test/resources/frozen/do-not-use-junit4`.

In this particular case you're adding a test within a class that extends `AcceptanceTestBase`, which itself is still using JUnit 4 constructs, as are all it's other subclasses. As such you would not be able to adopt JUnit Jupiter without also migrating `AcceptanceTestBase` and all it's subclasses. That leans towards the "not reasonably possible" to migrate to JUnit Jupiter, so it's fine to add the violations to the store for now.

We might add an issue or pull request to separately migrate `AcceptanceTestBase`, as that should help bring the JUnit Jupiter adoption forward.

timtebeek:
When you update the violation store there's two MRs that can potentially cause conflicts as they have a lot of changes in the stored violations: #1651 & #1653. Just so you know! :)

ChunjuanLi:
Hi @tomakehurst  @timtebeek @jamietanna  I am trying to fix the issue however I am not able to reproduce the issue now, and it seems wiremock always provides a random ID whenever ID is Null/Empty:
code : https://github.com/wiremock/wiremock/blame/master/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java#L228 

Can you please confirm if this is still a valid bug? Thanks!

PR #1629, opened 2021-10-08
Publish Unit Test Results

timtebeek:
Uses https://github.com/marketplace/actions/publish-unit-test-results#running-as-a-composite-action


timtebeek:
There's a few alternative actions out there as well, such as:
 - https://github.com/marketplace/actions/test-reporter
 - https://github.com/marketplace/actions/junit-report-action
Chose this one as at appears able to aggregate across matrix builds.

Another alternative might be to look at Gradle Enterprise Scans, possibly.

PR #1630, opened 2021-10-08
Drop JMock

timtebeek:
Closes #1617 #1458 #1450


tomakehurst:
Hey Tim, there seem to be a few places where your IDE has reformatted blocks with 8 space tabs rather than 4. Please could you revert these to 4?

PR #1631, opened 2021-10-09
Publish test results from forks

timtebeek:
Improvement on #1629 , as forks were failing to publish.

As per: https://github.com/marketplace/actions/publish-unit-test-results#support-fork-repositories-and-dependabot-branches


EnricoMi:
This workflow is never triggered because it can only be trigged by "CI" workflow, whereas your testing workflow is called "Build and test".

https://github.com/wiremock/wiremock/blob/eda4d2aa9206d6e24751fab8ad91f3c9a4320e08/.github/workflows/publish-test-results.yml#L5

PR #1634, opened 2021-10-10
Detect unused code with ArchUnit

timtebeek:
This is an adaptation of https://github.com/timtebeek/archunit-unreferenced ;
it detects classes that are not referenced outside tests, and therefore likely unused.

For now I've not yet deleted any unused methods of classes, instead these are [frozen](https://www.archunit.org/userguide/html/000_Index.html#_freezing_arch_rules), meaning current violations are tolerated, but new accepted violations require an update to the freeze store in `src/test/resources/frozen`.

In my experience it's a great tool to detect and remove unused code paths, but the slight chance of false positives can make it annoying as well if those occur too often. For common patterns of accepted unreferenced code we can expand the rules.

Probably best to also remove the existing violations, either as part of this MR, or separately.


timtebeek:
Fine to merge this one as-is; although we probably can delete quite a few more detected (but public!) methods still.

PR #1635, opened 2021-10-10
Detect & limit JUnit4 usage with ArchUnit

timtebeek:
Since we want to migrate to JUnit Jupiter, this change will allow us to detect & mark all existing usage of JUni4 as accepted violations, but not allow new JUnit4 tests to be added (without updating the freeze store).

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ following guidelines:
 on the mailing list to generate a discussion. This will avoid significant amounts of coding time spent on changes that ultimately get rejected.
 * Try to avoid reformats of files that change the indentation, tabs to spaces etc., as this makes reviewing diffs much
 more difficult.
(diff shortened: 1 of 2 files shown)


timtebeek:
Whenever tests are converted to JUnit Jupiter a local test run will automatically update the freeze store, which will then show up in the diff for that MR. That way we can keep tabs on how many tests still need to be converted, and once that number is down far enough we can again try automation with https://docs.openrewrite.org/tutorials/migrate-from-junit-4-to-junit-5, similar to what was started in #1616; they've even since added support for migrating WireMockRule to WireMockExtension. :tada:

tomakehurst:
Is it possible to give the freeze files meaningful names rather than just UUIDs?

Since they're very large, it might be nice not to have to rely on eyeballing them to see what they're for.

PR #1636, opened 2021-10-10
Apply ArchUnit GeneralCodingRules

timtebeek:
Still fails on these two instances of calling `printStackTrace`:
```
Method <com.github.tomakehurst.wiremock.jetty9.JettyHttpsFaultInjector$1.failed(java.lang.Throwable)> calls method <java.io.IOException.printStackTrace()> in (JettyHttpsFaultInjector.java:97)
Method <com.github.tomakehurst.wiremock.jetty9.JettyHttpsFaultInjector$1.succeeded()> calls method <java.io.IOException.printStackTrace()> in (JettyHttpsFaultInjector.java:88)
```
Seems to me like we'd want to do something else there; logging perhaps?


timtebeek:
Once those two violations are changed the tests should pass again; meaning we get four free checks on every code contribution.

PR #1648, opened 2021-10-14
Fixes #1647 by dereferencing wireMockServer after stopping.

### LINKED ISSUE #1647: Reusing WireMockExtension from abstract base class fails
We are running into an issue with the JUnit 5 WireMockExtension. When the extension is defined as a static final on an abstract base class and re-used in several child classes. The WireMock server is stopped after the first test class finishes (@AfterAll) and not restarted when the second test class is executed (@BeforeAll).

The second test class will execute `beforeAll` and call `startServerIfRequired`. The `wireMockServer` field is already initialized by the first test class and skip initialization of the WireMock server. The test will then try to connect to a stopped server.

A solution could be to dereference the WireMock server in the `stopServerIfRunning` method. Causing the second test to re-initialize the WireMock server:

```
private void stopServerIfRunning() {
    if (wireMockServer != null && wireMockServer.isRunning()) {
        wireMockServer.stop();
    }
   wireMockServer = null;
}
```

WireMock version: 2.31.0
Reproducer: https://github.com/timpeeters/wiremock-bug

timpeeters:
Fixes #1647 by dereferencing wireMockServer after stopping.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
@@ -119,9 +119,10 @@ public class WireMockExtension extends DslWrapper implements ParameterResolver, 
     }
 
     private void stopServerIfRunning() {
-        if (wireMockServer.isRunning()) {
+        if (wireMockServer != null && wireMockServer.isRunning()) {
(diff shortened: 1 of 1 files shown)


tomakehurst:
Already fixed in c4ece200f318a35a12f281cfc6bb2903800d904f

PR #1656, opened 2021-10-17
Test for `Strings#wrapIfLongestLineExceedsLimit` method.

nstdio:


oleg-nenashev (commented):
Sorry if that's a necroposting, catching up on historical PRs

The performance of the implementation depends on the JDK and on parallelization. I **guess**, in the majority of the use-cases it would be an overhead, because `PlainTextDiffRenderer` operates with smaller sizes if text entries most of the time, and brute force iterator is more efficient.

Moreover, the only usage of the method is `wrapIfLongestLineExceedsLimit` which in fact does *NOT* need calculation of a longest line size, but just a just finding a line that exceeds a limit. If we find for performance, I would suggest a different utility method.  Also, in fact `wrapIfLongestLineExceedsLimit` does another string conversion that also impacts performance

If you @nstdio are still interested, I would suggest to:

1. Merge this PR as a test one
2. Adept the wrapping methods instead for better performance, maybe by making all of them to operate via line streams instead of intermediate String conversions
3

nstdio:
Hey, @oleg-nenashev! Thanks for you're thoughts. My intent was very modest: write it in a more expressive way. Frankly, I'm not particularly interested in merging this after couple of years, especially when old version does the job very well :slightly_smiling_face:

oleg-nenashev:
@nstdio yeah, fair point. I joined the co-maintainers in July, so it takes some time to clear the backlog of PRs. Indeed it's weird to ask after 2 years, but it does not hurt. If you're fine with that, I will merge in the tests part.

src/main/java/com/github/tomakehurst/wiremock/common/Strings.java
```diff
@@ -20,13 +20,15 @@
 import static org.apache.commons.lang3.StringUtils.getLevenshteinDistance;
 
 import java.nio.charset.Charset;
+import java.util.regex.Pattern;
```
oleg-nenashev, line 23:
```suggestion
```

src/main/java/com/github/tomakehurst/wiremock/common/Strings.java
```diff
@@ -20,13 +20,15 @@
...
 import org.apache.commons.lang3.text.WordUtils;
 
 public class Strings {
 
   private Strings() {}
 
   public static final Charset DEFAULT_CHARSET = UTF_8;
+  private static final Pattern NEW_LINE_PATTERN = Pattern.compile("\n", Pattern.LITERAL);
```
oleg-nenashev, line 31:
```suggestion
```

src/main/java/com/github/tomakehurst/wiremock/common/Strings.java
```diff
@@ -63,16 +65,7 @@ public static String wrapIfLongestLineExceedsLimit(String s, int maxLineLength)
...
-    int longestLength = 0;
-    for (String line : lines) {
-      int length = line.length();
-      if (length > longestLength) {
-        longestLength = length;
-      }
-    }
-
-    return longestLength;
+    return NEW_LINE_PATTERN.splitAsStream(s).mapToInt(String::length).max().orElse(0);
```
oleg-nenashev, line 68:
```suggestion
    String[] lines = s.split("\n");
    int longestLength = 0;
    for (String line : lines) {
      int length = line.length();
      if (length > longestLength) {
        longestLength = length;
      }
    }

    return longestLength;
```

oleg-nenashev (approved):
Thanks for the update. Works for me as extra test coverage

oleg-nenashev (commented):
So the test failure is legit:

```
StringsTest > shouldWrapIfLongestLineExceedsLimit(String, String, int) > com.github.tomakehurst.wiremock.common.StringsTest.shouldWrapIfLongestLineExceedsLimit(String, String, int)[4] FAILED
    org.opentest4j.AssertionFailedError: expected: <abc
    ab
    cd
    abc
    d> but was: <abc
    ab
    cd
    abc
    d>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertEquals.failNotEqual(AssertEquals.java:197)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:182)
        at app//org.junit.jupiter.api.AssertEquals.assertEquals(AssertEquals.java:177)
        at app//org.junit.jupiter.api.Assertions.assertEquals(Assertions.java:1[141](https://github.com/wiremock/wiremock/actions/runs/5776401996/job/15655484729?pr=1656#step:9:142))
        at app//com.github.tomakehurst.wiremock.common.StringsTest.shouldWrapIfLongestLineExceedsLimit(StringsTest.java:43)
```

PR #1686, opened 2021-11-02
Delegate parsing of media types to library

jamietanna:
Instead of us handling any media-type related parsing, we should instead
delegate to a library for this.

Improvement to #1657.

---

Note that there are style issues, but it looks quite painful to solve without #1563


jamietanna:
@oleg-nenashev yes if you're still happy with it!

PR #1694, opened 2021-11-05
Extend LogNormal DelayDistribution to add support for CappedLogNormal

paul-poulain-vodafone:
A new `DelayDistribution` called `CappedLogNormal` has been added which behaves the same as it's superclass,  `LogNormal` except that it has an extra parameter which is the maximum value that should be used for the distribution delay. This is to prevent excessively long tails from the `LogNormal` potentially causing timeouts when simulated downstream systems would have timed out themselves sooner. An example configuration in a JSON mapping file (from the updated docs) might look as follows:

```json
"delayDistribution": {
        "type": "cappedlognormal",
        "median": 80,
        "sigma": 0.4,
        "maxValue": 200
}
```

which means the generated value will never exceed 200ms. 

## Docs entry

```
### Capped Lognormal delay

Whilst lognormal distribution is a pretty good approximation of long tailed
latencies centered on the 50th percentile, downstream production systems 
often have timeouts configured which would normally artificially truncate the
long tail. Therefore, the capped log normal takes an extra parameter which is
the maximum value that the delay distribution should use. This will have the 
effect of a main spike of responses around the 50th percentile but another
(generally smaller but that would depend on the sigma value) spike at the
maximum value.

It takes the same two parameters as lognormal plus an extra one:

-   median - The 50th percentile of latencies.
-   sigma - Standard deviation. The larger the value, the longer
    the tail.
-   maxValue - The maximum value at which any generated value should be capped.
 A good value to use for this might be the production maximum value or 95th
 percentile.


To use, instantiate a `new CappedLogNormal(median, sigma, maxValue)`, or via JSON:

```json
"delayDistribution": {
        "type": "cappedlognormal",
        "median": 80,
        "sigma": 0.4,
        "maxValue": 200
}
```
```


paul-poulain-vodafone:
Apologies about the slightly foo-barred commit history!

paul-poulain-vodafone:
Hey @tomakehurst , 

Thanks for your response and no worries. 

That is a good shout and I'm not sure why we didn't consider that originally. I'll get an update out when I get a chance. We've also got a small additional change to resample rather than capping to prevent having a secondary spike at the end of the long tail, so that'll be added in too.

paul-poulain-vodafone:
Abandoning this PR in favour of #3026 that enhances `LogNormal` rather adding a new `DelayDistribution` subclass, as @tomakehurst sensibly suggested, thanks.

PR #1695, opened 2021-11-05
Run tests on JDK 17

### LINKED ISSUE #1655: Build on & fully support JDK 17
Had a brief look at building on JDK 17, branched off the work Jamie's done for Gradle 7 in https://github.com/wiremock/wiremock/issues/1633.

It appears the use of `sun.security.x509` [crypto](https://github.com/wiremock/wiremock/blob/master/src/test/java/com/github/tomakehurst/wiremock/crypto/) is problematic on JDK 17.
From what I can see this was mostly contributed mid last year by @Mahoney
Hoping he can advice on what's best to future proof this bit of code.
Bouncy castle seems rather hefty; perhaps we can spin this off into a separate extension?

timtebeek:
Fixes #1655 

- [x] The trailing newlines now expected are puzzling; might need to split tests using Enabled/Disabled ranges, or widen assertions from `is` to `matches` with a regex
- [ ] Local `XmlTest.printsNamespacedXmlWhenPrefixDeclarationNotInScope` failure with
```
javax.xml.transform.TransformerException: java.lang.RuntimeException: Namespace for prefix 'fl' has not been declared.
	at java.xml/com.sun.org.apache.xalan.internal.xsltc.trax.TransformerImpl.transform(TransformerImpl.java:791)
	at java.xml/com.sun.org.apache.xalan.internal.xsltc.trax.TransformerImpl.transform(TransformerImpl.java:397)
	at com.github.tomakehurst.wiremock.common.xml.XmlNode.render(XmlNode.java:140)
	at com.github.tomakehurst.wiremock.common.xml.XmlNode.toString(XmlNode.java:129)
	at com.github.tomakehurst.wiremock.common.ListOrSingle.toString(ListOrSingle.java:41)
```
We only now enter the catch block of a method documented as nasty little hack:
https://github.com/wiremock/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/common/xml/XmlNode.java#L164
Unsure of the proper way to proceed to make this test work again.


src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
```diff
@@ -99,13 +99,9 @@ public static String prettyPrint(Document doc) {
...
 
     private static TransformerFactory createTransformerFactory() {
-        try {
-            TransformerFactory transformerFactory = (TransformerFactory) Class.forName("com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl").getDeclaredConstructor().newInstance();
-            transformerFactory.setAttribute("indent-number", 2);
-            return transformerFactory;
-        } catch (Exception e) {
-            return TransformerFactory.newInstance();
-        }
+        TransformerFactory transformerFactory = TransformerFactory.newInstance();
```
timtebeek, line 102: TransformerFactory.newInstance has a fallback to `com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl`, so it seemed to me like we can call this directly.

src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
```diff
@@ -99,13 +99,9 @@ public static String prettyPrint(Document doc) {
...
+        transformerFactory.setAttribute("indent-number", 2);
```
timtebeek, line 103: Our tests expect identation with 2 spaces; which in the catch block previously triggered on JDK 17 was not set.

src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
```diff
@@ -99,17 +99,9 @@ public static String prettyPrint(Document doc) {
...
-    try {
-      TransformerFactory transformerFactory =
-          (TransformerFactory)
-              Class.forName("com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl")
-                  .getDeclaredConstructor()
-                  .newInstance();
-      transformerFactory.setAttribute("indent-number", 2);
-      return transformerFactory;
-    } catch (Exception e) {
-      return TransformerFactory.newInstance();
```
timtebeek, line 111: PlainTextDiffRendererTest.wrapsLargeXmlBodiesAppropriatelyJre11() would fail with the following exception without this change:
```
java.lang.IllegalAccessException: class com.github.tomakehurst.wiremock.common.xml.Xml cannot access class com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl (in module java.xml) because module java.xml does not export com.sun.org.apache.xalan.internal.xsltc.trax to unnamed module @37858383
```

PR #1701, opened 2021-11-19
Resolve common static analysis issues

timtebeek:
Via: https://docs.openrewrite.org/reference/recipes/java/cleanup/commonstaticanalysis

Disabled:
org.openrewrite.java.cleanup.AddSerialVersionUidToSerializable
org.openrewrite.java.cleanup.ControlFlowIndentation
org.openrewrite.java.cleanup.EqualsAvoidsNull
org.openrewrite.java.cleanup.ExplicitInitialization
org.openrewrite.java.cleanup.HiddenField
org.openrewrite.java.cleanup.HideUtilityClassConstructor
org.openrewrite.java.cleanup.MethodNameCasing
org.openrewrite.java.cleanup.NeedBraces
org.openrewrite.java.cleanup.UnnecessaryThrows


timtebeek:
Some of these are perhaps a bit nitpicky, but overall they help the code readability. I had disabled a few that caused too many changes for little added value.

Also the added `final` here and there might technically be breaking a public API. I leave it up to you whether those classes were actually intended to be open for extension/override or not.

timtebeek:
@jamietanna Would you mind reviewing this one? :)

jamietanna (commented):
For the classes now being marked as `final`, does this have an impact we'll need to be careful of / at least call out in the release notes for consumers mocking/extending them?

src/test/java/com/github/tomakehurst/wiremock/BindAddressTest.java
```diff
@@ -78,23 +79,16 @@ public void stop() {
   @Test
   public void shouldRespondInTheBindAddressOnlyOnHttp() throws Exception {
     executeGetIn(localhost);
-    try {
-      executeGetIn(nonBindAddress);
-      fail("Should not accept HTTP connection to [" + nonBindAddress + "]");
-    } catch (Exception ex) {
-    }
+
+    assertThrows(RuntimeException.class, () -> executeGetIn(nonBindAddress));
```
jamietanna, line 83: We've changed the intent of this test, right? Should be:

```suggestion
    assertThrows(Exception.class, () -> executeGetIn(nonBindAddress));
```

To match the previous intent?

src/test/java/com/github/tomakehurst/wiremock/LogTimingAcceptanceTest.java
```diff
@@ -54,12 +54,12 @@ public void serveEventIncludesTotalAndServeDuration() {
 
   @Test
   public void includesAddedDelayInTotalWhenAsync() {
-    final int DELAY = 500;
+    final int delay = 500;
```
jamietanna, line 57: We should probably follow up and remove the `final`s from local varialbes, as I don't think we want them

src/test/java/com/github/tomakehurst/wiremock/matching/MockRequest.java
```diff
@@ -246,20 +246,9 @@ public boolean isMultipart() {
...
-                      return true;
-                    }
-                    return false;
-                  }
-                })
-            .get()
-        : null;
+    return getParts() == null || name == null
+        ? null
+        : from(multiparts).firstMatch(input -> name.equals(input.getName())).get();
```
jamietanna, line 251: I'd personally prefer we get rid of the ternary expression and replace it with an `if ... else`

jamietanna (commented):
Couple of minor comments, but looking good, thanks for doing this :raised_hands:

src/test/java/com/github/tomakehurst/wiremock/LogTimingAcceptanceTest.java
timtebeek: Done; made the diff quite a bit larger, with more potential for conflicts, but should be OK if merged soon-ish.

src/test/java/com/github/tomakehurst/wiremock/matching/MockRequest.java
timtebeek: Better indeed; changed.

timtebeek:
Thanks for the suggestions; all applied.

As for:
> For the classes now being marked as final, does this have an impact we'll need to be careful of / at least call out in the release notes for consumers mocking/extending them?

All of these now final classes already have a private constructor:
```
src/main/java/com/github/tomakehurst/wiremock/common/DateTimeParser.java
src/main/java/com/github/tomakehurst/wiremock/common/StreamSources.java
src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/HttpTemplateCacheKey.java
src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestLine.java
src/main/java/com/github/tomakehurst/wiremock/jetty9/JettyUtils.java
com.github.tomakehurst.wiremock.common.JettySettings.Builder
com.github.tomakehurst.wiremock.extension.responsetemplating.RenderCache.Key
```

So it would only affect users that mock those classes, in which case I really wonder why.
I'd say make them final for now, and open them up again if there's a convincing case.

jamietanna:
Sounds like a plan, thanks Tim - I didn't look that far :sweat_smile:

src/main/java/com/github/tomakehurst/wiremock/http/CaseInsensitiveKey.java
```diff
@@ -39,9 +39,7 @@ public boolean equals(Object o) {
 
     CaseInsensitiveKey that = (CaseInsensitiveKey) o;
 
-    if (key != null ? !key.equalsIgnoreCase(that.key) : that.key != null) return false;
-
-    return true;
+    return key == null ? that.key == null : key.equalsIgnoreCase(that.key);
```
jamietanna, line 42: personally, I'd prefer this to be:

```suggestion
    if (key == null) {
      return that.key == null;
    } else {
      key.equalsIgnoreCase(that.key);
    }
```

jamietanna:
Thanks Tim, overall changes are looking good, but I'd probably like Tom just to double check re the `final`s :+1:

src/main/java/com/github/tomakehurst/wiremock/http/CaseInsensitiveKey.java
timtebeek:
```suggestion
    if (key == null) {
      return that.key == null;
    }
    return key.equalsIgnoreCase(that.key);
```
You were missing a return statement; would you be OK with this instead? I don't like unnecessarily nested else statements when we can avoid them with a return.

jamietanna: Woops, good spot.

Yep, I like that too, that'd be appreciated 👍🏽

timtebeek:
Closing this one due to inactivity & conflicts; might want to cherry pick some of the manual fixes, and then optionally recreate the changes with the [updated common static analysis automation](https://docs.openrewrite.org/reference/recipes/java/cleanup/commonstaticanalysis).

PR #1707, opened 2021-11-23
Migration to jetty 11.

vmakieienko:
Please review my atempt to migrate your wanderfull project to jetty 11.
- Migrated code to compile with jetty 11 dependencies.
- 40 tests relates to HTTPS still fail.
- I had to create javax.servlet.http.HttpServletRequest to satisfy commons-fileupload-1.4. (Should be possible to remove with commons-fileupload-2.0 (https://issues.apache.org/jira/browse/FILEUPLOAD-309)

If you know in what direction to move with tests or other issue please let me know.


tomakehurst:
I'd love to be able to merge this but it would mean dropping support for Java 8, which we can't do for a while yet and not until at least v3.0.

We'd also need to get all the HTTPS tests passing and figure out a replacement for commons-fileupload, or an adaptation strategy less messy than recreating legacy interfaces.

vmakieienko:
> I'd love to be able to merge this but it would mean dropping support for Java 8, which we can't do for a while yet and not until at least v3.0.

I agree that java8 support is important. I think there could be v2 as is and v3 with jetty11 and java version > 8. I think there are projects needing it already. What would be the best way to do it?

> We'd also need to get all the HTTPS tests passing and figure out a replacement for commons-fileupload, or an adaptation strategy less messy than recreating legacy interfaces.

Looks like your knowlage of https/ssl is much beter than mine. Can you take a look at one of errors and point me in to the right direction of reading/investigation? For instance, test `HttpClientFactoryAcceptsTrustedCertificatesTest(trusted:"localhost"), certificateCN:"other.com", validCertificate?:true)` fails with respons containing `400: Invalid SNI`

tomakehurst:
Closing as we're now on Jetty 11

PR #1708, opened 2021-11-23
Using TCK Tested JDK builds of OpenJDK

carldea:
The AdoptOpenJDK has been discontinued since July 2021 (https://adoptopenjdk.net). Switching the distribution to Azul Zulu. When using Zulu you get all the latest updated (TCK Tested) builds for all versions of OpenJDK.
 
Please check out my article on Foojay.io:
https://foojay.io/today/github-actions-with-java-part-2/

Some of the tests fail on timing issues. I noticed them very close. Most of them pass.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        jdk: [8, 11]
+        jdk: [8.0.192, 8, 11.0.3, 11]
     runs-on: ${{ matrix.os }}
     env:
       JDK_VERSION:  ${{ matrix.jdk }}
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        jdk: [8, 11]
+        jdk: [8.0.192, 8, 11.0.3, 11]
```
tomakehurst, line 18: What's the purpose of having both major and patch version matrix versions?

carldea: @tomakehurst,
 
Glad you asked.

In short, it's good practice because vendors or projects that use this library often are in production with major fixed release versions, and the latest build can/has failed tests due to something in the build and nothing to do with your code.

Here's my write-up about fixed (major) releases:
[https://foojay.io/today/github-actions-with-java-part-2/](url)

But if you prefer I can remove the fixed versions. It's up to you.
Thank you,
Carl

oleg-nenashev (commented):
We are using the Timurin distributions in the recent versions. It is now verified against TCK: https://devops.com/eclipse-foundation-touts-java-se-temurin-certification-momentum/ , so I will close this pull request

PR #1717, opened 2021-12-02
Add missing `description` to the slim JAR's generated POM

jamietanna:
It's unclear as to why this is no longer working, but as it's currently
blocking the release being published, we'll resolve it for now.


tomakehurst:
Thanks!

PR #1721, opened 2021-12-02
add admin API endpoints to set specfic scenario states.

willKo:
## description 
This PR adds two admin endpoints:
### get the current state of a scenario: 
GET `__admin/scenario/state?stateName=scenario-name`
### set the current state of a scenario 
POST `__admin/scenario/state`
Body: 
```json
{ 
"scenarioName": "scenario-name",
"scenarioState": "scenario-state"
}
```

### further notes 
This allows setting a scenario to any state without knowing the current state. 
The workaround discussed in PR https://github.com/wiremock/wiremock/pull/1079 and issue https://github.com/wiremock/wiremock/issues/847 only allows resetting the scenario if you know the current state. 
Also, I do not think that it is the ideal usage to create additional "control mappings" to  achieve `__admin` functionality, therefore here is an alternative proposal. 
In contrast to PR https://github.com/wiremock/wiremock/pull/1079 this PR not just allows resetting a scenario, you can also switch to an intermediate or final state of a scenario. 

Would also enable PR:
https://github.com/wiremock/wiremock/pull/1079
and also fix issue
https://github.com/wiremock/wiremock/issues/847
(when setting scenario state in POST-Request to "Started") 

### Discussion 
If this PR is correct according to the Contribution Rules, please let me know, so I can fix it. 
If the API changes are not desirable (e.g. a path parameter would be nicer for the "state-name" in the GET request), please let me know.

### DIFF
--- a/docs-v2/_docs/stateful-behaviour.md
+++ b/docs-v2/_docs/stateful-behaviour.md
@@ -142,7 +142,21 @@ The state of all configured scenarios can be reset back to
 `Scenario.START` either by calling `WireMock.resetAllScenarios()` in
 Java, or posting an empty request to `http://<host>:<port>/__admin/scenarios/reset`.
 
+## Working with specific scenarios
 
+### Getting a specific scenario state
+The state of a specific scenario can be retrieved for a specific state.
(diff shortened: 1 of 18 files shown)


willKo:
pretty similar to https://github.com/wiremock/wiremock/pull/1046

PR #1726, opened 2021-12-07
chore: probably dead links

ldicarlo:


ldicarlo:
Actually there are other dead links if you want I can fix them

tomakehurst:
These links are both still valid - both link to working pages:
http://wiremock.org/docs/configuration/
http://wiremock.org/docs/junit-rule/

ldicarlo:
Wow they didn't work yesterday. This is strange, but yes you are right.

Issue #1584, opened 2021-08-30
MALFORMED_RESPONSE_CHUNK with the Accept-Encoding: [gzip] does not send the required HTTP 200 first.


andrew-yang:
- WireMocker version 2.25.1
- Starting up WireMock programmatically for integration tests. (Used to mockout a OAuth2 Resource Server endpoint)

I haven't yet created a sample project to reproduce, but the problem seems pretty simple. 
I have a stub that is supposed to result in a MALFORMED_RESPONSE_CHUNK. When I hit that stub with a request that has Accept-Encoding: gzip, the HTTP 200 response isn't sent only the garbage "lskdu018973t09sylgasjkfg1][]'./.sdlv". 
When I send the same request without the Accept-Encoding header, I properly get the HTTP 200 and then garbage data after.

Some debugging has led me into Jetty specifically here:
https://github.com/eclipse/jetty.project/blob/4be1e63bc11d1f63ff035018158383fad8d46055/jetty-server/src/main/java/org/eclipse/jetty/server/handler/gzip/GzipHttpOutputInterceptor.java#L125-L131

So it looks like the first HTTP 200 sent using flushBuffer is dropped on the floor instead of being sent over since nothing is in the buffer.
```
gzip:152, GzipHttpOutputInterceptor (wiremock.org.eclipse.jetty.server.handler.gzip)
commit:246, GzipHttpOutputInterceptor (wiremock.org.eclipse.jetty.server.handler.gzip)
write:108, GzipHttpOutputInterceptor (wiremock.org.eclipse.jetty.server.handler.gzip)
write:250, HttpOutput (wiremock.org.eclipse.jetty.server)
write:226, HttpOutput (wiremock.org.eclipse.jetty.server)
flush:408, HttpOutput (wiremock.org.eclipse.jetty.server)
flushBuffer:1039, Response (wiremock.org.eclipse.jetty.server)
malformedResponseChunk:70, JettyFaultInjector (com.github.tomakehurst.wiremock.jetty9)
...
run:834, Thread (java.lang)
```

tomakehurst:
Thanks for reporting this.

Does this also happen with the latest WireMock version?

andrew-yang:
Just quickly bumped my project to 2.30.1 and it looks like I still only see the garbage data getting read.

PR #1735, opened 2021-12-10
Issue 1584

deepti29jl:


timtebeek:
Just adding a link here to the issues for easier cross reference:
#1584: MALFORMED_RESPONSE_CHUNK with the Accept-Encoding: [gzip] does not send the required HTTP 200 first.

PR #1738, opened 2021-12-13
Fix admin requests with empty body to avoid response code 411

### LINKED ISSUE #1729: WireMock admin client reset fails with 411 through some proxies
When WireMock is put behind a proxy which handles HTTP specification strictly then invoking `reset()` (`POST__admin/reset`) or `resetMappings` (`POST __admin/mappings/reset`) through its client will fail with getting HTTP status 411.

The reason is that these HTTP calls are without request body and the underlying client doesn't add Content-Length=0 header either.

The issue originates from apache httpclient org.apache.hc.core5.http.protocol.RequestContent#process which skips adding Content-Length header (and also chunked encoding) when the body entity is null.

If this is not considered an apache httpclient issue then a network interceptor should be added in WireMock client to work it around.

WireMock version is 2.32.0

danielimre:
Changed admin client to add empty response entity for HTTP requests without body to have proper Content-Length header set in every case. Closes #1729

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
@@ -499,8 +499,10 @@ public class HttpAdminClient implements Admin {
     ClassicRequestBuilder requestBuilder =
         ClassicRequestBuilder.create(requestSpec.method().getName()).setUri(url);
 
-    requestBuilder.setEntity(
-        jsonStringEntity(Optional.ofNullable(requestBody).map(Json::write).orElse("")));
+    if (requestSpec.method().hasEntity()) {
+      requestBuilder.setEntity(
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
IMHO it is still valid

tomakehurst:
Think we still need a test case for this. Although this code is exercised generally by the acceptance tests, there's nothing that produces the problem this is attempting to fix.

oleg-nenashev:
@danielimre are you still interested in this pull request?

danielimre:
@oleg-nenashev sure. Let me check if I can create a reasonable test for this.

danielimre:
@oleg-nenashev I've put together some tests backing the changes. The simplest solution I've found without doing anything HTTP client specific stuff is to make use of wiremock itself to verify the behavior.

oleg-nenashev (approved):
Thanks for adding the test coverage!

oleg-nenashev:
Thanks @danielimre!

Issue #1399, opened 2020-10-26
On record save JSON data as `jsonBody`


martin-v:
`jsonBody` definitely makes the mock files much more readable, but unfortunately the record does not use this feature.
It would be really nice if the recorder persist new mocks with JSON data using this feature.

Probably I can implement this function if it is desired.

artmac:
I agree, because encoded json in as text is not really readable. And manually converting files to jsonBody requires effort which could be avoided.
 
@martin-v have you started to implemented?

martin-v:
Not yet. I wait for a response of the maintainer. No desire to do the wasteful work ;)

@tomakehurst WDYT?

tomakehurst:
I'd be happy to merge a PR implementing this.

Some things to consider:

* What happens if the response from the target server isn't valid JSON? Do we fall back on using a string? Just log a warning? Refuse to record the stub?
* To convert from a string to a JSON object you'll have to deserialise it, which means you'll be subject to however the `ObjectMapper` you're using is configured. This could mean e.g. conversion of nulls to removal of attribute keys.

martin-v:
I would suggest that the set of criteria for using `jsonBody` is:
* The `Content-Type` is one of 
```
application/json
application/x-json
application/ld+json
application/hal+json
application/javascript
application/x-javascript
text/json
text/x-json
text/javascript
text/x-javascript
```

* The size isn't to big, e.g. below 5MB
* The complete response ins valid JSON

Otherwise we fallback zu string.

Anything to add?

zufallsgenerator:
How about adding an option to activate this? While I'd find this very useful, I'd imagine people would be confused if the behaviour suddenly changes with a version upgrade.

alexshultsg:
The work around is to start recording by making POST call to /__admin/recordings/start
with, for example, this
{
  "targetBaseUrl": "<your target>",
  "extractBodyCriteria" : {
    "textSizeThreshold" : "10"
  }
}
and POST /__admin/recordings/stop to stop recording
body file will be saved in __files and mapping witl have bodyFileName field instead of body

tomakehurst:
> I would suggest that the set of criteria for using `jsonBody` is:
> 
> * The `Content-Type` is one of
> 
> ```
> application/json
> application/x-json
> application/ld+json
> application/hal+json
> application/javascript
> application/x-javascript
> text/json
> text/x-json
> text/javascript
> text/x-javascript
> ```
> 
[...]

PR #1747, opened 2021-12-17
Convert string body to json body if the value of content-type header contains 'json'.

Calvin-he:
Try to convert string body to json body if the value of content-type header contains 'json'. 
It will solve the Enhancement #1399: On record save JSON data as jsonBody.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
@@ -24,6 +24,7 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.github.tomakehurst.wiremock.common.Json;
 import com.github.tomakehurst.wiremock.extension.Parameters;
 import com.github.tomakehurst.wiremock.http.*;
+import com.github.tomakehurst.wiremock.matching.ContainsPattern;
 import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
```diff
@@ -262,6 +263,23 @@ private boolean isBinaryBody() {
...
+    if (jsonBody != null) {
+        return true;
+    }
+    if(headers == null || stringBody == null) {
+        return false;
+    }
+    HttpHeader contentTypeHeader = headers.stream()
+        .filter(header -> header.keyEquals(ContentTypeHeader.KEY)).findFirst().orElse(null);
+    if(contentTypeHeader != null) {
+        boolean isJson = contentTypeHeader.hasValueMatching(new ContainsPattern("json"));
```
jamietanna, line 275: @tomakehurst this may be one of those areas we'd talked about with respect to needing to do "proper" content-negotation, as I'd recommend we check that it is compatible with the media type subtype `*+json`

tomakehurst:
There's an edge case I'm concerned about here, which is users who are intentionally sending malformed JSON for negative testing. If we make this change, that won't be possible any more as the parser will refuse to create object when the stub is being built.

oleg-nenashev:
@Calvin-he I guess it can be closed given outstanding comment from Tom, right?

PR #1751, opened 2021-12-19
Temurin JDK

sullis:
NOTE: Adopt OpenJDK got moved to Eclipse Temurin and won't be updated anymore. It is highly recommended to migrate workflows from adopt to temurin to keep receiving software and security updates.

https://github.com/actions/setup-java/blob/main/README.md
﻿


tomakehurst:
Let's give it a try shall we?

PR #1758, opened 2021-12-24
docs: Fix example Gradle configuration

samhclark:
Currently, the instructions for using WireMock with Gradle are incorrect. The Java 8 and deprecated Java 7 artifacts are swapped, and the deprecated Java 7 artifact uses the current version (when it should end at 2.27.2). 

This PR corrects this, so the Gradle instructions match the Maven instructions. 

Please let me know if there is anything else I need to do (like running the `gen-docs.sh` script?), or if this PR is otherwise unacceptable for some reason. 

I know this is a small thing, but I've seen people pull in the old Java 7 artifact when copy-pasting from here on a new project.

### DIFF
--- a/docs-v2/_docs/download-and-installation.md
+++ b/docs-v2/_docs/download-and-installation.md
@@ -74,25 +74,25 @@ Java 7 standalone (deprecated):
 Java 8:
 
 ```groovy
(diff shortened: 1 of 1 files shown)


tomakehurst:
Wow, don't know how that stayed under the radar for so long.

Thanks!

PR #1770, opened 2022-01-18
Add missing backslashes in docker commands

lucianoqueiroz:
It fixes docker commands in docker section documentation


tomakehurst:
Thanks!

PR #1772, opened 2022-01-19
perf: Avoid double parsing response JSON body

htmldoug:
For a single request, we're double parsing the JSON body. The flame graph below shows an 18 KB JSON response--not particularly large. Eliminating the double parsing would cut out ~25% of the work.

ResponseDefinition getJsonBody

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateTransformer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateTransformer.java
@@ -103,7 +103,7 @@ public class ResponseTemplateTransformer extends ResponseDefinitionTransformer
             .build();
 
     if (responseDefinition.specifiesTextBodyContent()) {
-      boolean isJsonBody = responseDefinition.getJsonBody() != null;
+      boolean isJsonBody = responseDefinition.isJsonBody();
       HandlebarsOptimizedTemplate bodyTemplate =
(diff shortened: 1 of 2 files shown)


jamietanna (approved):
Looks good to me - @tomakehurst thoughts?

I'd say we'd be good to squash those two commits, as the license header change is brought in by Spotless as part of that prior commit, but as we usually squash-and-merge anyway, no need for you to do it now :+1:

htmldoug:
Thanks for the quick review! Commits squashed. (It needed a rebuild anyway.)

Would you be open to more aggressive changes to eliminate all four of those big bars?

All the JSON steps seem avoidable if we can get fiddle with the interfaces. Currently for `ResponseTemplateTransformer.transform`:
1. `byte[]` => `JsonNode` (unused)
2. ~`byte[]` => `JsonNode`~ (unused)
3. `byte[]` => `String` (handlebars input)
4. `String` => `JsonNode` (no additional transformation happens)
5. `JsonNode` => `byte[]`

The main barrier is that `Body` and `ResponseDefinitionBuilder` don't allow expressing JSON bodies without an intermediate `JsonNode`. Would you be open to changing that?

htmldoug:
Encouraged by https://github.com/wiremock/wiremock/issues/1153#issuecomment-1023189606, I went ahead and submit the more aggressive version of this in #1790 as an alternative.

htmldoug:
Closing in favor of #1790.

PR #1784, opened 2022-01-25
Fix proxyUrlPrefixToRemove with ResponseTemplateTransformer  (#1783)

### LINKED ISSUE #1783: When using Proxying with the ResponseTemplateTransformer the setting proxyUrlPrefixToRemove is ignored
WireMock version: 2.32.0

# How to reproduce:

Use the following mapping:

```json
{
  "request": {
    "method": "ANY",
    "urlPattern": "/proxy-test/.*"
  },
  "response": {
    "proxyBaseUrl" : "https://localhost:8099",
    "proxyUrlPrefixToRemove" : "/proxy-test"
  }
}
```

Use the setting "--global-response-templating". 

When sending a request to '/proxy-test/foo'
Expected behaviour: a proxy request to https://localhost:8099/foo
Actual behaviour: a proxy request to https://localhost:8099/proxy-test/foo

andrena-eike-kohnert:
Fixes #1783


Delfic:
Hey 👋 
Wouldn't a better fix be to change the `ResponseDefinitionBuilder.like` method to include the "missing" fields of `ResponseDefinition`?
They're handled by `ProxyResponseDefinitionBuilder` but that one doesn't have a `like` method 🤔 

We had a similar issue and fixing the `like` seems to be a more broad fix.

tomakehurst:
Fixed in 9aad8789d2c470eb9e92df83494a2dcffe90ad06

PR #1790, opened 2022-01-29
perf: Avoid unnecessary intermediate `JsonNode`

htmldoug:
This is the more aggressive version of https://github.com/wiremock/wiremock/pull/1772 that eliminates _all four_ major stacks in the flame graph below--not just the highlighted ones.

Dealing in terms of `Body` as a first-class entity separates out concerns nicely. PR diff is LOC negative already, but we could also delete `ResponseDefinitionBuilder.with...Body(...)`, `ResponseDefinition.get...Body`, and `ResponseDefinition(..., bytes[] body, ...)` if you don't mind the breaking API changes. We could also deprecate them for a while, or just leave them as is. Let me know your preference.

I'm already enjoying the performance benefits of this change via `ResponseDefinition.class.getDeclaredField("body").setAccessible(true)` etc., but it'd be nice to switch back to public APIs.

ResponseDefinition getJsonBody

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2021 Thomas Akehurst
+ * Copyright (C) 2011-2022 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


tomakehurst:
This looks good.

I think deprecating the `with...Body()` methods in 2.x then removing in 3 is the best approach. If you're happy to do add the deprecations I'll merge this one.

htmldoug:
There are 100+ uses of the `.withBody()` variants. Shall I migrate those at the same time?

htmldoug:
I'd suggest merging as-is to start getting the benefits of the perf improvement. Minimizing the public API footprint is an opportunity, but not critical. All the existing methods still work properly.

tomakehurst:
Agree it's worth merging now and building on. Thanks for submitting this!

PR #1800, opened 2022-02-07
Fix null-valued headers

htmldoug:
Found a bug with null-valued headers introduced in https://github.com/wiremock/wiremock/pull/1791.


htmldoug:
Holding for now. I need to double check this. The behavior of `getHeader`/`header` returning `null`/`""` is fairly nuanced.

PR #1816, opened 2022-02-21
Archunit 0.23.0

timtebeek:
Closes #1815.

Method references are now picked up, leading to fewer false positives in detecting unused methods.
Required a minor code change to now use `JavaCodeUnitAccess`.

Updated store to only freeze the for now disabled unused methods test, as the unused public methods test now no longer has a false positive. We can reconsider whether that test needs to remain disabled, as there should no longer be false positives.

Ideally and eventually we can remove true unused methods through #1702; until then this minimal change allows us to pick up new releases of ArchUnit.


jamietanna:
I'll try and have a look today, but @tomakehurst will still need to do the honours of merging as I've not yet got commit access ☺

jamietanna:
I think this LGTM - happy for Tom to have another pair of eyes and merge when ready :+1:

timtebeek:
Discussed outside GitHub; this one is good to go it seems! :)

jamietanna (approved):

src/test/java/com/github/tomakehurst/wiremock/archunit/UnusedCodeTest.java
```diff
@@ -184,10 +131,63 @@ public void check(JavaMethod javaMethod, ConditionEvents events) {
...
+              .and(
+                  not(
+                      describe(
+                          "are not builders",
+                          input ->
+                              input.getParameterTypes().size() <= 1
+                                  && input.getOwner().tryGetField(input.getName()).isPresent())))
               .should(beReferencedMethod)
-              .as("should use all non public methods")
+              .as("should use all methods")
```
tomakehurst, line 166: @timtebeek this message change still implies that we're failing on any unused public methods. Am I misunderstanding?

timtebeek: That's merely a result of the git diff rendering weirdly in the presence of indentation changes.
The `nonPublicMethodsShouldNotBeUnusedFrozen` isn't really new; it was there before already.
But because the `freeze` was moved from that test to `methodsShouldNotBeUnused` the diff is off.

There's nothing new in terms of what's enforced in this MR; all as it was, with `methodsShouldNotBeUnused` disabled for now. The (now) empty freeze store for `nonPublicMethodsShouldNotBeUnusedFrozen` was repurposed to serve `methodsShouldNotBeUnused`. Hence any confusion; but nothing new is enforced just yet.

tomakehurst: OK, got it (I think!). I'll merge.

timtebeek: Thanks! Glad to see this merged :)

PR #1828, opened 2022-03-02
Reduce threshold for flakey test

jamietanna:
We're seeing this test failing quite a bit, in cases where the delay is
499ms instead of 500ms, and therefore failing.

To avoid this, we can instead reduce the threshold to >= 499ms, and give
us a bit of breathing room for us to remediate this longer-term.


tomakehurst:
LGTM

PR #1829, opened 2022-03-02
Lint: remove `static` locking of `GuavaBase64Encoder`

jamietanna:
As highlighted by SonarQube, this isn't the most ideal way of setting
this up.

Instead of restoring it, we can instead replace this with
`java.util.Base64`, as the Guava encoder isn't part of our public
interface, and we've got access to the core Java package.


tomakehurst:
We need to be a bit careful here. IIRC we use the Guava encoder because it's quite a lot faster than than the one from `java.util`. I'm pretty sure we'd be undoing an explicit performance optimisation, and likewise by not caching it after creation.

PR #1830, opened 2022-03-04
Fix like creating complete response definition

### LINKED ISSUE #1783: When using Proxying with the ResponseTemplateTransformer the setting proxyUrlPrefixToRemove is ignored

Delfic:
Fixes #1783.

Changed the test of `like` to use the main object instead of the builder so that it's more resilient. Added changes to fix the missing recently added properties.
Should fix the aforementioned issue and an issue I was having while trying to use `like`.


src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
```diff
@@ -66,7 +67,13 @@ public static ResponseDefinitionBuilder like(ResponseDefinition responseDefiniti
             ? Parameters.from(responseDefinition.getTransformerParameters())
             : Parameters.empty();
     builder.wasConfigured = responseDefinition.isFromConfiguredStub();
-    return builder;
+
+    ProxyResponseDefinitionBuilder proxyResponseDefinitionBuilder = new ProxyResponseDefinitionBuilder(builder);
```
tomakehurst, line 71: I think we should do this conditionally, in order to maintain the same behaviour when building from scratch, so e.g.

```java
if (builder.proxyBaseUrl != null) {
  // Construct the ProxyResponseDefinitionBuilder
}

// Otherwise keep the builder as-is and return it.
```

tomakehurst:
Fixed in 9aad8789d2c470eb9e92df83494a2dcffe90ad06

PR #1865, opened 2022-04-25
Don't create request entity when it was not present in original request

### LINKED ISSUE #1864: Wiremock 2.33.0+ - UnsupportedMediaType error on the back end for a GET request proxied by WireMock
I am using WireMock together with ChromeDriver - wiremock is instantiated and passed as a proxy to the ChromeDriver.

Here is how wiremock is instantiated:

        var wireMockServer = new WireMockServer(wireMockConfig().port(8888)
            .bindAddress("127.0.0.1")
            .enableBrowserProxying(true));
        wireMockServer.start();

On the chromedriver I just add the capabilities:

    proxy.proxyType=manual
    proxy.sslProxy=127.0.0.1:8888

In my back end application under test, there is one (and only) GET request that started failing with the new 2.33.0 (and 2.33.1) version of WireMock. The exception is "javax.ws.rs.NotSupportedException: HTTP 415 Unsupported Media Type".

I assume this happens because of this change in 2.33.0: 
- Proxy will now send a request body for any request method (i assume this is #1765 ?)

I don't think this is right default behavior. When I didn't have a request entity the proxy shouldn't just add it for no reason.

Would it be a decent solution would be to check if the content length is 0 before adding the entity?

ubeogesh:
Fixes #1864

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
@@ -101,7 +101,10 @@ public class ProxyResponseRenderer implements ResponseRenderer {
     HttpUriRequest httpRequest = getHttpRequestFor(responseDefinition);
     addRequestHeaders(httpRequest, responseDefinition);
 
-    httpRequest.setEntity(buildEntityFrom(responseDefinition.getOriginalRequest()));
+    Request originalRequest = responseDefinition.getOriginalRequest();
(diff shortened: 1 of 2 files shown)


tomakehurst:
Thanks for contributing this @ubeogesh. Please can you run `./gradlew spotlessApply` and commit the changes to fix the formatting so that the build passes?

ubeogesh:
Done. Sorry I have missed it originally.

tomakehurst:
Great stuff, thanks!

PR #1873, opened 2022-05-12
Create request entity if content-length is set (#1872)

### LINKED ISSUE #1872: Proxy of empty POST requests missing to content-lenght header
- WireMock version: 2.33.2
- Main-Method of WireMockServerRunner without any options

If a POST request with an empty payload is sent via a proxy mapping, the “content-length” header is lost. This causes the server to respond with status 411 (“The request must be chunked or have a content length”).

## Example:
Stub Mapping:
```
{
  "request": {
    "urlPathPattern": "/example/.*",
    "method": "ANY"
  },
  "response": {
    "proxyBaseUrl": "https://api.app.mobiliar-sit.ch"
  }
}
```

Request: `curl -X POST --location "http://localhost:8080/example/empty-post" -H "Content-Length: 0"`

## Analysis
There have been several changes which, in my opinion, lead to the behavior.

- With #1625 the  ContentType.DEFAULT_BINARY is set as default. (Will be ignored, if content-type Header is set)
  - https://github.com/wiremock/wiremock/blob/59e8c001b2b2839dbff98026592915cd9cc7ae16/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java#L219
- With #1765 the body of all http-methods is proxied
- This led to #1864 , where the fix #1865 actually removes the empty body and thus also the `content-length` header
-- Issue was "Unsupported Media Type", which usually comes from a “wrong” content-type header.

## Proposed solution
- Don't set a default ContentType at https://github.com/wiremock/wiremock/blob/6728eb79f440f561772421095d64ce3bda541fa9/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java#L232
  - Use the content-type from header of the original request or set it to `null`
- Create and add "HttpEntity", if it's an empty byte[] but content-length is set
  -  Alternative: Always create HttpEntity. Drawback of this solution is, that the `content-length` Header will be set for all requests. (including GET).

vonatzigenc:
fixes #1872

To enable valid POST requests without a body, an empty entity is required.
Otherwise, the proxy request will not include the header content-length.
Additionally, the content-type will only be set if it is already present
in the originalRequest.
This prevents a new, additional header from being set.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/ProxyResponseRenderer.java
@@ -102,7 +102,8 @@ public class ProxyResponseRenderer implements ResponseRenderer {
     addRequestHeaders(httpRequest, responseDefinition);
 
     Request originalRequest = responseDefinition.getOriginalRequest();
-    if (originalRequest.getBody() != null && originalRequest.getBody().length > 0) {
+    if ((originalRequest.getBody() != null && originalRequest.getBody().length > 0)
+        || originalRequest.containsHeader(CONTENT_LENGTH)) {
       httpRequest.setEntity(buildEntityFrom(originalRequest));
     }
     CloseableHttpClient client = buildClient(serveEvent.getRequest().isBrowserProxyRequest());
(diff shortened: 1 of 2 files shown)


Megmeehey (approved):

Megmeehey:
[x2 that]

Issue #1689, opened 2021-11-02
Response header does not reflect the correct version for HTTP2 requests


piyushgupta243:
# Issue

The response header does not reflect the correct version for HTTP2 requests. Even though, wiremock replies in h2 but the version in the response header shows "HTTP/1.1 200". This is caused by the hardcoded value in the toString() method of Response.java class

```
    @Override
    public String toString() {
        StringBuilder sb = new StringBuilder();
        **sb.append("HTTP/1.1 ")**.append(status).append("\n");
        sb.append(headers).append("\n");
        // no longer printing body
        return sb.toString();
    }
```

## Environment

- [ ] Which version of WireMock you're using - 
2.31.0 standalone

- [ ] How you're starting and configuring WireMock, including configuration or CLI the command line
Starting wiremock in standalone mode inside a docker container

## Steps to reproduce

1. Insert some request-response data in wiremock
2. Try to make a request with h2
3. See the response on the console. You get the response like below for the h2 requests

```
Response:
HTTP/1.1 200
Content-Type: [application/json]
Matched-Stub-Id: [fc793a71-3651-4c9b-bafd-9678c6265950]
```

sunilkumar065:
Is this fixed in 2.32.0? I tried and below is the reponse
```
curl --http2 -s -I -X GET http://localhost:8080/api/test
```

```
HTTP/1.1 101 Switching Protocols

HTTP/2 200 
matched-stub-id: 908d8b29-fdf2-4d13-9435-240d804d4f02
vary: Accept-Encoding, User-Agent
```

PR #1890, opened 2022-06-02
Bugfix: response header not reflecting correct version of http #1689

SaiPoojithaDubba:
link to the issue: #1689


tomakehurst (changes requested):
This is a good start and I think only one change is needed, which is to move `Response.setProtocol()` to `Response.Builder.protocol()` so that `Response` remains immutable.

SaiPoojithaDubba:
> This is a good start and I think only one change is needed, which is to move `Response.setProtocol()` to `Response.Builder.protocol()` so that `Response` remains immutable.

Change committed!

tomakehurst (approved):

tomakehurst:
Thanks!

PR #1898, opened 2022-06-08
Fix proxyUrlPrefixToRemove property not being applied 

Ross-H-Projects:
The proxyUrlPrefixToRemove  property wasn not being properly applied during runtime. It was not being persisted through the ResponseDefinitionBuilder as it was being stored as a (not static) local variable in a static class. 

Also it will now conditionally return builder of ProxyResponseDefinitionBuilder if proxyBaseUrl is set


tompahoward:
Fixes #1783

tomakehurst:
Thanks

tompahoward:
Thank you for merging @tomakehurst. Do you happen to know when the next release will be?

tomakehurst:
I'm going to attempt it tomorrow if I can find time to merge a couple of other fixes first.

PR #1927, opened 2022-07-24
Fixed #1897 - Wrong port used on Junit Jupiter nested classes

### LINKED ISSUE #1897: Wrong port used on Junit Jupiter nested classes
**Version** : `2.33.2`

I have noticed two behaviours that don't seem to work as expected regarding nested classes and wiremock server ports.

### 1. The server is not restarted on the first nested class, even if the port is not the same
The first issue can be reproduced using a simplified version of wiremock unit test class `JunitJupiterExtensionDeclarativeWithConfiguredNestedTest`:
```java
@WireMockTest(httpPort = 8765)
class JunitJupiterExtensionDeclarativeWithConfiguredNestedTest {
  @Test
  void runs_on_the_supplied_port(WireMockRuntimeInfo wmRuntimeInfo) {
    assertThat(wmRuntimeInfo.getHttpPort(), is(8765));
  }

  @Nested
  @WireMockTest(httpPort = 8766)
  class RunsOn8766 {
    @Test
    void runs_on_the_supplied_port(WireMockRuntimeInfo wmRuntimeInfo) {
      assertThat(wmRuntimeInfo.getHttpPort(), is(8766));
    }
  }
}
```
When running the test in `RunsOn8766` class the wiremock server is not restarted, therefore it is still running on port `8765` causing the error : 
```
java.lang.AssertionError: 
Expected: is <8766>
     but: was <8765>
```

### 2. The server is restarted on a random port on second nested class
The second issue can also be reproduced by changing a little bit this same test class and forcing an order on the tests : 
```java
@WireMockTest(httpPort = 8765)
@TestClassOrder(ClassOrderer.OrderAnnotation.class)
class JunitJupiterExtensionDeclarativeWithConfiguredNestedTest {
  @Test
  void runs_on_the_supplied_port(WireMockRuntimeInfo wmRuntimeInfo) {
    assertThat(wmRuntimeInfo.getHttpPort(), is(8765));
  }

  @Nested
  @Order(0)
  class RunsOnInheritedPortFirstTry {
    @Test
    void runs_on_the_supplied_port(WireMockRuntimeInfo wmRuntimeInfo) {
      assertThat(wmRuntimeInfo.getHttpPort(), is(8765));
    }
  }

  @Nested
  @Order(1)
  class RunsOnInheritedPortSecondTry {
    @Test
    void runs_on_the_supplied_port(WireMockRuntimeInfo wmRuntimeInfo) {
      assertThat(wmRuntimeInfo.getHttpPort(), is(8765));
    }
  }
}
```
When running the test in `RunsOnInheritedPortSecondTry` class the wiremock server on `8765` has already been stopped (at the end of `RunsOnInheritedPortFirstTry` class). But when it restarts, it will restart on a random port and won't be able to retrieve the annotation on the parent class, causing an error like this :  
```
java.lang.AssertionError: 
Expected: is <8765>
     but: was <55464>
```

I have tried to fix the code myself and open a PR but it can be tricky because at some point we might have a server already running on  `8765` and not stopped yet, then one test restarting a server on another port, and then a second test that needs to reuse the first server on `8765`. I didn't know how to retrieve that one and couldn't start a second server on the same port for obvious reasons.

uc4w6c:
Fixed a bug where incorrect ports are used in test cases where nested classes exist and each class has a different `@WireMockTest` annotation.

Even if there are multiple `@WireMockTest` annotations, the above problem occurs because only one instance of `WireMockExtension` is created.
This problem is solved by managing WireMock Server for each class using `ExtensionContext.Store`.

Issue: https://github.com/wiremock/wiremock/issues/1897

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Thomas Akehurst
+ * Copyright (C) 2021-2022 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,14 +21,17 @@ import com.github.tomakehurst.wiremock.core.Options;
 import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
 import com.github.tomakehurst.wiremock.http.JvmProxyConfigurer;
 import com.github.tomakehurst.wiremock.junit.DslWrapper;
+import java.util.NoSuchElementException;
 import java.util.Optional;
 import org.junit.jupiter.api.extension.*;
-import org.junit.platform.commons.support.AnnotationSupport;
+import org.junit.jupiter.api.extension.ExtensionContext.Namespace;
+import org.junit.jupiter.api.extension.ExtensionContext.Store;
+import org.junit.platform.commons.util.AnnotationUtils;
 
 /**
  * JUnit Jupiter extension that manages a WireMock server instance's lifecycle and configuration.
  *
- * See http://wiremock.org/docs/junit-jupiter/ for full documentation.
+ * <p>See http://wiremock.org/docs/junit-jupiter/ for full documentation.
  */
 public class WireMockExtension extends DslWrapper
     implements ParameterResolver,
@@ -49,6 +52,10 @@ public class WireMockExtension extends DslWrapper
 
   private Boolean proxyMode;
 
+  private static final String WIRE_MOCK_SERVER_KEY = "wireMockServer";
+  private static final String WIRE_MOCK_RUNTIME_INFO_KEY = "wireMockRuntimeInfo";
+  private static final String PROXY_MODE_KEY = "proxyMode";
+
   public WireMockExtension() {
     configureStaticDsl = true;
     failOnUnmatchedRequests = false;
@@ -57,10 +64,11 @@ public class WireMockExtension extends DslWrapper
   /**
    * Constructor intended for subclasses.
    *
-   * The parameter is a builder so that we can avoid a constructor explosion or
+   * <p>The parameter is a builder so that we can avoid a constructor explosion or
    * backwards-incompatible changes when new options are added.
    *
-   * @param builder a {@link com.github.tomakehurst.wiremock.junit5.WireMockExtension.Builder} instance holding the initialisation parameters for the extension.
+   * @param builder a {@link com.github.tomakehurst.wiremock.junit5.WireMockExtension.Builder}
+   *     instance holding the initialisation parameters for the extension.
    */
   protected WireMockExtension(Builder builder) {
     this.options = builder.options;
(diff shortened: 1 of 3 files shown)


tomakehurst:
I've tried running the test cases you've added here with a breakpoint set in `WireMockExtension` line 176 and it seems this might be a JUnit bug, as sometimes the extension context is correctly returned as the nested class, but sometimes it's the parent class (even when running a single nested test).

uc4w6c:
@tomakehurst 
Thank you.

Do your tests fail sometimes?
If the test always succeeds, it may be called from `@WireMockTest` attached to the parent class.

If `@WireMockTest` is given to both the parent class and the nested class, two WireMockServers will be started on the ports given to each annotation.
And when the test case method is executed, the WireMockServer created in the method's class is used.
(Line 289 of the beforeEach method gets the WireMockServer from the ExtensionContext.Store)

tomakehurst:
To clarify - the tests pass or fail deterministically. What I mean is that there are some nested tests whose extension is handed the nested class as the context by JUnit and some where it gets the parent. There isn't any obvious difference between these cases.

uc4w6c:
I've tested it several times and it all passes on my local.
Can I try running GitHub Actions a few times?

Is `JunitJupiterExtensionDeclarativeWithConfiguredNestedTest` an example test class that fails?

tomakehurst:
Sorry, I didn't explain that very clearly - I just pasted your test case into the project without making any changes to the production code so that I could observe what happens when it fails.

One nested test with a fixed port gets given the nested class as its context, whereas the other one gets the parent class, which is why the current production code doesn't always set the correct port.

What I can't understand is why JUnit sometimes passes the child and sometimes the parent as the context, and whether this is in fact a bug with JUnit.

uc4w6c:
Okay.

The context of the beforeAll method is given the parent class and nested class, and the context of the beforeEach method is given the nested class.

---

I tried below code.

```
@WireMockTest(httpPort = 8765)
public class NestedTest {
    @Nested
    @WireMockTest(httpPort = 8766)
    class RunsOn8768 {
        @Test
        void runs_on_the_supplied_port(WireMockRuntimeInfo wmRuntimeInfo) {
            assertThat(wmRuntimeInfo.getHttpPort(), is(8766));
        }
    }
}
```

1. beforeAll
Context is given a `parent class`.

context.testDescriptor = ClassTestDescriptor: [engine:junit-jupiter]/[class:com.github.tomakehurst.wiremock.junit5.NestedTest]

2. beforeAll
Context is given a `nested class`.

context.testDescriptor = NestedClassTestDescriptor: [engine:junit-jupiter]/[class:com.github.tomakehurst.wiremock.junit5.NestedTest]/[nested-class:RunsOn8768]

3. beforeEach
Context is given a `nested class`.

context.testDescriptor = TestMethodTestDescriptor: [engine:junit-jupiter]/[class:com.github.tomakehurst.wiremock.junit5.NestedTest]/[nested-class:RunsOn8768]/[method:runs_on_the_supplied_port(com.github.tomakehurst.wiremock.junit5.WireMockRuntimeInfo)]

I think this is the expected behavior of JUnit.

An instance of WireMockExtension is shared even if multiple @WireMockTest are attached.

> Usually, an extension is instantiated only once.
https://junit.org/junit5/docs/current/user-guide/#extensions-keeping-state

And the reason why the current production code fails is that once a wireMockServer is created with WireMockExtension, even if the port is different, it is reused without creating a new one.

uc4w6c:
If there is a problem with the code, I will fix it, so could you please review it?

PR #1965, opened 2022-09-16
Downgrade SLF4J to 1.7.x

pkoenig10:
Fixes https://github.com/wiremock/wiremock/issues/1964

Versions 1.8.x and 2.x are incompatible with older versions. The vast majority of Java libraries are still using 1.7.x.

This PR downgrades SLF4J to the latest 1.7.x and updates the dependabot configuration to prevent upgrades to the 1.8.x and 2.x versions.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,8 +1,12 @@
 version: 2
 updates:
-- package-ecosystem: gradle
-  directory: "/"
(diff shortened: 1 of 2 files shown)


tomakehurst:
Dropped the version and relocated in 1f43b2794ca301a8504851b90d23522193bec5b1

PR #1981, opened 2022-10-12
Allow subclasses of WireMockExtension to get the ExtensionContext

tsmock:
This is useful when subclasses want to use annotations to help configure them.

Prior to the addition of the upstream WireMockExtension, I had a WireMockExtension (same class name) which allowed users to set transformers via an annotation.

In other subclasses, I used the ExtensionContext to backup/restore state, such that if a test wanted to modify the settings, it would be restored to the defaults for that class.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
@@ -115,12 +115,37 @@ public class WireMockExtension extends DslWrapper
   protected void onBeforeAll(WireMockRuntimeInfo wireMockRuntimeInfo) {}
 
   /**
+   * To be overridden in subclasses in order to run code immediately after per-class WireMock setup.
+   *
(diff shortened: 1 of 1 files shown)


oleg-nenashev (approved):
Would it make sense to deprecate the old methods?
Also needs review from @tomakehurst , because it's likely to collide with the extension API

oleg-nenashev:
Needs spotless, will do it

tsmock:
> Would it make sense to deprecate the old methods? Also needs review from @tomakehurst , because it's likely to collide with the extension API

I didn't deprecate the old methods because it was part of the API, and keeping them around was trivial to code (they are the same method, minus a parameter).

jayantp07:
Hi team, is there anything pending here? Would really appreciate this addition.

tomakehurst:
We're going to try to get this into the 3.6.0 release.

dieppa:
Thanks for the contribution, @tsmock 

Merged!

PR #1986, opened 2022-10-14
Fixed #1684 - Expose remove stub mapping by ID

### LINKED ISSUE #1684: Expose remove stub mapping by ID through admin client (HttpAdminClient)
Would be great to allow to delete a stub mapping by ID using admin client.

Current behaviour of method `wireMock.removeStubMapping` on external server leads to post request with json body that in our case matches too many mappings (we have a problem with multiple mappings to the same url and body but this is separate issue). 

After checking existing implementation method `removeStubMapping` on `HttpAdminClient` is using `OldRemoveStubMappingTask`. There is also `RemoveStubMappingTask` that probably was designed to be used against `/__admin/mappings/{id}` but is not exposed AFAIK

Would be great to expose existing `RemoveStubMappingTask` and allow to remove mappings by ID to allow better management of stubs (in ideal world stubs are well defined and existing method is good enough but this is not the case for us)

Patrolfr:
Hi, found this [good first issue](https://github.com/wiremock/wiremock/issues/1684) during Hacktoberfest.   
The issue seems self-explanatory, I will be happy to hear any feedback.


tomakehurst:
Generally this looks great.

Please can you replace the two unit tests you've added with an acceptance test case (in `StubbingAcceptanceTest` is fine). For features that are mostly about wiring everything together correctly, these tests are far more useful.

Patrolfr:
Sure, replaced unit tests with acceptance ones.

Patrolfr:
@tomakehurst Do you mind checking it again? Would be great to have the first PR merged 😉

PR #1991, opened 2022-10-17
Issue 1989: Support Content-Encoding: deflate

mrsalt:
This change adds support for Content-Encoding: deflate, and fixes a bug so that if an unrecognized Content-Encoding type is encountered we pass it through instead of removing it and making recorded stubs unusable.

Prior to these changes, if a Content-Encoding header is present, we remove it.  However, we only decompress the body if the Content-Encoding is gzip. If we're going to remove the header, we should also decode the content, and vice-versa.

Also, added .gitattributes so that when spotlessApply is used on Windows line endings are not changed to CRLF.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text eol=lf
--- a/src/main/java/com/github/tomakehurst/wiremock/common/BodyDecompressor.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/BodyDecompressor.java
@@ -1,0 +1,66 @@
+/*
+ * Copyright (C) 2022 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.common;
+
+import static com.google.common.net.HttpHeaders.*;
+
+import com.github.tomakehurst.wiremock.http.CaseInsensitiveKey;
+import com.github.tomakehurst.wiremock.http.HttpHeader;
+import com.github.tomakehurst.wiremock.http.HttpHeaders;
+import com.github.tomakehurst.wiremock.http.Response;
+import com.google.common.base.Predicate;
+import com.google.common.collect.ImmutableList;
+import java.util.List;
+
+public class BodyDecompressor {
+
+  private static final List<CaseInsensitiveKey> EXCLUDED_HEADERS =
+      ImmutableList.of(
+          CaseInsensitiveKey.from(CONTENT_ENCODING),
+          CaseInsensitiveKey.from(CONTENT_LENGTH),
+          CaseInsensitiveKey.from(TRANSFER_ENCODING));
+  private final byte[] bytes;
+  private final Predicate<HttpHeader> predicate;
+
+  public BodyDecompressor(Response response) {
+    this(response.getBody(), response.getHeaders());
+  }
+
+  public BodyDecompressor(byte[] body, HttpHeaders headers) {
+    if (body != null && body.length > 0 && headers != null) {
+      HttpHeader contentEncoding = headers.getHeader(CONTENT_ENCODING);
+      for (Compression algorithm : Compression.values()) {
+        if (contentEncoding.containsValue(algorithm.contentEncodingValue)) {
+          // Once we decompress the body, we need to remove these headers as they are no longer
+          // correct.
+          bytes = algorithm.decompress(body);
+          predicate = (header) -> !EXCLUDED_HEADERS.contains(header.caseInsensitiveKey());
+          return;
+        }
+      }
+    }
+    predicate = (header) -> true;
+    bytes = body;
+  }
+
+  public byte[] getBytes() {
+    return bytes;
+  }
+
+  public boolean shouldRetain(HttpHeader header) {
+    return predicate.test(header);
+  }
+}
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Compression.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Compression.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright (C) 2022 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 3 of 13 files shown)


src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
```diff
@@ -155,9 +155,11 @@ private void writeToMappingAndBodyFile(
 
     ResponseDefinitionBuilder responseDefinitionBuilder =
         responseDefinition().withStatus(response.getStatus()).withBodyFile(bodyFileName);
-    if (response.getHeaders().size() > 0) {
-      responseDefinitionBuilder.withHeaders(
-          withoutContentEncodingAndContentLength(response.getHeaders()));
+
+    if (response.getHeaders() != null) {
```
oleg-nenashev, line 159: I would keep the size check here too

src/test/java/com/github/tomakehurst/wiremock/recording/LoggedResponseDefinitionTransformerTest.java
```diff
@@ -69,22 +75,26 @@ public void applyWithBinaryBody() {
 
   @Test
   public void preservesHeadersExceptThoseSpecificallyExcluded() {
```
oleg-nenashev, line 77: Would make sense to duplicate the text to have both gzip and deflate

.gitattributes
```diff
@@ -0,0 +1 @@
+* text eol=lf
```
oleg-nenashev, line 1: I would handle it separately

src/main/java/com/github/tomakehurst/wiremock/common/BodyDecompressor.java
```diff
@@ -0,0 +1,66 @@
+/*
+ * Copyright (C) 2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2023 Thomas Akehurst
```

src/main/java/com/github/tomakehurst/wiremock/common/Compression.java
```diff
@@ -0,0 +1,47 @@
+/*
+ * Copyright (C) 2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2023 Thomas Akehurst
```

src/main/java/com/github/tomakehurst/wiremock/common/CompressionAlgorithm.java
```diff
@@ -0,0 +1,88 @@
+/*
+ * Copyright (C) 2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2023 Thomas Akehurst
```

src/main/java/com/github/tomakehurst/wiremock/common/Gzip.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2021 Thomas Akehurst
+ * Copyright (C) 2015-2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2015-2023 Thomas Akehurst
```

src/main/java/com/github/tomakehurst/wiremock/common/Zlib.java
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright (C) 2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2023 Thomas Akehurst
```

src/main/java/com/github/tomakehurst/wiremock/recording/LoggedResponseDefinitionTransformer.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2017-2021 Thomas Akehurst
+ * Copyright (C) 2017-2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2017-2023 Thomas Akehurst
```

src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2021 Thomas Akehurst
+ * Copyright (C) 2011-2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2011-2023 Thomas Akehurst
```

src/test/java/com/github/tomakehurst/wiremock/GzipAcceptanceTest.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2021 Thomas Akehurst
+ * Copyright (C) 2015-2022 Thomas Akehurst
```
oleg-nenashev, line 2:
```suggestion
 * Copyright (C) 2015-2023 Thomas Akehurst
```

oleg-nenashev (approved):
From what I see in the code, GZIP logic would benefit from more refactoring and encapsulation in the future, but it is already a good generalization step, and a good feature. +1 from me, and will try to get pipeline running by the headers date patch

.gitattributes
mrsalt: Fair enough

src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorder.java
mrsalt: My memory is fuzzy, but I think there's no harm in preserving the size check too.  I must have changed this as either my code changes made it possible for this to be null or I realized it was possible to be null even without my changes.

.gitattributes
mrsalt: I remember now why I added this.  If you're building on Windows, and spotlessApply, line endings are converted to CRLF for ALL files.  Definitely not what you want.

I can revert this and just leave it as a local file, but this is definitely a useful change for any other developer on Windows.

oleg-nenashev: Indeed. Normally there is a global Git config (`core.autocrlf`) that should take care of that. I do not mind if this file is merged though

https://docs.github.com/en/get-started/getting-started-with-git/configuring-git-to-handle-line-endings

oleg-nenashev:
@mrsalt Are you still interested in this PR? There is another one from @craftsman228 in the same area, and IIUC the quick discussion with @tomakehurst the change would need some refactoring. Would you be interested to work together in the coming weeks? Or would you be fine if we take the lead and refactor this pull request to get it  over the line? Thanks for your time!

mrsalt:
> @mrsalt Are you still interested in this PR? There is another one from @craftsman228 in the same area, and IIUC the quick discussion with @tomakehurst the change would need some refactoring. Would you be interested to work together in the coming weeks? Or would you be fine if we take the lead and refactor this pull request to get it over the line? Thanks for your time!

@oleg-nenashev I apologize for being slow to reply.  I'm not sure what specifically needs to be done, so it sounds better for you all to take the lead.  I'd like to review what you do though.

oleg-nenashev:
@mrsalt ack, thanks! I think we still need one implementation to be landed

PR #1993, opened 2022-10-18
removes maven dir in output shadow jar

krageon:
I think this is okay as I cannot find a reference to it either in the code or in the dependencies of any of the build stages (transitive or otherwise).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -294,8 +294,7 @@ shadowJar {
 
   mergeServiceFiles()
 
-  exclude 'META-INF/maven/javax.servlet/**'
-  exclude 'META-INF/maven/org.apache.commons/commons-text/**'
+  exclude 'META-INF/maven/**'
   exclude 'module-info.class'
   exclude 'handlebars-*.js'
 }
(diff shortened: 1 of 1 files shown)


krageon:
This fixes https://github.com/wiremock/wiremock/issues/1990 . It's been a few years since I did java things so my tooling fu is admittedly a little weak. If I missed something I'd like to figure out what that is. Thanks :)

tomakehurst:
I'm wondering if we can just cut the whole `META-INF/maven` directory. I'm not sure it's serving any purpose, even for the dependencies that are actually present.

krageon:
This was my initial resolution, but I figured I'd leave the changes to the minimum required to get things resolved. I'm perfectly willing to remove the whole thing, so I will do that after I eat. An hour or so.

krageon:
@tomakehurst all done chief

krageon:
I've done a little test with a simple mapping 

```
{
    "request": {
        "method": "GET",
        "url": "/simple"
    },
    "response": {
        "status": 200,
        "body": "Hello world!",
        "headers": {
            "Content-Type": "text/plain"
        }
    }
}
```

Which still works

Issue #1679, opened 2021-10-26
KeyStoreSource doesn't use provided type


kingsfleet:
We are using to wiremock-standalone-8 version 2.31 to create a wire mock instance locally and found that we needed to override the keystore type but that the standard KeyStoreSource didn't use the provided type:

    public KeyStore load() {
        InputStream instream = null;
        try {
            KeyStore trustStore = KeyStore.getInstance(KeyStore.getDefaultType());
            instream = createInputStream();
            trustStore.load(instream, keyStorePassword);
            return trustStore;
        } catch (Exception e) {
            return throwUnchecked(e, KeyStore.class);
        } finally {
            if (instream != null) {
                try {
                    instream.close();
                } catch (IOException ioe) {
                    throwUnchecked(ioe);
                }
            }
        }
    }

The line should be:

KeyStore trustStore = KeyStore.getInstance(keyStoreType);

This is a problem when working on a FIPS system because the default keystrokes have to be a particular BouncyCastle type.

PR #2003, opened 2022-10-26
Fix of #1679 - Respect keyStoreType when loading keystores.

kaarefc:
Use the provided keyStoreType when loading keystores. 
Note that documentation in https://wiremock.org/docs/https/ doesn't actually work - keystores are always attempted to be loaded with the default keystore type. 
This seems to be a regression. The type is respected in the Java 7 version of wiremock 2.27.2.


kaarefc:
Please let me know if there is anything further that should be done to this pull request.

tomakehurst:
Thanks for this, sorry it's taken a while to merge!

Issue #1892, opened 2022-06-02
Feature: negate condition in verify()


ashakirin:
## Feature requests

Provide an option to negate verification, like:

```
verify(getRequestedFor("/my_url"))
.withHeader("X-MyHeader").not(containing("value1"))
```

SatyamAK:
Hi, I am trying to understand what not will return, is it part of request pattern builder ? From my understanding so far not is not implemented at all, do we need to make not function in RequestPatternBuilder?

manishaag7:
@tomakehurst

exijn:
Hi, I'd like to work on this feature.

Would a generic negating matcher in the verification DSL be the preferred direction, or do you have a specific API shape in mind? I can prepare a small proposal with tests before implementing it.

Thanks.

PR #2006, opened 2022-10-31
New Feature: not() in verify #1892

SatyamAK:
Implemented a not method in WireMock.java which takes StringPatternValue parameter and returns the opposite of the result. 
For example, verify(getRequestedFor(urlEqualTo("/header/not")).withHeader("X-Thing", not(containing("Four"))));

The functionality has been properly tested, unit and acceptance tests are passing.

New File Added: NotPattern
File modified: StringValuePatternJsonDeserializer

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/NotPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/NotPattern.java
@@ -1,18 +1,20 @@
 package com.github.tomakehurst.wiremock.matching;
 
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
 
 public class NotPattern extends StringValuePattern{
 
     private StringValuePattern unexpectedPattern;
 
-    public NotPattern(@JsonProperty("not") StringValuePattern unexpectedPattern){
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/tomakehurst/wiremock/matching/NotPattern.java
```diff
@@ -0,0 +1,38 @@
...
+    private MatchResult invert(final MatchResult matchResult){
+        return new MatchResult() {
+            @Override
+            public boolean isExactMatch() {
+                return !matchResult.isExactMatch();
+            }
+
+            @Override
+            public double getDistance() {
+                return 1.0 - matchResult.getDistance();
```
tomakehurst, line 34: I think this could result in the situation where `isExactMatch()` returns true but the distance is still > 0.

The logic here should probably be:
if (exact match)
  distance = 0;
else
  distance = 1.0

tomakehurst (changes requested):

SatyamAK (commented):
Hi @tomakehurst we have made the changes according to your suggestions.

src/main/java/com/github/tomakehurst/wiremock/matching/NotPattern.java
SatyamAK: Hello @tomakehurst we have made changes, sorry for commenting this twice, I commented the previous comment in overall thread, but I felt the comment belonged to this request thread.

tomakehurst (approved):

PR #2019, opened 2022-11-16
Fixed clock skew problems with dynamically created certificates

DrMetallius:
There is an issue when using dynamically generated certificates for TLS connections to Wiremock regarding the certificate validity start time. The certificates are generated with validity start time being the same instant as the request arrives. If Wiremock runs at the same machine as the client, the problem doesn't manifest itself as they are both using the same clock. However, if the client is somewhere else and its time lags behind the Wiremock machine's time, even if it's an infinitesimal difference, the certificates won't be marked as valid.

To fix the clock skew problem, the certificates are now generated using a day before the present moment as the start time. This avoids the problem.


tomakehurst:
@Mahoney WDYT?

Mahoney (approved):
Seems reasonable; I did some research, there doesn't seem to be anything intrinsically dangerous about setting validity to the past.

Issue #2024, opened 2022-11-21
New feature: Make a callback before responding to the request (preServeAction)


mgorb:
## Feature

In some cases there is a need to make a request before serving a response to the client. Since today have have a webhook called **postServeAction** we can reuse this logic and add support for callbacks that must be made before responding to the request.
Example usage:
```json
{
    "request": {
        "url": "/make-payment",
        "method": "GET"
    },
    "response": {
        "status": 302,
        "headers": {
            "Location": "https://www.example.com/successful-payment"
        }
    },
    "preServeActions": [
        {
            "name": "webhook",
            "parameters": {
                "method": "POST",
                "url": "https://www.example.com/payment",
                "headers": {
                    "Content-Type": "application/json"
                },
                "body": "{ \"paymentStatus\": \"DONE\"}"
            }
        }
    ]
}
```

PR #2025, opened 2022-11-21
#2024 Add support for preServeAction

mgorb:
Solves #2024


tomakehurst:
We've now got a `ServeEventListener` extension point that can be triggered before matching, after matching or after completion.

PR #2035, opened 2022-11-29
Skips NPE when a multipart could not be parsed

sickmartian:
Recording the body instead.

This can happen when content of the multipart is not a file, so there is no header with content disposition giving it a filename. E.g. an OData $batch request.

Wiremock is currently failing the request, giving a 500 back because of the NPE, even when proxying when we got proper data back.

**edit** went a bit further on the library and saw what was going on, I assumed the lib I was using bad generating the request wrong and hence couldn't be parsed, but really wiremock assumes is going to be files whose parsing will work, not a bad assumption but it doesn't fit certain cases and the NPE is bad.


sickmartian:
Might be related to https://github.com/wiremock/wiremock/issues/702 although the description says valid json or xml, this might be a more general case.

tomakehurst:
Thanks!

Issue #2034, opened 2022-11-29
More missing equals and hashCode Methods


jnt0r:
# Issue guidelines

I already opened a related issue https://github.com/wiremock/wiremock/issues/1908 for the equals Method in MultipartValuePattern. Now I struggled with some more classes than cannot be compared because they use the default equals Implementation with the object id. 

For me, it is a best practice to always implement equals methods to be able to compare objects based on their values and not on their instance.

## Feature requests

I would like to add equals and hashCode method to every class that has to do with request/response matching. This will make it easier to test the classes correctly and allow users to implement new use cases.

jnt0r:
Especially I am missing equals for BinaryEqualTo Matcher.

jnt0r:
Hi @tomakehurst, what do you think about my feature request?

PR #2037, opened 2022-12-09
Adding missing equals Methods and Tests

jnt0r:
PR for #2034


jnt0r:
Hi @tomakehurst, any feedback on this? Would be nice to have these features ;)

tomakehurst:
I'm curious as to why you feel these are required on these classes. Can you share some examples of where the lack of these is a problem?

oleg-nenashev:
@tomakehurst  Without these methods, using hash-based, sets or sorted collections (or just `sort()`) may lead to undesired results.  I can imagine that someone uses WireMock as an API and wants to organize data before passing the parameters (e.g. for UX purposes). But I am not sure whether it is the use-case @jnt0r has in mind

jnt0r:
I'm building a layer above WireMock to dynamically create Mappings at runtime via proxyMappings. Every proxied request is recorded and then transformed to the format and checks we need and dynamically saved as a mapping. To detect if a mapping already exists, I'm using equals method. Without overriding these methods, you cannot compare mappings and parts of them.

tomakehurst:
@oleg-nenashev I understand in general what the impact of not including these methods is, but when I see people asking about them in internal classes where they're not needed for WireMock to work correctly I worry we're encouraging coupling to implementation details.

Having said that I think @jnt0r has a reasonable use case in mind here so I think we should merge this.

One comment about the change - I notice there are no test assertions for the `hashCode()` implementations - could we add these too?

jnt0r:
Alright, thanks @tomakehurst.

I updated the PR and added the tests for the hashCode methods.

PR #2054, opened 2023-01-02
Allow fixing class loader in ClasspathFileSource

derari:
I sometimes experience problems with ClassPathFileSource when I am using a dependency with Wiremock files and the web framework changes the context class loader to do its own magic.

This small change allows to pin the class loader of a ClassPathFileSource. Existing behavior is not modified.


tomakehurst:
Thanks!

Issue #1909, opened 2022-06-28
Support disabling of browser proxy pass-through


tomakehurst:
Currently when browser (forward) proxying is enabled, unmatched requests will be passed through to the target indicated by the originating request.

When purely using this mode for stub serving i.e. playback this isn't always desirable. Although it's possible to set a default stub in order to prevent this, we then lose the diff report when a request isn't matched.

We should therefore add a toggle to enable/disable the passthrough. This should probably be possible (or overridable at least) via the global settings feature so that a record/playback workflow can be completed without needing to restart the instance.

kapishmalik:
@tomakehurst can I take this up?

tomakehurst:
@kapishmalik yes, please do!

kapishmalik:
@tomakehurst cool, thanks. You can assign this issue to me.

tomakehurst:
One design comment - it would be useful for this to be switchable at runtime without restarting. If we allow it to be controlled/overridden by a flag in the global settings object, then it could be toggled on/off via an API call.

kapishmalik:
@tomakehurst just want to check one thing that if passthrough is disabled then what response it should return (404 response  or dummy response)?

kapishmalik:
I took the assumption and raised the PR for the same.

tomakehurst:
> @tomakehurst just want to check one thing that if passthrough is disabled with browser caching then what response it should return (404 response or dummy response)?

I'm not sure browser caching should have any bearing on the behaviour of the pass-through. Can you explain the situation you have in mind in a bit more detail?

kapishmalik:
@tomakehurst we can close this issue. It is done with documentation.

tomakehurst:
Closed in 6bb45ebdc786937896281a711e8bac377235bf03

PR #2087, opened 2023-02-27
Support disabling of browser proxy pass-through

kapishmalik:
resolves => #1909 

Added flag in global settings to control proxy pass-through. By default, it will be true and allows the request to go to the original URL. If flag will be false, it will return nearest matching stubbed request.

@tomakehurst  you can review this PR.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
@@ -134,7 +134,7 @@ public class WireMockConfiguration implements Options {
   private boolean stubCorsEnabled = false;
   private boolean disableStrictHttpHeaders;
 
-  private Boolean proxyPassThrough;
+  private boolean proxyPassThrough = true;
 
   private Limit responseBodySizeLimit = UNLIMITED;
 
@@ -156,7 +156,7 @@ public class WireMockConfiguration implements Options {
     return wireMockConfig();
   }
 
-  public WireMockConfiguration proxyPassThrough(Boolean proxyPassThrough) {
+  public WireMockConfiguration proxyPassThrough(boolean proxyPassThrough) {
     this.proxyPassThrough = proxyPassThrough;
     return this;
   }
@@ -545,8 +545,7 @@ public class WireMockConfiguration implements Options {
   @Override
   public Stores getStores() {
     if (stores == null) {
-      stores = proxyPassThrough != null
-          ? new DefaultStores(filesRoot, proxyPassThrough) : new DefaultStores(filesRoot);
+      stores = new DefaultStores(filesRoot, proxyPassThrough);
     }
 
     return stores;
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
@@ -25,7 +25,19 @@ import static com.google.common.collect.Maps.newLinkedHashMap;
 import static java.util.Arrays.asList;
 import static java.util.Collections.emptyList;
 
-import com.github.tomakehurst.wiremock.common.*;
+import com.github.tomakehurst.wiremock.common.AsynchronousResponseSettings;
+import com.github.tomakehurst.wiremock.common.BrowserProxySettings;
+import com.github.tomakehurst.wiremock.common.ClasspathFileSource;
+import com.github.tomakehurst.wiremock.common.DataTruncationSettings;
+import com.github.tomakehurst.wiremock.common.FileSource;
+import com.github.tomakehurst.wiremock.common.HttpsSettings;
+import com.github.tomakehurst.wiremock.common.JettySettings;
+import com.github.tomakehurst.wiremock.common.Limit;
+import com.github.tomakehurst.wiremock.common.NetworkAddressRules;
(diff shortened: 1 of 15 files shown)


kapishmalik:
@tomakehurst I have added acceptance tests.

tomakehurst:
Again, looking great.

One thing I meant to mention (apologies for not remembering before) - please can you also add a flag to `Options` for this, so that it's also possible to set it via the CLI and on programmatic startup?

The value from this can be used to set the default in the global settings I think.

kapishmalik:
@tomakehurst I have added flag pass proxy pass through as option via the CLI and on programmatic startup.

src/main/java/com/github/tomakehurst/wiremock/global/GlobalSettings.java
```diff
@@ -26,21 +26,21 @@ public class GlobalSettings {
   private final DelayDistribution delayDistribution;
   private final Parameters extended;
 
+  private final Boolean proxyPassThrough;
```
tomakehurst, line 29: I think this should be the primitive `boolean` here and elsewhere it's used as I don't think there's any reason for it to be null, and it's consistent with other flags in the configuration.

kapishmalik: @tomakehurst incorporated this change

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -122,6 +134,8 @@ public class WireMockConfiguration implements Options {
   private boolean stubCorsEnabled = false;
   private boolean disableStrictHttpHeaders;
 
+  private Boolean proxyPassThrough;
```
tomakehurst, line 137: Can we change this to `boolean`?

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -142,6 +156,11 @@ public static WireMockConfiguration options() {
     return wireMockConfig();
   }
 
+  public WireMockConfiguration proxyPassThrough(Boolean proxyPassThrough) {
```
tomakehurst, line 159: Can we change this to `boolean`?

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
kapishmalik: If I change this to primitive type then it will always take the default value as false so I feel here it is not required. Let me know if you feel otherwise.

src/main/java/com/github/tomakehurst/wiremock/global/GlobalSettings.java
```diff
@@ -78,10 +86,13 @@ public int hashCode() {
   }
 
   public static class Builder {
+
     private Integer fixedDelay;
     private DelayDistribution delayDistribution;
     private Parameters extended;
 
+    private Boolean proxyPassThrough;
```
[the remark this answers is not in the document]
kapishmalik: yea sure. Missed.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
kapishmalik: Won't be required if I won't incorporate above comment.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
kapishmalik: I am assigning this value as true as I can see done for others and incorporating this change as well.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
kapishmalik: incorporated

kapishmalik:
@tomakehurst I have incorporated all comments. Could you please check this PR.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -526,7 +545,7 @@ public ProxySettings proxyVia() {
   @Override
   public Stores getStores() {
     if (stores == null) {
-      stores = new DefaultStores(filesRoot);
+      stores = new DefaultStores(filesRoot, proxyPassThrough);
```
tomakehurst, line 548: I feel like `proxyPassThrough` shouldn't be passed in here.

We want the `proxyPassThrough` default value to be used regardless of whether we're using the default stores implementation or one supplied by the user.

kapishmalik: incorporated

kapishmalik:
@tomakehurst I have incorporated review comment changes. You can review and let me know.

tomakehurst:
OK, I think this looks good so I'll merge it assuming the build is OK.

One last request - please could you open a PR against the [docs repo](https://github.com/wiremock/wiremock.org-sources) to add the relevant documentation for this?
I think the standalone (for the CLI bits) and the proxying pages are the ones to aim for.

kapishmalik:
Sure @tomakehurst I ll add another PR for adding relevant documentation.

PR #2048, opened 2022-12-19
Refactoring

ullenius:
Minor refactoring of code smells:

* Removing unused imports
* Remove unnecessary method calls
* Migrate to Java 8 methods
* Use Java 8 type inference
* Remove redundant `throws` clause

PR #2116, opened 2023-03-26
Fix broken hashcode

ullenius:
Fixes broken `hashCode`-implementation in `Body`


ullenius:
Split from #2048

oleg-nenashev (approved):
LGTM

PR #2123, opened 2023-03-31
Fix JUnit5 extension bug when using both declarative + programmatic simultaneously 

### LINKED ISSUE #2122: Junit5 port configuration gets set incorrectly when mixing declarative and programmatic
#### Context
I was updating a Junit 4 project to using Junit 5 and rewriting some tests using `WireMockClassRule` to use `WireMockExtension` instead. Following the https://wiremock.org/docs/junit-jupiter/ doc, I first ended up setting up the `@WireMockTest` annotation at the top of my class, following the declarative approach from this guide. I then tried to use the `@RegisterExtension` approach with static members of my test class. 
Doing this I forgot to remove the `@WireMockTest` annotation. So while I did not end up actually using the declarative instance in my tests, I did keep the annotation.

#### Issue
The issue that occurred is as follows: 

My programmatically instantiated WireMockExtension instances, static members of my JUnit5 test class, were ignoring my configured port, and were using a random one instead. So the situation was the following:

```java
@WireMockTest
public class MyTest {

  @RegisterExtension
  static WireMockExtension wm = WireMockExtension.newInstance()
      .options(wireMockConfig().port(44444))
      .build();

  @Test
  public void mytest() {
    final int port = wm.getPort();
    assertThat(port, is(44444)); // this fails! The port is random each run.
  }
}
```

I investigated this issue and observed that when you don't use static members you don't have this issue, and the port gets set correctly. Debugging this further I realised that there are two separate instances of `WireMockExtension` and finally I realised that `@WireMockTest` is creating its own instance that I was not using.

The unexpected part however is that the code ends up behaving strangely in the static case. One would expect that when combining static members and the declarative style that the declarative instance would use the declaratively defined options (port, etc... Like `@WireMockTest(httpsEnabled = true, httpsPort = 8443)`) and the programmatically declared instances would use exactly the supplied options for every instance (albeit combining both approaches _intentionally_ would be a strange strategy). 

I discovered that there is a bug that essentially makes the static members use the declarative configuration defaults when the annotation is present, instead of using the explicit provided options. 

### A fix

I will be opening a PR to fix the above.  By tracking which instance of the JUnit5 WireMockExtension is a "declarative" instance and which is a "non-declarative" instance, it will ignore the declarative configuration when it is not meant to be declarative (particularly static members are not).  

#### Some more thoughts

As mentioned prior, I do think that it is a bit of an odd strategy to use both simultaneously, so I'm not sure if maybe that should produce a warning of some kind or if it is best to just leave it up to the programmer to know what they are doing there. 🤷‍♂️ 

Regardless, if it is physically possible to do both things at once then the issue as described above seems like an unexpected behaviour.

#### Bug report details

**WireMock Version:** both `2.32.0` and `3.0.0-beta-7` (latest)

parawanderer:
This fixes #2122, please see the issue described there for details.

This PR makes it so that the options of httpPort/httpsPort/httpsEnabled/proxyMode and anything else that might be configured through the `@WireMockTest` attribute for a declarative setup in JUnit5 does not override programmatically configured options.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Thomas Akehurst
+ * Copyright (C) 2021-2023 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -28,7 +28,7 @@ import org.junit.platform.commons.support.AnnotationSupport;
 /**
  * JUnit Jupiter extension that manages a WireMock server instance's lifecycle and configuration.
  *
- * See http://wiremock.org/docs/junit-jupiter/ for full documentation.
+ * <p>See http://wiremock.org/docs/junit-jupiter/ for full documentation.
  */
 public class WireMockExtension extends DslWrapper
     implements ParameterResolver,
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
+1 from me. With explicit configuration in test, there should be either the programmatic way taking a priority (this PR) or an immediate error. I prefer the approach in this PR too

tomakehurst:
Please can we fix the formatting issues so that the tests can run?

tomakehurst:
Thanks, and sorry for the wait!

Issue #570, opened 2017-01-12
JsonSchema Matcher for RequestBody Verification


jbspeakr:
As the current RequestPatternBuilder.withRequestBody(...) is not properly customizable from the outside using Predicates (see #101) or via custom StringValuePattern implementations (as they are registered within the StringValuePatternJsonDeserializer), I'd like to open this feature request:

It would be great to be possible to match a RequestBody against an expected JsonSchema, e.g. using https://github.com/daveclayton/json-schema-validator

**Example:** 
```
wireMockServer.verify(anyRequestedFor(urlEqualTo("/verify/this"))
        .withRequestBody(matchingJsonSchema(MY_SCHMEMA_DEFINITION)));
```

Also (to avoid feature requests like this in the future), it would be great if providing custom matchers would be possible/ easier.

tomakehurst:
Custom matchers are both possible and easy :-)
See: http://wiremock.org/docs/extending-wiremock/#custom-request-matchers

Validation of requests against Swagger and RAML (which both use JSON Schema) is on the way.

jbspeakr:
ah, I somehow missed the RequestPatternBuilder.forCustomMatcher()... thanks for that!
also I am looking forward to the new features ;)

tomakehurst:
Revisiting this after a number of conversations with users, and given the increasing prevalence of OpenAPI.

tomakehurst:
Added in bb62fc9a6b78cca9a7e3aa3c5e648b90de64e58a

PR #2134, opened 2023-04-13
JSON schema matcher

tomakehurst:
Adds support for matching request content (primarily the body) against a JSON schema document, per #570


build.gradle
```diff
@@ -94,6 +94,8 @@ dependencies {
   }
   api "commons-io:commons-io:2.11.0"
 
+  api 'com.networknt:json-schema-validator:1.0.78'
```
oleg-nenashev, line 97: Looks to be the best possible choice among available validators. I used https://github.com/everit-org/json-schema before but there are strange deprecation events that happened this week

src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
```diff
@@ -33,34 +33,7 @@
...
-import com.github.tomakehurst.wiremock.matching.NotPattern;
-import com.github.tomakehurst.wiremock.matching.RegexPattern;
-import com.github.tomakehurst.wiremock.matching.RequestPattern;
-import com.github.tomakehurst.wiremock.matching.RequestPatternBuilder;
-import com.github.tomakehurst.wiremock.matching.StringValuePattern;
-import com.github.tomakehurst.wiremock.matching.UrlPathPattern;
-import com.github.tomakehurst.wiremock.matching.UrlPathTemplatePattern;
-import com.github.tomakehurst.wiremock.matching.UrlPattern;
-import com.github.tomakehurst.wiremock.matching.ValueMatcher;
+import com.github.tomakehurst.wiremock.matching.*;
```
oleg-nenashev, line 36: 🔴 It actually violates Google Java Style Guide: https://google.github.io/styleguide/javaguide.html#s3.3-import-statements

src/test/java/com/github/tomakehurst/wiremock/verification/diff/PlainTextDiffRendererTest.java
```diff
@@ -15,22 +15,7 @@
...
-import static com.github.tomakehurst.wiremock.client.WireMock.including;
-import static com.github.tomakehurst.wiremock.client.WireMock.matching;
-import static com.github.tomakehurst.wiremock.client.WireMock.matchingJsonPath;
-import static com.github.tomakehurst.wiremock.client.WireMock.matchingXPath;
-import static com.github.tomakehurst.wiremock.client.WireMock.post;
-import static com.github.tomakehurst.wiremock.client.WireMock.requestMatching;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlMatching;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlPathEqualTo;
-import static com.github.tomakehurst.wiremock.client.WireMock.urlPathTemplate;
+import static com.github.tomakehurst.wiremock.client.WireMock.*;
```
oleg-nenashev, line 18: Codestyle

src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
```diff
@@ -1047,4 +1030,31 @@ public GlobalSettings getGlobalSettings() {
   public static GlobalSettings getSettings() {
     return defaultInstance.get().getGlobalSettings();
   }
+
+  public enum JsonSchemaVersion {
```
oleg-nenashev, line 1034: Should it be a `common.Json` nested class? It is a general-purpose class, and it looks a little odd here

oleg-nenashev (commented):
Looks good overall. Not blocking by codestyle, not sure what are the actual codestyle practices here

build.gradle
tomakehurst: Agree this looks like the best one.

src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
tomakehurst: It's here because it's used in the DSL, and `WireMock` tries to be the entrypoint for that.

Perhaps this is something to consider more deeply in a 4.0 refactor rather than worrying too much now.

PR #2136, opened 2023-04-14
Add support filename template

craftsman228:
In scope of this PR implemented:
 - Support filename template(via `CommandLineOptions` and `WireMockConfiguration`;
 - Removed `SafeFilename` util class. Replaced with new `FilenameMaker` class which based on filename template;
 - Small code refactoring;


oleg-nenashev (commented):
Is there a related GitHub issue? Would be nice to have it or some description in the pull request as https://github.com/wiremock/wiremock/blob/master/CONTRIBUTING.md#before-opening-a-pr recommends.

Bonus points for a documentation update too.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -597,6 +602,13 @@ public String bindAddress() {
     return bindAddress;
   }
 
+  @Override
+  public FilenameMaker getFilenameTemplate() {
```
tomakehurst, line 606: The name of this method and what it actually returns are not consistent with each other. It might make sense to return the template string and initialise the filename maker somewhere common, rather than doing it inside the `Options` impls.

src/main/java/com/github/tomakehurst/wiremock/standalone/CommandLineOptions.java
```diff
@@ -396,6 +400,9 @@ public CommandLineOptions(String... args) {
       stores.getSettingsStore().set(newSettings);
     }
 
+    TemplateEngine templateEngine =
```
tomakehurst, line 403: Thinking we'd be better off initialising this inside `FilenameMaker` and passing the template as a string constructor parameter.

src/main/java/com/github/tomakehurst/wiremock/common/filemaker/FilenameMaker.java
```diff
@@ -0,0 +1,93 @@
...
+
+  public FilenameMaker(TemplateEngine templateEngine) {
+    this.templateEngine = templateEngine;
+    this.filenameTemplate = DEFAULT_FILENAME_TEMPLATE;
+  }
+
+  public String filenameFor(StubMapping stubMapping) {
+    FilenameTemplateModel model = new FilenameTemplateModel(stubMapping);
+    String safeFilename = makeSafeFileName(model, "json");
+    return templateEngine.getUncachedTemplate(filenameTemplate).apply(safeFilename);
```
tomakehurst, line 51: 2 things here:
1) We need to render the template first, then sanitise the result (so that we're sure it's a valid filename).
2) The default structure should come from a default template string, rather than replicating the string concatenation logic from the old code.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -597,6 +599,12 @@ public String bindAddress() {
     return bindAddress;
   }
 
+  @Override
+  public FilenameMaker getFilenameMaker() {
+    filenameMaker = new FilenameMaker();
```
tomakehurst, line 604: Doesn't seem like there's any way for the user to pass their own template when using this options impl.

Which also shows we're missing some tests. Please can you add acceptance test cases, both for this way of supplying options and the CLI alternative.

craftsman228: You're right. Will add the tests.

src/test/java/com/github/tomakehurst/wiremock/common/SafeNamesTest.java
tomakehurst, line 1: Since this has been deleted but the same logic is still present in `FilenameMaker` I think we need a new test class that runs a very similar set of tests against the default template.

craftsman228: Agree. `FilenameMakerTest` test class had been added. Please take a look.

src/test/java/com/github/tomakehurst/wiremock/common/FilenameMakerTest.java
```diff
@@ -1,66 +1,79 @@
...
+  @Test
+  public void generatesNameFromStubNameWhenPresent() {
+    StubMapping mapping =
+        WireMock.get("/named").withName("This is a NAMED stub").willReturn(ok()).build();
 
-        assertThat(
-                filenameMaker.filenameFor(mapping),
-                is("get-named-" + mapping.getId() + ".json"));
-    }
+    assertThat(filenameMaker.filenameFor(mapping), is("get-named-" + mapping.getId() + ".json"));
```
tomakehurst, line 44: We seem to have lost the logic for including the name when it's present - this only uses the method + URL + ID.

tomakehurst:
When I try running it standalone and creating a persistent stub I get an NPE thrown. I'll push a failing test for this + the name logic to your branch.

src/test/java/com/github/tomakehurst/wiremock/common/FilenameMakerTest.java
craftsman228: Added logic for smart detection `name` inside the stub and decoration existing template. Also added logic for detection `url` or `urlPath` because it could be both cases.

src/main/java/com/github/tomakehurst/wiremock/common/filemaker/FilenameMaker.java
```diff
@@ -0,0 +1,102 @@
...
+import java.util.regex.Pattern;
+import org.apache.commons.lang3.StringUtils;
+
+public class FilenameMaker {
+  public static final String DEFAULT_FILENAME_TEMPLATE =
+      "{{{request.method}}}-{{{request.url}}}-{{{id}}}";
+  private static final String NAME_TEMPLATE = "{{{name}}}-";
+  private static final Pattern NON_ALPHANUMERIC = Pattern.compile("[^\\w-.]");
+  private static final String DEFAULT_EXTENSION = ".json";
+  private static final String POINT = ".";
```
lmartelli, line 33: I believe `DOT` is more commonly used in this context.

PR #2139, opened 2023-04-18
ConsoleNotifyingWiremockNetworkTrafficListener accepts custom encoding

### LINKED ISSUE #1541: With --print-all-network-traffic switch enabled, console log is littered with decoding errors for proxied request/response bodies which are not decodable
WireMock version: Latest master branch as of 6/22/2021
Starting `com.github.tomakehurst.wiremock.standalone.WireMockServerRunner` with arguments `--port 1080 --verbose --enable-browser-proxying --print-all-network-traffic`

The console has a lot of errors coming from the code below (`ConsoleNotifyingWiremockNetworkTrafficListener`).  It is blindly decoding body bytes, assuming (I think) that they always contain UTF-8 text bytes.  When they do not (an image response, for example, or some other encoding of text), an error is thrown and hits the console with a stack trace (see the two commented out lines).  I don't think this needs to be too smart, but it may make sense to simply log  "[Non-UTF-8 Response]" or something in the catch handler, or a hex representation of the bytes.

```
 @Override
    public void incoming(Socket socket, ByteBuffer bytes) {
        try {
            CONSOLE_NOTIFIER.info("Incoming bytes: " + decoder.decode(bytes));
        } catch (CharacterCodingException e) {
            //CONSOLE_NOTIFIER.error("Problem decoding network traffic", e);
        }
    }

    @Override
    public void outgoing(Socket socket, ByteBuffer bytes) {
        try {
            CONSOLE_NOTIFIER.info("Outgoing bytes: " + decoder.decode(bytes));
        } catch (CharacterCodingException e) {
           // CONSOLE_NOTIFIER.error("Problem decoding network traffic", e);
        }
    }
```

Our use case is: we are using browser-proxying mode and snapshotting to record all requests, and were trying to debug an unrelated issue.  This was making the logs very noisy.

gsmith85:
Allows specification of a custom `Charset` for on
`ConsoleNotifyingWiremockNetworkTrafficListener` allowing flexibility for service mocks that include non UTF-8 characters in their responses.

Additionally:
- Adds test coverage for `ConsoleNotifyingWiremockNetworkTrafficListener`.
- Changes behavior of `incoming` and `outgoing` so that if a ByteBuffer with a Charset other than the one specified by
`ConsoleNotifyingWiremockNetworkTrafficListener` is provided, a succinct message is logged. This replaces the previous behavior that dumped the entire CharacterCodingException. This was extremely noisy and did not provide useful additional diagnostics.

Fixes #1541


src/main/java/com/github/tomakehurst/wiremock/http/trafficlistener/ConsoleNotifyingWiremockNetworkTrafficListener.java
```diff
@@ -16,44 +16,63 @@
...
     implements WiremockNetworkTrafficListener {
-  private static final ConsoleNotifier CONSOLE_NOTIFIER = new ConsoleNotifier(true);
+  private static final ConsoleNotifier DEFAULT_CONSOLE_NOTIFIER = new ConsoleNotifier(true);
+  private static final Charset DEFAULT_CHARSET = StandardCharsets.UTF_8;
+
+  private final Notifier notifier;
+  private final Charset charset;
+  private final CharsetDecoder charsetDecoder;
+
+  ConsoleNotifyingWiremockNetworkTrafficListener(Notifier notifier, Charset charset) {
```
oleg-nenashev, line 37: Maybe public too? Can be done later if needed

oleg-nenashev (commented):
Looks good to me. Will delve into the merge conflict

src/main/java/com/github/tomakehurst/wiremock/http/trafficlistener/ConsoleNotifyingWiremockNetworkTrafficListener.java
gsmith85: It's been a little bit, but this choice was made around the fact that this type is named `ConsoleNotifying*`.  It seemed off to allow for instantiating it with a custom notify that could notify to places other than the console.  So per that consideration and Effective Java [Item 13](https://github.com/HugoMatilla/Effective-JAVA-Summary#13-minimize-the-accessibility-of-classes-and-members), I'm intentionally minimizing accessibility.

I do think there's potentially an opportunity here to create a simple or [skeletal](https://github.com/HugoMatilla/Effective-JAVA-Summary#18-prefer-interfaces-to-abstract-classes)  type that would expose the ability to create a `WiremockNetworkTrafficListener` with an arbitrary notifier charset combo.

I'd probably have that be structured as:

```
public final class WiremockNetworkTrafficListeners {
  private WiremockNetworkTrafficListeners() {}

  public static create(Notifier notifier, CharSet charset) {
    return new SimpleWiremockNetworkTrafficListener(notifier, charset);
  }

  public static createConsoleNotifying() {
    return new SimpleWiremockNetworkTrafficListener(DEFAULT_CONSOLE_NOTIFIER, DEFAULT_CHARSET)
  }
  
  public static createConsoleNotifying(CharSet charset) {
    return new SimpleWiremockNetworkTrafficListener(DEFAULT_CONSOLE_NOTIFIER, charset)
  }  

  public static create()
}

class SimpleWiremockNetworkTrafficListener implements WiremockNetworkTrafficListener {
  // Basically all the implementation that currently exists in ConsoleNotifying
}
```

In Wiremock 2 you'd probably keep a shim `ConsoleNotifyingNetworkTrafficListener` type, but in 3 I'd drop it altogether for the `WiremockNetworkTrafficListeners` factory pattern.

Let me know what you think.  But in any case, as long as it's okay with you let's do it as part of a separate PR?

gsmith85:
Fixed the Spotless issues @oleg-nenashev so hopefully with your approval and once CI runs we can get this in 🙌 .

gsmith85:
Not sure why, but I unless I'm going crazy, I had to run Spotless a second time to completely address fix identified issues.  So apologies for not clearing them all on last commit I _think_ we should be good now.

oleg-nenashev:
> Not sure why, but I unless I'm going crazy, I had to run Spotless a second time to completely address fix identified issues. So apologies for not clearing them all on last commit I think we should be good now.

No problems at all. Commits are free, we can always squash-merge if needed 😄 
I am also a frequent victim of some of the rules we have enforced, so I totally understand

src/main/java/com/github/tomakehurst/wiremock/http/trafficlistener/ConsoleNotifyingWiremockNetworkTrafficListener.java
oleg-nenashev: I think it would be a good improvement for WireMock's API extensibility!
If you are open to submitting a separate PR for that, indeed it would be the best approach

oleg-nenashev (approved):
LGTM, and thanks!

Issue #2137, opened 2023-04-18
Unexpected interplay: repeatsAsScenarios and StubMappingTransformer


gsmith85:
I am using `StubMappingTransformers` to rewrite timestamps out of a series of requests that should logically behave as a Record and Playback scenario.

A simplified example of this are the following two ordered requests:

```
{“foo” : “bar”, “timestamp” : 1} // Response is “baz”
{“foo” : “bar”, “timestamp” : 2} // Response is “qux”
```

Both requests are rewritten, using a `StubMappingTransformer` to: `{“foo” : “bar”, “timestamp” : “${json-unit.any-number}"}`

Unfortunately, `StubMappingTransformers` don’t run until after scenario creation, so despite the requests being identical the aren’t correlated.  

https://github.com/wiremock/wiremock/blob/60e9e858068548786af4a1a434b52fd1376c4d43/src/main/java/com/github/tomakehurst/wiremock/recording/SnapshotStubMappingPostProcessor.java#L74 .

The result is that during replay the response is “baz”, “baz” rather than “baz”, “qux” because both requests will match the first of two `StubMappings` for request `{“foo” : “bar”, “timestamp” : X}`.

I personally believe that this represents a subtle bug and the reasonable behavior is to correlate scenarios post `StubMappingTransformer`; in replay mode “baz” “qux” should be the result.

The change to fix this is minimal and I’ve made it locally, but before opening a PR, I wanted to engage with the community to see if there is anything subtle that I’m missing.  FWIW The `SnapshotStubMappingPostProcessor` has no test coverage that is contrary to this, and the existent documentation is ambiguous.

gsmith85:
Perhaps an even larger problem with the existing implementation of `SnapshotStubMappingPostProcessor` and a strong indication that we should run SnapshotStubMappingTransformers first, is that because transforms currently happen last, if a transformer modifies the stub's response for a stub where the response body is extracted, it will not reflect the change made by the transform while the `List<StubMapping>` returned by `process` will.

PR #2140, opened 2023-04-19
Correctly handle scenarios with transformed stubs

gsmith85:
Revises SnapshotStubMappingPostProcessor to run stub transformers first.  That way, if the transformer results in StubMapping requests being equal they are properly
 deduplicated or turned into scenarios.

Cleans up process method to make clearer the three phases of processing.

Additionally:
- Converts SnapshotStubMappingPostProcessorTest.TEST_STUB_MAPPINGS to instance variable to avoid sharing mutable state between test cases which caused the suite to be order dependent.

Fixes: #2137


gsmith85:
Hey @tomakehurst gentle bump on this.  This issue has my team maintaining a kludgy internal fork of Wiremock, I'd love to get this into the main repo :).

oleg-nenashev (commented):
Resolved the merge conflicts. Now needs @tomakehurst

tomakehurst:
Can we fix the build errors also? Then I'll take a look.

oleg-nenashev:
@tomakehurst AFAICT there are codestyle errors only.

gsmith85:
> @tomakehurst AFAICT there are codestyle errors only.

Merging did actually result in introducing a bug, I've resolved it.  All tests are passing now, and Spotless has been reapplied so this one should be good to go too.

oleg-nenashev:
Thanks @gsmith85 , and sorry for messing up the merge

gsmith85:
@tomakehurst @oleg-nenashev are we a go here?  Anything else you need form me to get this merged?

oleg-nenashev (approved):
I had to sleep on it, and after all I think that the new behavior is wha8t we need for the real use-cases. There is a risk of slight performance degradation, but it `repeatAsScenarios` is rather an advanced feature so it should not impact many users.

Merging it towards 3.0 Beta and annotation as a breaking change is something I'm +1 for. Needs Tom ofc

src/test/java/com/github/tomakehurst/wiremock/recording/SnapshotStubMappingPostProcessorTest.java
```diff
@@ -57,21 +63,49 @@ public StubMapping apply(StubMapping stubMapping) {
...
 
     assertThat(actual, hasSize(2));
     assertThat(actual.get(0).getRequest().getUrl(), equalTo("/foo/transformed"));
     assertThat(actual.get(1).getRequest().getUrl(), equalTo("/bar/transformed"));
   }
 
   @Test
-  public void processExtractsBodiesWhenMatched() {
+  public void
+      process_withShouldRecordRepeatsAsScenariosAndTransformer_runsTransformerBeforeScenarioProcessor() {
```
tomakehurst, line 75: I'm finding these test names hard to make sense of, and they're inconsistent in style with the rest of the codebase. Please could we make them statements of outcome and camel cased?

src/main/java/com/github/tomakehurst/wiremock/recording/SnapshotStubMappingPostProcessor.java
```diff
@@ -47,30 +49,46 @@ public SnapshotStubMappingPostProcessor(
...
     }
 
     if (shouldRecordRepeatsAsScenarios) {
       new ScenarioProcessor().putRepeatedRequestsInScenarios(processedStubMappings);
     }
 
-    // Run any stub mapping transformer extensions
-    return processedStubMappings.stream().map(transformerRunner).collect(Collectors.toList());
+    // 3. Extract response bodies to a separate file, if applicable.
+    extractStubMappingBodies(processedStubMappings);
```
tomakehurst, line 78: This is a separate behaviour change and should be part of a separate PR. Also, I think there's a better way of doing this generally, which is to define a size threshold at which stub bodies are extracted as a startup parameter. That way it works in more contexts that just the recorder and allows for more control.

src/main/java/com/github/tomakehurst/wiremock/recording/SnapshotStubMappingPostProcessor.java
```diff
@@ -47,30 +49,46 @@ public SnapshotStubMappingPostProcessor(
   }
 
   public List<StubMapping> process(Iterable<StubMapping> stubMappings) {
-    final Multiset<RequestPattern> requestCounts = HashMultiset.create();
-    final List<StubMapping> processedStubMappings = new ArrayList<>();
-
+    // 1. Run any applicable StubMappingTransformers against the stub mappings.
+    ArrayList<StubMapping> transformedStubMappings = new ArrayList<>();
```
tomakehurst, line 53: This plus the next few lines could be done more neatly by mapping over a stream.

gsmith85: If we're using `Iterable` as the type for `stubMappings` this actually may be worse from a readability perspective:

```java
List<StubMapping> transformedStubMappings = 
        StreamSupport.stream(stubMappings.spliterator(), false)
                .map(transformerRunner)
                .collect(toList());
```

Note that when using `Iterables` you must use `StreamSupport` to gain access to a `Steam`.

So if you really want to use streams here I'd propose we change the method signature of `process` as well to use `Collection` what you end up with is:

```java
public List<StubMapping> process(Collection<StubMapping> stubMappings) {
    // 1. Run any applicable StubMappingTransformers against the stub mappings.
    List<StubMapping> transformedStubMappings =
            stubMappings.stream().map(transformerRunner).collect(toList());
    ...
```

`Collection` is a bit more narrow than `Iterable` but I think general enough and appropriate for the semantics of `stubMappings`.

src/test/java/com/github/tomakehurst/wiremock/recording/SnapshotStubMappingPostProcessorTest.java
gsmith85: Apologies for going in a different direction with test naming here.

A bit on rationale, the pattern `method_condition_expectedOutcome` was chosen based on capturing the recommended components for unit test naming per [Google Testing](https://testing.googleblog.com/2014/10/testing-on-toilet-writing-descriptive.html) and specifically I've included outcomes in this context because documenting that aspect feels valuable.  The behavior of the `SnapshotStubMappingPostProcessor` is a bit subtle, I want to make sure each test's goal is articulated so the reader isn't left attempting to parse the intended behavior from the assertion set.

All of this can be done in a fashion more aligned to the conventions of this project.  I've switched over to a `method<With>condition<Should>expectedOutcome` format, so I think we're in good shape.

gsmith85:
@tomakehurst I believe I've addressed your comments, appreciate the guidance on this one.  Let me know if you need any other tweaks and I can be more responsive to help get this pushed over the line.

tomakehurst:
Thanks @gsmith85 !

PR #2157, opened 2023-04-21
code changes for adding support for form parameter matching

### LINKED ISSUE #383: Please add Form parameter matching ( Not query parameter matching)
Hi,

I am using WireMock 2.0.10-beta version. I didn't see any API for Form parameter matching. Not query parameter matching.

Proposed API

withFormParam("p1", equalTo("v1"));

OR

withRequestParam("p1", equalTo("v1"));  //  Work for both GET & Non-GET methods

Thank you!

kapishmalik:
Resolves #383 

Code changes for adding support for form parameter matching. 

Added tests for mentioned scenarios:

- Correct serialisation and deserialisation of the matcher.
- Parameter keys that contain characters that would be URL encoded, in particular array-style keys like address[postcode]
- Expected parameter values containing characters that would be URL encoded.
- Diff report


src/test/java/com/github/tomakehurst/wiremock/matching/RequestPatternTest.java
```diff
@@ -35,437 +54,498 @@
 
 public class RequestPatternTest {
 
-    @Test
```
oleg-nenashev, line 38: Would be better to keep formatting of the original tests even if it violates the standard. Reviiews with `?w=1`  is a solution, of course

src/main/java/com/github/tomakehurst/wiremock/http/Request.java
```diff
@@ -60,6 +60,10 @@ interface Part {
 
   QueryParameter queryParameter(String key);
 
+  FormParameter formParameter(String key);
```
oleg-nenashev, line 63: Without default implementation, it breaks binary compatibility

oleg-nenashev (changes requested):
Thanks for the pull request! The binary compatibility issue in the interface needs to be fixed before this pull request can be merged.

oleg-nenashev:
FWIW, there is https://github.com/wiremock/wiremock/issues/2153 for providing clear separation between really public APIs and not so ones

src/test/java/com/github/tomakehurst/wiremock/matching/RequestPatternTest.java
kapishmalik: If I keep it same, then spotless check will fail.

src/main/java/com/github/tomakehurst/wiremock/http/Request.java
kapishmalik: incorporated

kapishmalik:
incorporated review comment changes

oleg-nenashev (approved):
LGTM

src/main/java/com/github/tomakehurst/wiremock/http/Request.java
```diff
@@ -60,6 +60,14 @@ interface Part {
 
   QueryParameter queryParameter(String key);
 
+  default FormParameter formParameter(String key) {
```
tomakehurst, line 63: Can we ditch these defaults and ensure we've implemented these everywhere instead?

tomakehurst:
Apart from the couple of minor comments above, looks good. Happy to merge this once these are resolved and Oleg's comment on the doc PR is also resolved.

tomakehurst:
Apart from the couple of minor comments above, looks good. Happy to merge this once these are resolved and Oleg's comment on the doc PR is also resolved.

src/main/java/com/github/tomakehurst/wiremock/http/Request.java
kapishmalik: Done. We have implemented these everywhere. Oleg suggested adding this as it might break the implementation for those who might be using this interface as this is part of the library and exposed.

src/test/java/com/github/tomakehurst/wiremock/matching/RequestPatternTest.java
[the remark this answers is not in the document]
kapishmalik: I just added my change but spotless check started failing. It was very minor test so I just removed it for now. Let me know if you have any work around to fix this.

src/main/java/com/github/tomakehurst/wiremock/http/Request.java
kapishmalik: I feel that it is very unlikely that anyone using it as we are using this interface as base for our receiving requests, matching, and stubbing it wherever required. Consumers of Wiremock may not be using it for any of these purposes. But it is always better to add a safety net.

tomakehurst:
Looks like you've deleted your changes to `RequestPatternTest` entirely now?

We need some serialisation/deserialisation tests, which I think belong in there.

kapishmalik:
> Looks like you've deleted your changes to `RequestPatternTest` entirely now?
> 
> We need some serialisation/deserialisation tests, which I think belong in there.

as discussed, first applied formatting changes followed by added new UTs.

src/test/java/com/github/tomakehurst/wiremock/matching/RequestPatternTest.java
```diff
@@ -252,6 +256,70 @@ public void bindsToJsonCompatibleWithOriginalRequestPatternWithQueryParams() thr
...
+            + "        },                     \n"
+            + "        \"key2\": {          \n"
+            + "            \"matches\": \"value2\" \n"
+            + "        }                      \n"
+            + "    }                          \n"
+            + "}",
+        actualJson,
+        true);
+  }
+
```
tomakehurst, line 322: Please can we also have a test case showing that serialisation works correctly?

kapishmalik: sure, incorporated

kapishmalik: I believe you mean deserialization. I have now both tests serialization and deserialization.

tomakehurst:
Great stuff, thanks!

PR #2172, opened 2023-05-05
Provide separate Jetty 10 version

henrik242:
This is needed when using other dependencies with a Jetty 10 dependency (such as Solr 9.2) together with Wiremock.


oleg-nenashev (changes requested):
So far this version is not really separate. Gradle definition needs to be changed to build both artifacts. Or, if @tomakehurst sees it as important use-cases, it could be maintained in a branch. Maybe switching to https://github.com/wiremock/wiremock-testcontainers-java is actually a better solution for this case

tomakehurst:
I'm still thinking Docker or standalone are better options than us having to maintain both Jetty 10 and 11 versions.

oleg-nenashev:
Closing the pull request based on the discussion above. The pull request would need a significant rework to have it compatible with the 3.x extensions rework anyway

henrik242:
In case anyone needs it, here's an updated jetty10 fork of wiremock 3.3.1: https://github.com/henrik242/wiremock/tree/3.3.1.jetty10

oleg-nenashev:
Nice @henrik242! If you'd like to document the fork as a temporary solution somewhere on the website, it might be helpful. Meanwhile I will refactor this ticket a bit for better discoverability

henrik242:
@oleg-nenashev Sure, any suggestions to where it could be documented?

henrik242:
Here's the jetty10 fork of 3.4.1: https://github.com/henrik242/wiremock/tree/3.4.1.jetty10

tomakehurst:
@henrik242 we've started working towards separate JARs for different Jetty versions. Might be a good pattern to copy for 10 also:
https://github.com/wiremock/wiremock/tree/reta-issue-2395-jetty-12

PR #2175, opened 2023-05-11
feat: add a requestedFor method allowing to pass Http method as parameter

ytvnr:
It is useful for parameterized test, when you also use `request(String method, UrlPattern urlPattern)`

So you are able to write your test like this:

```java
@ParameterizedTest
@ValueSource(strings = {"GET", "POST", "PUT", "HEAD", "TRACE", "PATCH", "OPTIONS", "DELETE", "ANY", "RANDOM"})
public void verifyRequestedForSameMethodAsRequest(String method) {
    wiremock.stubFor(request(method, urlEqualTo("/methods")).willReturn(ok("response"));
    // ... Test stuff here
    verify(requestedFor(method, urlEqualTo("/methods")));
}
```

- I look at https://wiremock.org/docs/verifying/ and there is not an exhaustive list of the applicable method, so I did not modified the documentation

## References

- https://github.com/wiremock/wiremock/issues/2174

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -840,6 +840,10 @@ public class WireMock {
     return new RequestPatternBuilder(RequestMethod.ANY, urlPattern);
   }
 
+  public static RequestPatternBuilder requestedFor(String method, UrlPattern urlPattern) {
+    return new RequestPatternBuilder(RequestMethod.fromString(method), urlPattern);
+  }
+
   public static RequestPatternBuilder requestMadeFor(
       String customMatcherName, Parameters parameters) {
     return RequestPatternBuilder.forCustomMatcher(customMatcherName, parameters);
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
Definitely +1. I would rather deprecate the older methods TBH, because they may cause quite a lot of confusion.  Also agree about the documentation, though it would be great to document this particular method there

ytvnr:
@oleg-nenashev, thank you for the review.
Do you want me to deprecate the older methods in another commit ?
Maybe it would be the opportunity to add methods overload taking the `RequestMethod` as a parameter and not just a string ?
We would have the additional:
- `RequestPatternBuilder requestedFor(HttpMethod method, UrlPattern urlPattern)`
- `MappingBuilder request(HttpMethod method, UrlPattern urlPattern)`
wdyt ?

oleg-nenashev:
No strong opinion, I guess @tomakehurst should make a call on deprecation

tomakehurst:
Please could you run `./gradlew spotlessApply` and push the changes?

tomakehurst:
Re deprecation, I assume you're talking about `getRequestedFor(...)`, `postRequestedFor(...)` etc.?

If so, I don't think we ought to deprecate. They're convenience DSL methods for the most commonly used HTTP methods, and can co-exist with the general case.

tomakehurst:
Merging the latest from master should fix the test error

ytvnr:
It should be ok @tomakehurst (at least it `./gradlew check` worked locally 😄 )

oleg-nenashev (approved):

PR #2192, opened 2023-05-27
Add convenient method for matching absence of query param in a request

G-Basak:
This pull request adds a convenient method for matching absence of query param in a request.
Usage example :
```java
//after this pull request
verify(
    getRequestedFor(urlPathEqualTo("without/queryParam"))
        .withoutQueryParam("test-param"));
```

It was possible to verify the same even without this method as shown below, but felt counter-intuitive (given there is already a `withoutHeader` method).
```java
//before this pull request
verify(
    getRequestedFor(urlPathEqualTo("without/queryParam"))
        .withQueryParam("test-param", absent()));
```

## References

- This change was inspired by `withoutHeader` and follows similar pattern.
https://github.com/wiremock/wiremock/blob/2f8b5263d984b3cf212f898e52d4d1b4a52d338a/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java#L166
- There is another pull request for adding similar method for form param.
https://github.com/wiremock/wiremock/pull/2193

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java
@@ -193,6 +193,11 @@ public class RequestPatternBuilder {
     return this;
   }
 
+  public RequestPatternBuilder withoutQueryParam(String key) {
+    queryParams.put(key, MultiValuePattern.absent());
+    return this;
+  }
+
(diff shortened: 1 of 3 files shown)


oleg-nenashev (approved):
Looks good to me, because indeed there are `without` methods and it's just some syntax sugar. Tagging @tomakehurst to get his opinion so that we're aligned on the vision for API sugar

PR #2198, opened 2023-05-30
2197 | added field "mappings" to StubMapping and its processing if it exists

### LINKED ISSUE #2197: Support multiple mappings files when remotely loading stubs from the filesystem
### Proposal

While it's possible at present to push a collection of stub mappings from the file system to a remote instance via `wireMock.loadMappingsFrom("/my/path")` this only supports single stub mapping files, not the newer multiple mapping file format (with the `"mappings"` key at the root).

This would be desirable so that sets of related stubs can be combined into single files, without needing to merge the entire set into a single file for remote import.

### References

https://github.com/wiremock/wiremock/issues/1921

DjerohN:
This pull request adds field "mappings" to StubMapping class and processing to load() method of RemoteMappingsLoader class if present

## References

- Resolves #2197 issue in a simple way for now, because I don't know plans for 3.0 imports

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/standalone/RemoteMappingsLoader.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/standalone/RemoteMappingsLoader.java
@@ -26,10 +26,13 @@ import com.github.tomakehurst.wiremock.client.WireMock;
 import com.github.tomakehurst.wiremock.common.BinaryFile;
 import com.github.tomakehurst.wiremock.common.ContentTypes;
 import com.github.tomakehurst.wiremock.common.FileSource;
+import com.github.tomakehurst.wiremock.common.Json;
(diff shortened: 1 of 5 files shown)


DjerohN:
@tomakehurst @oleg-nenashev

tomakehurst:
This isn't the right approach - there's already `StubMappingCollection` for this purpose which is used when reading the `mappings` directory via `JsonFileMappingsSource`. I suggest we adapt `RemoteMappingsLoader` to work the same way.

DjerohN:
Used `StubMappingCollection` as @tomakehurst  suggested

tomakehurst:
Please could you run `./gradlew spotlessApply` and push the fixes?

DjerohN:
Done

tomakehurst:
Thanks!

PR #2207, opened 2023-06-07
Add WireMock co-maintainers role and document the merge process

oleg-nenashev:
This change should help us to increase velocity of changes

- [x] Add a role of a WireMock Co-maintainer, which also stems from a co-maintainer in the [Community charter](https://github.com/wiremock/community/blob/main/governance/README.md)
- [x] Document the Merge process for WireMock 3,x Beta
- [x] Document the process in the community repo, e.g. make a call regarding the CLA for co-maintainers

## Motivation

There are a lot of open pull requests that are important to end users and contribute to the WireMock 3.x scope. They keep accumulating overlaps and potential merge conflicts, and also do not get evaluation in the Beta releases.

The only current maintainer, Tom Akehurst, combines a number of roles and does not have much bandwidth to review and merge pull requests. So we need more review and maintainer bandwidth

## Nominated co-maintainers

I would suggest to make the following contributors co-maintainers, or to work towards enabling them.

- @oleg-nenashev 
- @Mahoney 
- @timtebeek 
- @oleg-nenashev 
- @rodolpheche - if we include WireMock Docker into the scope now or later

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,3 +106,45 @@ The website sources are located here: [wiremock/wiremock.org](https://github.com
 All the documentation is located under the `_docs` directory as Markdown files,
 and it can be edited with all modern text editors and IDEs.
 See the repository's contributor guide for more information.
+
+## Merge process
+
+Merges to this repository can be performed by the WireMock maintainer ([Tom Akehurst](https://github.com/tomakehurst))
+and by _co-maintainers_ assigned by him.
+This is a [community role](https://github.com/wiremock/community/blob/main/governance/README.md)
+designed for WireMock itself and other key repositories,
+specifically to facilitate review and changes while WireMock 3 is in beta
+and receives a lot of incremental patches.
+
+- The maintainers are responsible to verify the pull request readiness
+  in accordance with contributing guidelines (e.g. code quality, test automation, documentation, etc.).
+  The pull request can only be approved if these requirements are met
+- In the beginning, a review by one co-maintainer is required for the merge,
+  unless there are negative reviews and unaddressed comments by other contributors
+- After the approval, it is generally recommended to give at least 24 hours for reviews before merging
+
+Beyond WireMock 3.x Beta releases, the scope of responsibilities for co-maintainers
+is yet to be determined based on the experiences with this role for the Beta versions.
+
+### What can be merged by co-maintainers
+
+While WireMock 3.x is in Beta, co-maintainers can merge the following pull requests:
+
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -106,3 +106,45 @@ The website sources are located here: [wiremock/wiremock.org](https://github.com
...
+
+- Minor features and improvements that do not impact the WireMock architecture
+- Refactorings, including the major ones, e.g. Guava replacement
+- Test Automation
+- Non-production repository changes: documentation (including Javadoc), GitHub Actions, bots and automation
+- Dependency updates for shaded dependencies, patch/minor versions for projects following the Semantic Versioning notation
+
+### What CANNOT be merged by co-maintainers without BDFL’s approval
+
+The following changes need a review by Tom Akehurst before being merged.
```
jamietanna, line 141: Although I agree with this - if something was to be merged, it's still possible to revert it before the next release, right? I _assume_ Tom right now is the only person releasing the library, and it's unlikely(?) that folks are generally using SNAPSHOTs or pulling the latest commits, so it may be less worrisome if something did get merged and needs to get unpicked?

jamietanna:
As someone who's previously been involved in the project - and has some arguably co-contributor access - would we also be doing a review of who currently has write access to the repo and reviewing their access?

I'd be very happy if this review found that my access wasn't necessary and you'd like to remove it for now, until I get to the point of being more involved in code contributions to Wiremock.

oleg-nenashev:
Thanks for the feedback @jamietanna!

> Although I agree with this - if something was to be merged, it's still possible to revert it before the next release, right? I assume Tom right now is the only person releasing the library, and it's unlikely(?) that folks are generally using SNAPSHOTs or pulling the latest commits, so it may be less worrisome if something did get merged and needs to get unpicked?

The idea is to actually automate the releases and to make them possible for co-maintainers. For the continuous delivery it is important to keep the main branch ready for release, so reverting is not the best approach. It also becomes a manual procedure once there is a merge conflict on the top of it. 

So it is better to not merge changes that have not yet received a signoff

> As someone who's previously been involved in the project - and has some arguably co-contributor access - would we also be doing a review of who currently has write access to the repo and reviewing their access?

Yes. Some audit would be needed. At the same time, there is no intent to revoke permissions of any active contributor

oleg-nenashev:
Approved by @tomakehurst in the DM chat

jamietanna (approved):
Makes sense, thanks for the clarification!

PR #2209, opened 2023-06-08
Codebase improvements

pks-1981:
Codebase improvements

## References

N/A

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/SnapshotDslAcceptanceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/SnapshotDslAcceptanceTest.java
@@ -79,7 +79,7 @@ public class SnapshotDslAcceptanceTest extends AcceptanceTestBase {
   }
 
   @Test
-  public void snapshotRecordsAllLoggedRequestsWhenNoParametersPassed() throws Exception {
+  void snapshotRecordsAllLoggedRequestsWhenNoParametersPassed() throws Exception {
     targetService.stubFor(
         get("/one")
             .willReturn(
@@ -119,7 +119,7 @@ public class SnapshotDslAcceptanceTest extends AcceptanceTestBase {
   }
 
   @Test
-  public void supportsFilteringByCriteria() throws Exception {
+  void supportsFilteringByCriteria() throws Exception {
     client.get("/things/1");
     client.get("/things/2");
     client.get("/stuff/1");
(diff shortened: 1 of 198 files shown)


tomakehurst:
Could you elaborate the description of this a bit?

"Codebase improvements" is very generic and it's hard to review 223 files without some understanding of the expected scope of the changes.

pks-1981:
@tomakehurst 
Hello
Added description

pks-1981:
@tomakehurst @timtebeek 
Hello.
Plz review it.
I can split this MR into 4-5 parts if needed.

pks-1981:
@tomakehurst 
Sorry to ask you to look again...
But my MR is need?
Its main task is to reduce the number of Sonar or/and "IDEA inspect code" issues in tests

src/test/java/com/github/tomakehurst/wiremock/SnapshotDslAcceptanceTest.java
```diff
@@ -120,7 +119,7 @@ void snapshotRecordsAllLoggedRequestsWhenNoParametersPassed() throws Exception {
   }
 
   @Test
-  void supportsFilteringByCriteria() {
+  public void supportsFilteringByCriteria() throws Exception {
```
tomakehurst: Not sure about this - adding `public` and `throws Exception` where not needed just adds noise IMO.

pks-1981: I feel more comfortable when IDEA no show warning)))

oleg-nenashev (commented):
TBH I would rather suppress the sonar warning, because "In this context, JUnit5 test classes can have any visibility but private, however, it is recommended to use the default package visibility, which improves readability of code.". Not sure how much it improves readability, especially for those who are used to JUnit 4 public requirement

A call for @tomakehurst to make on the code style IMHO

tomakehurst:
I marginally prefer not to have to write a redundant `public` but I don't care massively - I don't think readability is massively impacted either way. Tempted to just disable the rule.

pks-1981:
@tomakehurst 
A little sorry for the work done, but it happens.
close then?

tomakehurst:
Yeah, let's close this one and create some smaller, more focussed ones if you'd still like to proceed.

And let's leave the test method visibility alone - this just inflates the patch size without adding any real value IMO.

PR #2213, opened 2023-06-09
Add changelog generation with Release Drafter

oleg-nenashev:
subj

### DIFF
--- a/.github/workflows/changelog-draft,yml
+++ b/.github/workflows/changelog-draft,yml
@@ -1,0 +1,19 @@
+name: Release Drafter
+
+on:
+  push:
+    branches:
+      - main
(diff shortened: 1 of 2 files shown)


.github/workflows/changelog-draft,yml
```diff
@@ -0,0 +1,19 @@
+name: Release Drafter
+
+on:
+  push:
```
tomakehurst, line 4: Better to make it a manual trigger maybe?

oleg-nenashev: Auto-generation is actually handy

oleg-nenashev:
OK, let's give it a try

PR #2215, opened 2023-06-11
Replace Guava classes

pks-1981:
Replace Guava classes
com.google.common.base.Preconditions.checkNotNull -> Objects.requireNonNull
com.google.common.base.Charsets.UTF_8 -> java.nio.charset.StandardCharsets.UTF_8
com.google.common.base.Objects -> java.util.Objects

## References

https://github.com/wiremock/wiremock/issues/2111

## Submitter checklist

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -106,3 +106,45 @@ The website sources are located here: [wiremock/wiremock.org](https://github.com
 All the documentation is located under the `_docs` directory as Markdown files,
 and it can be edited with all modern text editors and IDEs.
 See the repository's contributor guide for more information.
+
+## Merge process
+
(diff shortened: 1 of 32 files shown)


pks-1981:
@tomakehurst @timtebeek
Hello
Plz review it

tomakehurst:
I'm not keen on replacing `firstNonNull` with `Optional` from a performance perspective since the former doesn't create a redundant object.

In general also, I think the goal here isn't to eliminate Guava completely. If it's providing something useful that doesn't have a direct JDK equivalent then let's keep it.

pks-1981:
> I'm not keen on replacing `firstNonNull` with `Optional` from a performance perspective since the former doesn't create a redundant object.
> 
> In general also, I think the goal here isn't to eliminate Guava completely. If it's providing something useful that doesn't have a direct JDK equivalent then let's keep it.

revert commits?

PR #2234, opened 2023-06-25
Feature Request: Matching Request method with isOneOf & isNoneOf

pks-1981:
Feature Request: Matching Request method with isOneOf & isNoneOf.

Adding the ability to specify a list of request methods.

https://github.com/wiremock/wiremock/issues/1434
Based on https://github.com/wiremock/wiremock/pull/1734

## Submitter checklist


pks-1981:
@tomakehurst
I will try to make a normal description as soon as possible.

But waiting review)))

pks-1981:
@tomakehurst 
Hello.
May be only methods?
            + "      \"method\" : \"GET\",                    \n"
            + "      \"methods\" : [\"GET\"]                  \n"

pks-1981:
@timtebeek @oleg-nenashev 
If you don't mind take a look...
For some advice)))

oleg-nenashev (commented):
So far looks good. A bit more of work is needed to land it but I think it could be integrated soon

pks-1981:
> So far looks good. A bit more of work is needed to land it but I think it could be integrated soon

unfortunately i have no time and ideas, right now
So i am be glad if someone help me)

pks-1981:
@tomakehurst 
Unfortunately I can't figure out how to implement DiffLine
I need help

src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
```diff
@@ -128,6 +131,73 @@ public String getName() {
         };
   }
 
+  public RequestPattern(
```
oleg-nenashev, line 134: This is when we switch to the `withSomething()` pattern 💀 
But I think we could at least avoid code duplication in the constructor, right?

src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
```diff
@@ -139,6 +209,7 @@ public RequestPattern(
       @JsonProperty("urlPathPattern") String urlPathPattern,
       @JsonProperty("urlPathTemplate") String urlPathTemplate,
       @JsonProperty("method") RequestMethod method,
+      @JsonProperty("methods") Set<RequestMethod> methods,
```
oleg-nenashev, line 212: Hmm. Was the previous constructor supposed to retain binary compatibility?

src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
pks-1981: Now RequestPattern with RequestMethod method use RequestPattern with Set<RequestMethod> methods

src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
pks-1981: It's better to leave it for now. IMHO

pks-1981:
@oleg-nenashev 
could you help me please?
With - The PR request is well described and justified, including the body and the references
My English not so good for it(

src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
```diff
@@ -46,29 +46,20 @@ public static RequestMethod fromString(String value) {
...
+  public static Set<RequestMethod> fromSet(Set<String> values) {
+    return values.stream().map(RequestMethod::fromString).collect(Collectors.toSet());
+  }
+
   @JsonValue
   public String value() {
     return name;
   }
 
-  public boolean isOneOf(RequestMethod... methods) {
```
oleg-nenashev, line 54: Now it is a breaking change, right?

pks-1981: @oleg-nenashev 
Hello
I think RequestMethodMatcher and its use in RequestPattern and Diff are more suitable

oleg-nenashev: Yes, I do not disagree. The problem is the API breakage. We cannot track down the usages, so it would ne nice to keep historical methods though it would be perfectly fine to deprecate them and the point users to as proper usage pattern in the annotation

pks-1981: My IDE tells me that the method is not currently in use.
image
So I don't know what to say...

AlexOblizin (commented):
I think it is nice

src/main/java/com/github/tomakehurst/wiremock/verification/diff/Diff.java
```diff
@@ -319,12 +308,12 @@ private DiffLine<String> addUrlSectionWithSpacer(
...
 
   private void addMethodSection(List<DiffLine<?>> diffLineList) {
-    DiffLine<RequestMethod> methodSection =
+    DiffLine<Set<RequestMethod>> methodSection =
         new DiffLine<>(
             "HTTP method",
-            requestPattern.getMethod(),
-            request.getMethod(),
-            requestPattern.getMethod().getName());
+            new RequestMethodMatcher(requestPattern.getMethods()),
```
tomakehurst, line 314: It would be great if we could preserve the existing behaviour when only a single method is specified rather than it always appearing as a collection

src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
```diff
@@ -46,19 +47,18 @@ public static RequestMethod fromString(String value) {
...
-    return asList(methods).contains(this);
-  }
-
   public MatchResult match(RequestMethod method) {
-    boolean getOrHeadMatch =
-        this.equals(GET_OR_HEAD) && (method.equals(GET) || method.equals(HEAD));
-    return MatchResult.of(this.equals(ANY) || this.equals(method) || getOrHeadMatch);
+    return MatchResult.of(this.equals(ANY) || this.equals(method));
```
tomakehurst, line 61: Looks like we've lost the correct GET_OR_HEAD matching here

src/main/java/com/github/tomakehurst/wiremock/http/RequestMethodMatcher.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.github.tomakehurst.wiremock.matching.MatchResult;
+import com.github.tomakehurst.wiremock.matching.NamedValueMatcher;
+import java.util.Set;
+
+public class RequestMethodMatcher implements NamedValueMatcher<Set<RequestMethod>> {
+
+  private final Set<RequestMethod> requestMethods;
+
+  public RequestMethodMatcher(Set<RequestMethod> requestMethods) {
```
tomakehurst, line 26: It would be great if when serialised this was wrapped in a `oneOf` element in a similar fashion to other matchers, a) because this communicates the intent more clearly to the reader and b) permits us to e.g. add `noneOf` in the future.

tomakehurst (changes requested):
Thanks for your patience with this!

A few minor suggestions, but looking good otherwise.

src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
pks-1981: I modified the WireMock#getOrHead(UrlPattern urlPattern) method to
BasicMappingBuilder(Set.of(RequestMethod.GET, RequestMethod.HEAD), urlPattern)
The tests are working correctly.
WireMockClientAcceptanceTest#
testGetOrHeadRequestWhenGetMatchesShouldReturnAResponseBody
testGetOrHeadRequestWhenHeadMatchesShouldNotReturnAResponseBody
testGetOrHeadRequestWhenNoMethodNotMatchesShouldReturn404
testGetOrHeadRequestWhenPathDoesNotMatchShouldReturn404
I think that's enough.
What do you say?

src/main/java/com/github/tomakehurst/wiremock/http/RequestMethodMatcher.java
pks-1981: Unfortunately, I don't quite understand(
Maybe we’ll leave it like this for now?))) and change it when needed?)))

tomakehurst: What I mean is instead of
```
"method": ["GET", "POST"]
```

We'd do
```
"method": {
  "oneOf": ["GET", "POST"]
}
```

That way we give ourselves the option of putting another operator there if we ever need to. The problem with waiting until later to do that is that it'd then be a breaking change.

pks-1981: I managed to do it like oneOfRequestMethods: [methods]
I may add noneOfRequestMethods: [methods]

pks-1981: I do "isNoneOf" right away )

pks-1981: @tomakehurst
I hope this is the final version)
Please look)

pks-1981:
@tomakehurst hello)
Plz look at my updated MR

pks-1981:
@tomakehurst Hi! What about my MR? We are also waiting for 4.0?

PR #2247, opened 2023-07-05
Fix flood json string can not be null or empty from mapping matcher

emilianoalvarez91:
When using body patterns and matchesJsonPath with either `$.` or `$[?.` there are several occasions that wiremock throws a warning with the error:
`Warning: JSON path expression {Expression} failed to match document 'null' because of error 'json string can not be null or empty'`
This floods the entire wiremock logs leaving verbose pretty useless.

## References
https://github.com/wiremock/wiremock/issues/1817

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPattern.java
@@ -81,7 +81,6 @@ public class MatchesJsonPathPattern extends PathPattern {
               "Warning: JSON path expression '%s' failed to match document '%s' because %s",
               expectedValue, value, error);
 
-      notifier().info(message);
       return MatchResult.noMatch(SubEvent.warning(message));
     }
   }
@@ -108,7 +107,6 @@ public class MatchesJsonPathPattern extends PathPattern {
           .min(Comparator.comparingDouble(MatchResult::getDistance))
           .orElse(MatchResult.noMatch(subEvents));
     } catch (SubExpressionException e) {
-      notifier().info(e.getMessage());
       return MatchResult.noMatch(SubEvent.warning(e.getMessage()));
     }
   }
--- a/src/test/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPatternTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPatternTest.java
@@ -100,16 +100,14 @@ public class MatchesJsonPathPatternTest {
   }
 
   @Test
-  public void providesSensibleNotificationWhenJsonMatchFailsDueToInvalidJson() {
-    Notifier notifier = setMockNotifier();
-
+  public void providesEventMessageWhenJsonMatchFailsDueToInvalidJson() {
     StringValuePattern pattern = WireMock.matchingJsonPath("$.something");
     MatchResult match = pattern.match("Not a JSON document");
 
     assertFalse(match.isExactMatch(), "Expected the match to fail");
-    checkWarningMessageAndEvent(
-        notifier,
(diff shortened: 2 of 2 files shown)


oleg-nenashev (commented):
Looks like it's quite urgent based on votes.
Would be nice to improve propagation.

src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPattern.java
```diff
@@ -110,6 +112,8 @@ protected MatchResult isAdvancedMatch(String value) {
     } catch (SubExpressionException e) {
       notifier().info(e.getMessage());
       return MatchResult.noMatch(SubEvent.warning(e.getMessage()));
+    } catch (NullPointerException e) {
+      return MatchResult.noMatch();
```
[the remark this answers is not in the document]
emilianoalvarez91: Sure, that makes sense! I'll propagate the message

oleg-nenashev (approved):

oleg-nenashev:
Definitely needs a fix. Leaving to @tomakehurst to comments on the implementation style, the NPE approach is a matter for discussion

emilianoalvarez91:
@oleg-nenashev thanks. Yes the reason in which I was explicit with the NPE is that I want to guarantee that specific error and capture it accordingly in isAdvancedMatch, otherwise if I rethrow the initial error in getExpressionResult, there's no guarantee that isAdvancedMatch function captures the same error in the future. If @tomakehurst you have a better suggestion, I'm happy to make the change.

src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPattern.java
```diff
@@ -110,6 +112,8 @@ protected MatchResult isAdvancedMatch(String value) {
     } catch (SubExpressionException e) {
       notifier().info(e.getMessage());
       return MatchResult.noMatch(SubEvent.warning(e.getMessage()));
+    } catch (NullPointerException e) {
```
tomakehurst, line 115: I'm not keen on catching NPE. Something like this might be better (around line 94):
```java
if (expressionResult == null || expressionResult.isEmpty()) {
        if (AbsentPattern.class.isAssignableFrom(valuePattern.getClass())) {
          expressionResult = ListOrSingle.of((String) null);
        } else {
          return MatchResult.noMatch();
        }
      }
```

src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonPathPattern.java
```diff
@@ -110,6 +112,8 @@ protected MatchResult isAdvancedMatch(String value) {
     } catch (SubExpressionException e) {
       notifier().info(e.getMessage());
```
tomakehurst, line 113: We could just drop this completely. I'm not sure it's really helping, and I plan to remove all of these at some point anyway now that we've got sub-events.

tomakehurst:
An additional test case or two would definitely be a good idea also

emilianoalvarez91:
@tomakehurst I followed your suggestion of removing the notifier and by that we don't need to catch the specific error. Do you think we should go in this direction or should I rework the other solution?

oleg-nenashev (approved):
I would put the message to the `debug`-level logging, but otherwise +1

tomakehurst (approved):

tomakehurst:
LGTM.

I suspect we now also have this problem in terms of stacking up a loads of sub-events for the same reason, but we can address this separately.

oleg-nenashev:
> I suspect we now also have this problem in terms of stacking up a loads of sub-events for the same reason, but we can address this separately.

Yeah, very likely

PR #2254, opened 2023-07-10
Remove dependencies on new:ing collections with Guava (refs #2111)

tomasbjerre:
A small step removing some imports from Guava.

## References

#2111 

## Submitter checklist


tomasbjerre:
I think the failing `Build and test / build (macos-latest, 17) (pull_request)` here might be a because of a timing issue, a toggling test case maby?

Mahoney:
Yes, that's almost certainly just a performance issue... we should probably bump the expectation up to 10 seconds or something to prevent it happening again. The problem it's checking for was making them take minutes, so that should be OK.

oleg-nenashev (approved):

oleg-nenashev:
Created https://github.com/wiremock/wiremock/issues/2259 for the flake, merging:

PR #2255, opened 2023-07-10
ServerEvent: replace Guava predicate and function with STL (refs #2111)

tomasbjerre:
These are the last imports of `Predicate` and `Function` from Guava. Perhaps they are left for a reason?

## References

#2111 

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
@@ -16,8 +16,6 @@
 package com.github.tomakehurst.wiremock.core;
 
 import static com.github.tomakehurst.wiremock.common.ParameterUtils.getFirstNonNull;
-import static com.github.tomakehurst.wiremock.stubbing.ServeEvent.NOT_MATCHED;
-import static com.github.tomakehurst.wiremock.stubbing.ServeEvent.TO_LOGGED_REQUEST;
 import static com.google.common.collect.Iterables.contains;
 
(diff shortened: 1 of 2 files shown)


Mahoney (commented):
My understanding is we want to make this breaking change before going out of beta so I'd guess it's oversight, but perhaps @tomakehurst can confirm as `NOT_MATCHED` and `TO_LOGGED_REQUEST` are public and fairly widely used.

oleg-nenashev (approved):
I am in favor of doing this change for 3.0. If we want to reduce Guava footprint, removing it from APIs is definitely a right thing to do

oleg-nenashev (commented):
The proposal from @tomasbjerre is to drop static fields entirely as a part of the change. I will document the new locations afterwards

```
  public static final Function<ServeEvent, LoggedRequest> TO_LOGGED_REQUEST =
      ServeEvent::getRequest;

  public static final Predicate<ServeEvent> NOT_MATCHED = ServeEvent::isNoExactMatch;
```

oleg-nenashev:
@tomasbjerre WDYT?

tomasbjerre:
If you want me to just remove the fields I can do that.

oleg-nenashev:
@tomasbjerre would be great, thanks!

tomasbjerre:
> @tomasbjerre would be great, thanks!

Fixed

tomakehurst (approved):

oleg-nenashev (approved):
+1 for the new version, will ship it.
Thanks a lot @tomasbjerre !

PR #2256, opened 2023-07-10
Replace Guava Files management with STL (refs #2111)

tomasbjerre:
Removing imports on Guava Files and using Java standard library.

## References

#2111 

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/AbstractFileSource.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/AbstractFileSource.java
@@ -18,10 +18,10 @@ package com.github.tomakehurst.wiremock.common;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.github.tomakehurst.wiremock.security.NotAuthorisedException;
-import com.google.common.io.Files;
 import java.io.File;
 import java.io.IOException;
 import java.net.URI;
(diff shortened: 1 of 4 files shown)


oleg-nenashev (changes requested):
Thanks! When you use `Paths.get()` please make sure to catch [InvalidPathException](https://docs.oracle.com/javase/8/docs/api/java/nio/file/InvalidPathException.html) in all usages Oracle was so kind to make it a runtime exception instead of `IOException`, and we had a bunch of bugs in Jenkins because of that

tomasbjerre:
> Thanks! When you use `Paths.get()` please make sure to catch [InvalidPathException](https://docs.oracle.com/javase/8/docs/api/java/nio/file/InvalidPathException.html) in all usages Oracle was so kind to make it a runtime exception instead of `IOException`, and we had a bunch of bugs in Jenkins because of that

Good to know! When I looked at the individual changes here I cannot see that catching `InvalidPathException` would change any behavior.

In production code `AbstractFileSource` was throwing `RuntimeException` before and will do it now.

In test cases I usually just let it crash with whatever exception is thrown, it seems to me that this is the strategy here as well.

oleg-nenashev (approved):
Well, you're totally right. Current handling is not what I would expect, but indeed the change does not alter the behavior. Let's merge it, and thanks!

PR #2263, opened 2023-07-12
feat: optionally fail request if several mappings match

tomasbjerre:
This PR adds a feature that, if enabled, will fail requests if several mappings are matched.

I have found myself, several times, trying to understand why some mapping is not working and later come the the conclusion that some other mapping was serving the request. This would help me reduce time spent on such problems.

## References

https://github.com/wiremock/wiremock.org/pull/127

Would also solve #2565

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
@@ -137,5 +137,7 @@ public interface Options {
 
   boolean getTemplateEscapingDisabled();
 
-  boolean failIfMultipleMappingsMatch();
+  default boolean failIfMultipleMappingsMatch() {
+    return false;
+  }
 }
--- a/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
@@ -136,4 +136,6 @@ public interface Options {
   Set<String> getTemplatePermittedSystemKeys();
 
   boolean getTemplateEscapingDisabled();
+
+  boolean failIfMultipleMappingsMatch();
 }
--- a/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
@@ -137,7 +137,7 @@ public interface Options {
 
   boolean getTemplateEscapingDisabled();
 
-  default boolean failIfMultipleMappingsMatch() {
+  default boolean getFailIfMultipleMappingsMatch() {
     return false;
   }
 }
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
@@ -119,7 +119,7 @@ public class WireMockApp implements StubServer, Admin {
             extensions.ofType(ResponseDefinitionTransformerV2.class),
             stores.getFilesBlobStore(),
             ImmutableList.copyOf(extensions.ofType(StubLifecycleListener.class).values()));
-    stubMappings.withFailIfMultipleMappingsMatch(options.failIfMultipleMappingsMatch());
+    stubMappings.withFailIfMultipleMappingsMatch(options.getFailIfMultipleMappingsMatch());
     nearMissCalculator = new NearMissCalculator(stubMappings, requestJournal, scenarios);
     recorder =
         new Recorder(this, extensions, stores.getFilesBlobStore(), stores.getRecorderStateStore());
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
@@ -118,8 +118,8 @@ public class WireMockApp implements StubServer, Admin {
             extensions.ofType(ResponseDefinitionTransformer.class),
             extensions.ofType(ResponseDefinitionTransformerV2.class),
             stores.getFilesBlobStore(),
-            ImmutableList.copyOf(extensions.ofType(StubLifecycleListener.class).values()),
-            options.failIfMultipleMappingsMatch());
+            ImmutableList.copyOf(extensions.ofType(StubLifecycleListener.class).values()));
+    stubMappings.withFailIfMultipleMappingsMatch(options.failIfMultipleMappingsMatch());
     nearMissCalculator = new NearMissCalculator(stubMappings, requestJournal, scenarios);
     recorder =
         new Recorder(this, extensions, stores.getFilesBlobStore(), stores.getRecorderStateStore());
@@ -140,8 +140,7 @@ public class WireMockApp implements StubServer, Admin {
       Map<String, ResponseDefinitionTransformerV2> v2transformers,
       Map<String, RequestMatcherExtension> requestMatchers,
       FileSource rootFileSource,
-      Container container,
-      boolean failIfMultipleMappingsMatch) {
+      Container container) {
 
     this.stores = new DefaultStores(rootFileSource);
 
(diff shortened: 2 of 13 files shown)


src/main/java/com/github/tomakehurst/wiremock/core/Options.java
```diff
@@ -136,4 +136,6 @@ default Function<Extensions, NotMatchedRenderer> getNotMatchedRendererFactory()
   Set<String> getTemplatePermittedSystemKeys();
 
   boolean getTemplateEscapingDisabled();
+
+  boolean failIfMultipleMappingsMatch();
```
oleg-nenashev, line 140: Needs a default return if we do not want to break binary compatibility of a public interface

src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
```diff
@@ -139,7 +140,8 @@ public WireMockApp(
       Map<String, ResponseDefinitionTransformerV2> v2transformers,
       Map<String, RequestMatcherExtension> requestMatchers,
       FileSource rootFileSource,
-      Container container) {
+      Container container,
+      boolean failIfMultipleMappingsMatch) {
```
oleg-nenashev, line 144: Breaks binary compatibility too. Maybe it is time to start switching to the `public withFailIfMultipleMappingsMatch(boolean)` pattern for new fields that are not critical to set?

src/main/java/com/github/tomakehurst/wiremock/core/Options.java
oleg-nenashev: Also should be `failIfMultipleMappingsMatch()` to follow the pattern

src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
```diff
@@ -66,183 +73,203 @@ public AbstractStubMappings(
...
-                    stubMapping.isIndependentOfScenarioState()
-                        || scenarios.mappingMatchesScenarioState(stubMapping))
-            .findFirst()
-            .orElse(StubMapping.NOT_CONFIGURED);
+    final StubMapping matchingMapping = this.getMatchingMapping(request, subEvents);
 
     subEvents.forEach(initialServeEvent::appendSubEvent);
 
-    scenarios.onStubServed(matchingMapping);
+    this.scenarios.onStubServed(matchingMapping);
```
oleg-nenashev, line 89: A bunch of unrelated refactorings, but not blocking the PR because of it. I see no strict need to use `this.` in the code or in the code style though

src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
```diff
@@ -33,14 +33,16 @@ public StoreBackedStubMappings(
       Map<String, ResponseDefinitionTransformer> transformers,
       Map<String, ResponseDefinitionTransformerV2> v2transformers,
       BlobStore filesBlobStore,
-      List<StubLifecycleListener> stubLifecycleListeners) {
+      List<StubLifecycleListener> stubLifecycleListeners,
+      boolean failIfMultipleMappingsMatch) {
```
oleg-nenashev, line 37: breaking change too

src/main/java/com/github/tomakehurst/wiremock/stubbing/InMemoryStubMappings.java
```diff
@@ -22,14 +22,15 @@
 
 public class InMemoryStubMappings extends StoreBackedStubMappings {
 
-  public InMemoryStubMappings() {
+  public InMemoryStubMappings(boolean failIfMultipleMappingsMatch) {
```
oleg-nenashev, line 25: Breaking change

oleg-nenashev (changes requested):
Requesting changes, There is a bunch of API breaking changes that are IMHO not needed. QUite opposite, maybe it is better to have a `withFailIfMultipleMappingsMatch(boolean)` notation and to take a default value from a static final variable that we could make customizable later once WireMock gets full configuration engine

I also mentioned a bunch or refactorings in the code. No strong opinion about them

tomakehurst:
Let's be careful about merging this one as it's changing the most performance-sensitive part of the codebase

src/main/java/com/github/tomakehurst/wiremock/stubbing/InMemoryStubMappings.java
tomasbjerre: fixed

src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
tomasbjerre: fixed

src/main/java/com/github/tomakehurst/wiremock/core/Options.java
tomasbjerre: I think this is fixed

tomasbjerre:
Perhaps some of the API comments here would have been avoided if the things that are considered API is in its own Gradle project, as discussed in https://github.com/wiremock/wiremock/issues/2266 .

tomasbjerre:
> Requesting changes, There is a bunch of API breaking changes that are IMHO not needed. QUite opposite, maybe it is better to have a `withFailIfMultipleMappingsMatch(boolean)` notation and to take a default value from a static final variable that we could make customizable later once WireMock gets full configuration engine
> 
> I also mentioned a bunch or refactorings in the code. No strong opinion about them

I'm not sure I understand what you mean here exactly.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
oleg-nenashev: Works for me.

src/main/java/com/github/tomakehurst/wiremock/core/Options.java
```diff
@@ -136,4 +136,8 @@ default Function<Extensions, NotMatchedRenderer> getNotMatchedRendererFactory()
   Set<String> getTemplatePermittedSystemKeys();
 
   boolean getTemplateEscapingDisabled();
+
+  default boolean failIfMultipleMappingsMatch() {
```
oleg-nenashev, line 140: This class uses mixed patterns but I guess `getSomething()` dominates. Should it be ` getFailIfMultipleMappingsMatch()` . No strong opinion

oleg-nenashev (commented):
All key comments are addressed. I guess now we need a call on the performance overhead, but I have no objections otherwise

P.S: We have no perf tests to do measurements

src/main/java/com/github/tomakehurst/wiremock/core/Options.java
tomasbjerre: fixed

oleg-nenashev (approved):

ArthurCarliDkt:
Hello,
Is this PR blocked ? Is it still intended to be merged ? How can we make it progress ?
This feature would be valuable for us in our organization.

tomakehurst:
I don't think I agree with the utility of this feature for a few reasons:
* You can (per the original problem statement) determine which stub actually matched both from the response headers and the request log.
* Having overlapping stubs with varying levels of specificity + priorities is core to how WireMock works.
* If you really want to trigger test failure when there are multiple matches it's quite easy to write your own extension for this (see below)

```java
public class TmpTest {

    @RegisterExtension
    static WireMockExtension wireMockExtension = WireMockExtension
            .newInstance()
            .options(wireMockConfig().extensions(new FailOnMultipleMatches()))
            .build();

    private WireMockTestClient client;

    @BeforeEach
    void init() {
        client = new WireMockTestClient(wireMockExtension.getPort());
    }

    @Test
    void fail_on_muliple_matches() {
        stubFor(get(urlPathEqualTo("/things")).willReturn(ok("Loose match")));
        stubFor(get(urlPathEqualTo("/things")).withQueryParam("type", equalTo("games")).willReturn(ok("Tight match")));

        // Will only match the loose stub
        WireMockResponse response = client.get("/things?type=underpants");
        assertThat(response.statusCode(), is(200));

        // Will match both stubs
        response = client.get("/things?type=games");
        assertThat(response.statusCode(), is(500));
    }

    public static class FailOnMultipleMatches implements ResponseDefinitionTransformerV2 {

        @Override
        public String getName() {
            return "fail-on-multiple-matches";
        }

        @Override
        public ResponseDefinition transform(ServeEvent serveEvent) {
            long exactMatchCount = findNearMissesFor(serveEvent.getRequest()).stream()
                    .filter(nearMiss -> nearMiss.getMatchResult().isExactMatch())
                    .count();

            return exactMatchCount <= 1 ?
                    serveEvent.getResponseDefinition() :
                    responseDefinition().withStatus(500).withBody("Too many matches!").build();
        }

        @Override
        public boolean applyGlobally() {
            return true;
        }
    }
}
```

PR #2264, opened 2023-07-12
[ISSUE-2260] Support for multipart/related without Content-Dispositio…

hladysz:
Support for multipart/related without Content-Disposition headers - recording and replaying

## References

- https://github.com/wiremock/wiremock/issues/2260

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorderTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorderTest.java
@@ -421,7 +421,7 @@ public class StubMappingJsonRecorderTest {
           + "		\"url\": \"/multipart/content\",							\n"
           + "		\"multipartPatterns\" : [ {									\n"
           + "			\"name\" : \"binaryFile\",								\n"
-          + "			\"matchingType\" : \"ALL\",								\n"
+          + "			\"matchingType\" : \"ANY\",								\n"
           + "			\"headers\" : {											\n"
           + "				\"Content-Disposition\" : {							\n"
           + "					\"contains\" : \"name=\\\"binaryFile\\\"\"		\n"
@@ -432,7 +432,7 @@ public class StubMappingJsonRecorderTest {
           + "			} ]														\n"
           + "		}, {														\n"
           + "			\"name\" : \"textFile\",								\n"
-          + "			\"matchingType\" : \"ALL\",								\n"
+          + "			\"matchingType\" : \"ANY\",								\n"
           + "			\"headers\" : {											\n"
           + "				\"Content-Disposition\" : {							\n"
           + "					\"contains\" : \"name=\\\"textFile\\\"\"		\n"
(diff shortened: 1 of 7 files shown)


src/test/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorderTest.java
```diff
@@ -421,7 +421,7 @@ public void decompressesGzippedResponseBodyAndRemovesContentEncodingHeader() {
           + "		\"url\": \"/multipart/content\",							\n"
           + "		\"multipartPatterns\" : [ {									\n"
           + "			\"name\" : \"binaryFile\",								\n"
-          + "			\"matchingType\" : \"ALL\",								\n"
+          + "			\"matchingType\" : \"ANY\",								\n"
```
oleg-nenashev, line 424: Is this change really needed?

oleg-nenashev (approved):
Sounds about right from the code. I haven't tested it, but the potential impact is isolated enough

tomakehurst:
This fixes the (now deprecated) legacy recorder, which we'll be removing soon.

Does this issue affect the current recorder, and if so could we fix it there instead?

src/test/java/com/github/tomakehurst/wiremock/stubbing/StubMappingJsonRecorderTest.java
hladysz: looking

hladysz: Yes, there's an example of mis-match in `src/test/java/com/github/tomakehurst/wiremock/matching/RequestPatternTest.java#doesNotMatchExactlyWhenManyMultipartPatternsSupposeToMatchAllDistinctParts()`

hladysz:
> This fixes the (now deprecated) legacy recorder, which we'll be removing soon.
> 
> Does this issue affect the current recorder, and if so could we fix it there instead?

Thanks. Where can I find the new and improved recorder?

tomakehurst:
`com.github.tomakehurst.wiremock.recording.Recorder` is the place to start for the new one

hladysz:
> com.github.tomakehurst.wiremock.recording.Recorder is the place to start for the new one

looking.

hladysz:
> `com.github.tomakehurst.wiremock.recording.Recorder` is the place to start for the new one

This one can't be executed in a standalone version, can it?

hladysz:
> `com.github.tomakehurst.wiremock.recording.Recorder` is the place to start for the new one

If this is what I think it is (https://github.com/wiremock/wiremock/blob/5f0677f0a0fd5af009b5a834ec72602b0e55fefc/src/main/java/com/github/tomakehurst/wiremock/recording/RequestBodyAutomaticPatternFactory.java#L73) then it would be a lot of work to support multipart in there. Can I pair with someone if that needs to be done as a part of this fix?

hladysz:
Not too sure what's the significance of this one: https://github.com/wiremock/wiremock/blob/5f0677f0a0fd5af009b5a834ec72602b0e55fefc/src/main/java/com/github/tomakehurst/wiremock/recording/RequestBodyAutomaticPatternFactory.java#L73

tomakehurst:
> > `com.github.tomakehurst.wiremock.recording.Recorder` is the place to start for the new one
> 
> This one can't be executed in a standalone version, can it?

Yes it can, but you start it via an API call or from the recorder UI page, rather than via the CLI.

hladysz:
> > > `com.github.tomakehurst.wiremock.recording.Recorder` is the place to start for the new one
> > 
> > 
> > This one can't be executed in a standalone version, can it?
> 
> Yes it can, but you start it via an API call or from the recorder UI page, rather than via the CLI.

Thank you for your response.

oleg-nenashev:
@hladysz @tomakehurst I guess this PR is a no-go since the old recorder is now deprecated, right?

oleg-nenashev:
Thanks @hladysz . I will review it with the team

PR #2275, opened 2023-07-17
Fix admin request crashing when timing responseSendTime is null

emilianoalvarez91:
The Timing class has 3 private properties addedDelay, processTime and responseSendTime of time Integer, however the accessors public functions return an int type. When any of these is null it causes a crash and this happened when calling http://127.0.0.1:{port}/__admin/requests.

2023-07-17 14:57:42.948 Unrecoverable error handling admin request
wiremock.com.fasterxml.jackson.databind.JsonMappingException: Cannot invoke "java.lang.Integer.intValue()" because "this.responseSendTime" is null (through reference chain: com.github.tomakehurst.wiremock.admin.model.GetServeEventsResult["requests"]->java.util.ArrayList[8]->com.github.tomakehurst.wiremock.stubbing.ServeEvent["timing"]->com.github.tomakehurst.wiremock.common.Timing["responseSendTime"])
Caused by: java.lang.NullPointerException: Cannot invoke "java.lang.Integer.intValue()" because "this.responseSendTime" is null
	at com.github.tomakehurst.wiremock.common.Timing.getResponseSendTime(Timing.java:58)

## References
No reference

## Submitter checklist


tomakehurst:
I think it would be better just to make these fields nullable. Returning 0 seems misleading to me.

oleg-nenashev (approved):
+1 from me. Formally it is an API breaking change, but I think it is reasonable in this case

tomakehurst (approved):

PR #2280, opened 2023-07-18
Introduced contender in the dynamic scenarios.

ioanngolovko:
Implemented the demo of dynamic scenarios.

[Sluck discussion](https://wiremock-community.slack.com/archives/C053Y3N34C8/p1689588297883109)

Special notes.

1. There is no caching with ttl for generated  dynamic cases now. Suggest using [caffeine](https://github.com/ben-manes/caffeine).
2. Not all the cases are tested now. But we have at least base concepts.
3. I don't know hot to do [this](https://github.com/ioanngolovko/wiremock/commit/fb38929d50dca64f37988c6954226d5c3be99f00#diff-0a56c3890aebfa812c8cdd9edf7dede3d1db31fdd3ea88090fa47f7402775d76R37) better and independent from transformers. Any ideas to make it clear?
4. I was only able to resolve dynamic key with a PathPattern due to simplicity. Maybe someone can suggest an option for UrlPattern ?

Any of your ideas are welcome.

## References

- https://github.com/wiremock/wiremock/issues/1142

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/ResponseDefinitionBuilder.java
@@ -22,11 +22,13 @@ import com.fasterxml.jackson.databind.JsonNode;
 import com.github.tomakehurst.wiremock.common.Json;
 import com.github.tomakehurst.wiremock.extension.Parameters;
 import com.github.tomakehurst.wiremock.http.*;
+import com.github.tomakehurst.wiremock.stubbing.ServeEvent;
 import com.google.common.collect.ImmutableList;
 import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Function;
 
 public class ResponseDefinitionBuilder {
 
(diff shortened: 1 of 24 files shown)


oleg-nenashev (commented):
> How about the new transformParameterResolvers?

I would probably call them `transformParameterProviders` to follow the general approach, but otherwise it is a reasonable addition to WireMock. It would technically allow to put the rest of the implementation to an extension, right? It might be a more natural fit to have this code inside https://github.com/wiremock/wiremock-extension-state, CC @dirkbolte 


I still need some time to review and answer questions from the PR title. Ultimately, we will need @tomakehurst to weight in on new extensions and code bits, but it is definitely a great start, thanks you!

dirkbolte:
I'm not sure whether I grasped everything you want to cover.  https://github.com/wiremock/wiremock-extension-state should work together with scenarios, so you should be able to store everything you need to build dynamic responses in your scenarios. Glad if I can support you in making the extension more useful.

ioanngolovko:
> I'm not sure whether I grasped everything you want to cover. https://github.com/wiremock/wiremock-extension-state should work together with scenarios, so you should be able to store everything you need to build dynamic responses in your scenarios. Glad if I can support you in making the extension more useful.

Hi. You can destroy this PR due to https://github.com/wiremock/wiremock-extension-state is able to manage complex cases.

ioanngolovko:
https://github.com/wiremock/wiremock-extension-state is able to simulate complex scenarios with states. Therefore this PR is closed. Thank you all for participation.

PR #2288, opened 2023-07-30
Proposal for: Flaky port allocation tests

pks-1981:
Proposal for: Flaky port allocation tests
(https://github.com/wiremock/wiremock/issues/2281)

## Submitter checklist


pks-1981:
@oleg-nenashev @tomakehurst 
If you support this decision, then I will complete it to the end.

pks-1981:
@tomakehurst 
I remember what you said about modifiers...
But may I correct them in the affected tests?

oleg-nenashev (approved):
Looks okay to me. I wonder whether WireMock users are also impacted, because right now we apply solution now on tests but not to the existing `dynamicPort()` API. 

My personal preference would be to update existing API method to sync internally, want to introduce new API with respective guidelines

pks-1981:
@tomakehurst @oleg-nenashev
please look MR how it will be possible

tomakehurst:
I think this could be improved by adding some thing like `setDynamicPortAllocationStrategy` to the config object, maybe even with a system property setting the default, so that we can keep calling `.dynamicPort()` rather than changing this in 60-odd files. This would also allow us to easily toggle between the two to compare the impact of each.

I'm wondering if by storing ports we've used  in a hash map and never freeing them we're risking running out in future.

Also this won't help when test parallelism > 1 as we're storing the used ports on the heap in a single process.

PR #2299, opened 2023-08-04
Change Queue on to Deque in InMemoryRequestJournalStore

pks-1981:
Change Queue on to Deque in InMemoryRequestJournalStore and ServeEvent

https://github.com/wiremock/wiremock/issues/2296

## Submitter checklist


pks-1981:
@tomakehurst @oleg-nenashev
please look MR how it will be possible

oleg-nenashev (approved):
I would definitely more confident about such a change if we had some performance test to prove impact when recording the journal and in the getter. My gut feeling is that it is a positive improvement, so I am +1 for trying it out in the beta

pks-1981:
@tomakehurst @oleg-nenashev
MR finished. please look how it will be possible.

src/main/java/com/github/tomakehurst/wiremock/stubbing/ServeEvent.java
```diff
@@ -46,7 +46,7 @@ public class ServeEvent {
   private final LoggedResponse response;
   private final Timing timing;
 
-  private final ConcurrentLinkedQueue<SubEvent> subEvents;
+  private final ConcurrentLinkedDeque<SubEvent> subEvents;
```
tomakehurst, line 49: I don't think we should change it here since this collection is never likely to get beyond single digits in length and we never reverse it.

Also, this is a breaking change, and while we can still do this it's got me thinking that maybe we should expose this as a List on the public methods of this class rather than a Queue.

pks-1981: Done

src/test/java/com/github/tomakehurst/wiremock/ProxyAcceptanceTest.java
```diff
@@ -407,7 +406,7 @@ public void doesNotDuplicateCookieHeaders() {
     testClient.get("/duplicate/cookies", withHeader("Cookie", "session=1234"));
 
     LoggedRequest lastRequest =
-        getLast(target.find(getRequestedFor(urlEqualTo("/duplicate/cookies"))));
+        getFirst(target.find(getRequestedFor(urlEqualTo("/duplicate/cookies"))), null);
```
tomakehurst, line 409: I think this will be a breaking change for a lot of folks. How about we modify AbstractRequestJournal.getRequestsMatching(...) to reverse the results so that they'll come back in the same order as previously.

pks-1981: But we will return to what we left...
Breaking changes it's normal when update version IMHO...
And as a user, I'm waiting in "history" for more recent entries to be at the top...

tomakehurst: I mean reverse the result of the find query rather than the whole list, which would have a pretty trivial impact by comparison and only affect finding vs. fetching the whole journal.

Agree it's in principle acceptable to create breaking changes, but this one is avoidable and my instinct is that it'll annoy people.

tomakehurst:
Thanks!

PR #2303, opened 2023-08-08
Add healthcheck endpoint to WireMock Standalone

ikalu:
Add healthcheck endpoint to WireMock Standalone

#[2164](https://github.com/wiremock/wiremock/issues/2164)

#[958](https://github.com/wiremock/wiremock/pull/958)

## Submitter checklist


oleg-nenashev (commented):
Thank you! IMHO it would be nice to preserve the author in the original commits. It does not look like the author is retained unless the github UI is wrong

Regarding the change substance, I think this is something we should merge. I will discuss it with @tomakehurst once he's back

src/main/java/com/github/tomakehurst/wiremock/admin/tasks/HealthCheckTask.java
```diff
@@ -0,0 +1,59 @@
...
+import com.github.tomakehurst.wiremock.common.url.PathParams;
+import com.github.tomakehurst.wiremock.core.Admin;
+import com.github.tomakehurst.wiremock.http.ResponseDefinition;
+import com.github.tomakehurst.wiremock.stubbing.ServeEvent;
+
+public class HealthCheckTask implements AdminTask {
+  @Override
+  public ResponseDefinition execute(Admin admin, ServeEvent serveEvent, PathParams pathParams) {
+
+    return admin.isHealthy()
```
tomakehurst, line 34: Since this always returns true at the moment I suggest we get rid of the admin call altogether, along with the new base interface and method. Can always be added back in if we want to put some additional checking logic in there.

ikalu: Done. Let me know if there's any other feedback.

PR #2310, opened 2023-08-10
Network rules bug

Mahoney:
There's a bug with the network rules - rules allowing / denying by host pattern are basically ignored because we resolve the name to IP addresses before applying the rules.

Unfortunately it's hard to fix with the current architecture. In order to get the host rules to apply we need to push the resolution to IP addresses down into the individual `NetworkAddressRange` subtypes, but we then need `NetworkAddressRange` instances in the `allowed` set to behave differently to `NetworkAddressRange` instances in the `denied` set - specifically, a host that resolves to multiple IP addresses should only be included when *all* of the IPs match in an allowed rule but included when *any* of the IPs match in a denied rule.

There's also an issue around what to do if the host cannot be resolved - ideally in an `allowed` range it would not be included but in a `denied` range it would be included.

See the failing tests to understand the issues.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/NetworkAddressRange.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/NetworkAddressRange.java
@@ -21,6 +21,7 @@ import com.google.common.net.InetAddresses;
 import java.math.BigInteger;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.util.Arrays;
 import java.util.Objects;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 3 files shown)


Mahoney:
I've raised this as a diff against https://github.com/wiremock/wiremock/pull/2309 because it depends on it - obviously when that PR is merged this should be changed to merge into master, and also obviously this should remain a draft until that happens.

tomakehurst:
Fixed in 14ac072d06fb0d684070d0fca715ff19e3fc915f and 027ddafc38a1048d6bd64d896b07ee35b0268894

PR #2334, opened 2023-08-29
Exposing MappingsSource for extension point

bharatnpti:
Exposing MappingsSource for extension point

## References

https://github.com/wiremock/wiremock/issues/2331

## Submitter checklist

- [ x] Recommended: Join [WireMock Slack](https://slack.wiremock.org/) to get any help in `#help-contributing` or a project-specific channel like `#wiremock-java`
- [ x] The PR request is well described and justified, including the body and the references
- [ x] The PR title represents the desired changelog entry
- [ x] The repository's code style is followed (see the contributing guide)
- [ x] Test coverage that demonstrates that the change works as expected
- [ x] For new features, there's necessary documentation in this pull request or in a subsequent PR to [wiremock.org](https://github.com/wiremock/wiremock.org)

Documentation PR - https://github.com/wiremock/wiremock.org/pull/149

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
@@ -61,6 +61,7 @@ public class WireMockApp implements StubServer, Admin {
   private final NearMissCalculator nearMissCalculator;
   private final Recorder recorder;
   private final List<GlobalSettingsListener> globalSettingsListeners;
+  private final Map<String, MappingsLoaderExtension> mappingsLoaderExtensions;
 
   private Options options;
 
@@ -116,6 +117,7 @@ public class WireMockApp implements StubServer, Admin {
     recorder =
         new Recorder(this, extensions, stores.getFilesBlobStore(), stores.getRecorderStateStore());
     globalSettingsListeners = List.copyOf(extensions.ofType(GlobalSettingsListener.class).values());
+    this.mappingsLoaderExtensions = extensions.ofType(MappingsLoaderExtension.class);
 
     this.container = container;
     loadDefaultMappings();
@@ -124,6 +126,7 @@ public class WireMockApp implements StubServer, Admin {
   public WireMockApp(
       boolean browserProxyingEnabled,
       MappingsLoader defaultMappingsLoader,
+      Map<String, MappingsLoaderExtension> mappingsLoaderExtensions,
       MappingsSaver mappingsSaver,
       boolean requestJournalDisabled,
       Integer maxRequestJournalEntries,
@@ -137,6 +140,7 @@ public class WireMockApp implements StubServer, Admin {
 
     this.browserProxyingEnabled = browserProxyingEnabled;
     this.defaultMappingsLoader = defaultMappingsLoader;
+    this.mappingsLoaderExtensions = mappingsLoaderExtensions;
     this.mappingsSaver = mappingsSaver;
     this.settingsStore = stores.getSettingsStore();
     requestJournal =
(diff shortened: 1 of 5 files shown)


oleg-nenashev (commented):
This is a binary compatible change, and we will be able to merge it into WireMock 3.1 or so

@tomakehurst Would it make sense to start putting new classes into `org.wire,ock`?

bharatnpti:
hi @tomakehurst, @oleg-nenashev,

can you please review this and let me know if there is any feedback

oleg-nenashev (approved):
Sounds about right. We should really start adding Javadoc for the new extension points though

src/main/java/com/github/tomakehurst/wiremock/extension/MappingsSourceExtension.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.extension;
+
+import com.github.tomakehurst.wiremock.standalone.MappingsSource;
+
+public abstract class MappingsSourceExtension implements MappingsSource, Extension {}
```
tomakehurst, line 20: Suggest we make this an interface rather than abstract class so that extensions wishing to implement more than one extension point in addition to this one can do so.

src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
```diff
@@ -236,6 +240,8 @@ private List<RequestFilterV2> getV2StubRequestFilters() {
 
   private void loadDefaultMappings() {
     loadMappingsUsing(defaultMappingsLoader);
+    if (mappingsSourceExtensions != null)
```
tomakehurst, line 243: It looks like we're only loading from the sources provided via extensions, and continuing only to save to the default one.

I guess this is intentional? You probably don't want to save a stub to every source, although you might want to save it back to the one you loaded it from.

If the plan was for these to be read-only it would probably make more sense just to implement `MappingsLoader` rather than `MappingsSource`.

tomakehurst (changes requested):

src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
bharatnpti: Hi @tomakehurst,

We do not have any identifier in StupMapping to indicate the source from which a particular stub was loaded.

I would prefer to keep it read-only for now, If you are fine with this I will make the change to implement MappingsLoader rather than MappingsSource.

tomakehurst: I think that would make most sense. Seems a bit misleading to implement the save methods when there's no way they'll be called.

src/main/java/com/github/tomakehurst/wiremock/extension/MappingsSourceExtension.java
bharatnpti: changed to interface

src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
bharatnpti: changes done

src/main/java/com/github/tomakehurst/wiremock/extension/MappingsSourceExtension.java
bharatnpti: hi @tomakehurst,

the changes has been done, can you please review it now and merge

tomakehurst (approved):

tomakehurst:
Looks good, will merge.

Please could you create a PR with some documentation in [wiremock.org](https://github.com/wiremock/wiremock.org)?

bharatnpti:
Hi @tomakehurst,

thank you for merging.
Documentation PR also already raised here - https://github.com/wiremock/wiremock.org/pull/149

tomakehurst:
Sorry, totally missed that!

PR #2339, opened 2023-08-30
Stop returning 500s for unmatched path patterns

Mahoney:
The Diff rendering did not handle the `SubExpressionException` thrown in
e.g. `MatchesJsonPathPattern`. This makes it catch that exception and
output it in the diff.

## References

Reinstates https://github.com/wiremock/wiremock/pull/2336 after revert in https://github.com/wiremock/wiremock/pull/2338

## Submitter checklist


src/test/java/com/github/tomakehurst/wiremock/verification/diff/DiffTest.java
```diff
@@ -524,4 +524,24 @@ public void includeSchemeIfSpecified() {
...
+
+    assertThat(
+        diff.toString(),
+        is(
+            junitStyleDiffMessage(
+                "ANY\n" + "/thing\n" + "\n" + "$.accountNum [equalTo] 1234",
+                "ANY\n"
+                    + "/thing\n"
+                    + "\n"
+                    + "Warning: JSON path expression '$.accountNum' failed to match document '' because of error 'json string can not be null or empty'")));
```
tomakehurst, line 545: Does this look OK with the ascii table diff renderer?

Mahoney: I think so - added test here:
https://github.com/wiremock/wiremock/pull/2339/commits/c956ace0c1a6e674293b954592619270f16c98b9#diff-9f49c014a4dac088bc7f29ce321b2a663878282e1f3c7c1d4fc00fb592a39348

src/test/resources/exception-diff-sample_json-path.txt
```diff
@@ -0,0 +1,16 @@
...
+                                               =======================
+
+-----------------------------------------------------------------------------------------------------------------------
+| Closest stub                                             | Request                                                  |
+-----------------------------------------------------------------------------------------------------------------------
+                                                           |
+ANY                                                        | ANY
+/thing                                                     | /thing
+                                                           |
+$.accountNum [equalTo] 1234                                | Warning: JSON path expression '$.accountNum' failed to<<<<< Body does not match
```
tomakehurst, line 12: I feel like here, on the Request side it should just say `(empty)` rather than rendering an error message. We shouldn't really treat it as an error as when scanning a whole set of stubs matching a request with an empty body it's perfectly legit to test it against a stub that requires one.

Mahoney: OK - I've reworked it so the diff rendered outputs no body if there is no body, and the body itself if the expression does not match, which seems to be the standard way.

tomakehurst (approved):

oleg-nenashev:
Is it a regression @Mahoney ?

Mahoney:
Yes, looks like it is a regression

oleg-nenashev (approved):
Thanks! I think it does not matter too much for such an old one

Mahoney:
It's actually quite nasty - if you create a stub using this matcher, you get a 500 for any request that doesn't match a stub instead of a 404 with a nice diff.

PR #2350, opened 2023-08-31
Fix custom response templating

manofthepeace:
## References
fixes https://github.com/wiremock/wiremock/issues/2349

when manually creating a `ResponseTemplateTransformer` via `WireMockConfiguration.extentions`. It gets added to the loadedExtensions, but then gets overwritten by the put. 

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java
@@ -122,13 +122,14 @@ public class Extensions implements WireMockServices {
         new ArrayList<>(ofType(TemplateModelDataProviderExtension.class).values());
 
     if (options.getResponseTemplatingEnabled()) {
-      final ResponseTemplateTransformer responseTemplateTransformer =
-          new ResponseTemplateTransformer(
-              getTemplateEngine(),
-              options.getResponseTemplatingGlobal(),
-              getFiles(),
-              templateModelProviders);
-      loadedExtensions.put(responseTemplateTransformer.getName(), responseTemplateTransformer);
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
Looks about right. If you could create a test case for this regression, it would be awesome

tomakehurst:
Please could you run `./gradlew spotlessApply` so that the tests can be executed?

manofthepeace:
@tomakehurst done

manofthepeace:
@oleg-nenashev I have added a simple test that would fail without the PR.

tomakehurst:
Not a necessary fix per https://github.com/wiremock/wiremock/issues/2349#issuecomment-1706382406. 

Documentation will be updated soon to reflect the new, simplified way to enable templating.

oleg-nenashev:
@tomakehurst so it would be a new documented breaking change, right?

tomakehurst:
Yes, will add to the release note and update the main doc in due course.

tomakehurst:
Again, I think the docs need improving here - if you want to add custom helpers you do it via another extension point now:
https://wiremock.org/docs/extensibility/adding-template-helpers/

It shouldn't ever be necessary as an end user to new up your own `ResponseTemplateTransformer` now. If there's anything that can't be customised via an extension or config param, we should be making these available instead.

manofthepeace:
OK, I can confirm all this is working as you said right above. Thanks for that, I totally missed that doc page. I think this should be noted in the migration steps of the release page. Since before we had to create the `ResponseTemplateTransfomer`, it seems sort of natural to still create by using the api available, which I went too far in the rabbit hole to adapt to the internal stuff that possible shoudn't be touched

PR #2366, opened 2023-09-07
Fix standalone missing filename extension bug

### LINKED ISSUE #2361: Recording creates mapping files without JSON extension resulting in inability to find them on subsequent instances
### Proposal

**Expected behavior:** Record/playback functionality works as described in the docs. In other words, recording can be started and API calls can be proxied to an external service, the external service's responses can be received and the request/response pairs saved in the `mappings/` directory once recording is stopped. These mappings should be able to be played back by subsequent instances of the server, not just by the instance that recorded them.

**Actual behavior:** When using the record/playback functionality, mappings are incorrectly stored _without_ the json extension. This means that anything that requires using these mappings (e.g admin API calls to `GET /__admin/mappings` or subsequent runs when recording is NOT running) will not find the mappings. Note that the mappings are still found for the same instance that recorded them, but stopping and starting a new instance or using a new instance on a different host fail to find the mappings.

### Reproduction steps

Prerequisite: You are on wiremock version `3.0.x` (we're using the [standalone jar](https://wiremock.org/docs/standalone/java-jar/), not sure if this matters)

1. Start the wiremock server via `java -jar wiremock-standalone-<version>.jar` (we're also passing `--port 57247`, but this likely doesn't matter)
2. Start recording as described in [the docs](https://wiremock.org/docs/record-playback/). We're using Postman to make a POST request to `http://localhost:57247/__admin/recordings/start` with a body like:
```json
{
  "targetBaseUrl" : "...",
  "captureHeaders" : {
    "our-header" : { }
  },
  "persist" : true
}
```
3. After hitting a few APIs via the Wiremock server, stop recording. We're using a POST request to `http://localhost:57247/__admin/recordings/stop`.
4. Notice that the recorded mappings files in the `mappings/` directory (sibling to your `jar` file unless you've specified an alternative root directory via `root-dir`) do _not_ have a `.json` extension
5. Make a GET request to `http://localhost:57247/__admin/mappings` and notice that the mappings you recorded ARE returned
6. Stop and restart the server instance
7. Make another GET request to `http://localhost:57247/__admin/mappings` and notice that 0 mappings are returned even though they exist in `mappings`.

Here's a video depicting the issue (and solution of changing mappings files to have a json extension):

https://github.com/wiremock/wiremock/assets/8343465/37161bf0-b9c7-45cb-8a88-42d5de8cd202

### References

_No response_

tomakehurst:
Fixes #2361

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/filemaker/FilenameMaker.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/filemaker/FilenameMaker.java
@@ -37,13 +37,12 @@ public class FilenameMaker {
   private final String filenameTemplate;
 
   public FilenameMaker() {
-    this.templateEngine = defaultTemplateEngine();
-    this.filenameTemplate = DEFAULT_FILENAME_TEMPLATE + DEFAULT_EXTENSION;
+    this(null);
   }
 
   public FilenameMaker(String filenameTemplate) {
     this.templateEngine = defaultTemplateEngine();
-    this.filenameTemplate = filenameTemplate;
+    this.filenameTemplate = filenameTemplate != null ? filenameTemplate : DEFAULT_FILENAME_TEMPLATE + DEFAULT_EXTENSION;
   }
 
   public FilenameMaker(String filenameTemplate, String extension) {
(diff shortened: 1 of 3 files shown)


oleg-nenashev (approved):
Yeah, this is a good change 👍

Issue #2373, opened 2023-09-13
Move Webhooks extension into the core


tomakehurst:
### Proposal

Due mostly to side effects of JAR shading the webhooks extension has proved brittle and prone to breaking in ways that are hard to detect in the core test suite.

The extension itself is small and doesn't use any dependencies that aren't already present in WireMock. Given these points and its popularity it seems sensible to move it into the core.

The following changes would also simplify and improve the dev experience:
* Enable it by default (since it's activated via specific data on individual stubs this shouldn't be an issue for non-webhook users)
* Add `WebhookTransformer` as a new extension type so that it can be applied without having to new up the extension yourself.
* Make it implement the new `ServeEventListener` interface in addition the the legacy `PostServeAction` so that it can be activated either way.
* Add sub-events containing the details of the request and response made by the extension (note, this might be tricky when we're using external `Stores` implementations since webhooks are triggered asynchronously and could thus generate the sub-event after completion of the main response and thus the last point at which the serve event is saved).

oleg-nenashev:
Fine with me

oleg-nenashev:
@tomakehurst Just to double check, do we want to move it now before the WireMock 3 announcements? I would guess so, and I am rather +1 even if this is a formally breaking change for those who depend on the WireMock Webhook Extension artifact and will need to adapt their code

tomakehurst:
Yeah, it's still broken in some contexts so I'm happy for it to be a technically breaking change if it fixes the issue once and for all.

I've nearly finished implementing this so let's leave the announcement until Friday, then hopefully this will be done and included.

tomakehurst:
Completed in 81dd277ebc45ddfb485d71544d9825ebdead19af

PR #2376, opened 2023-09-14
Move webhooks to core

tomakehurst:
Implements #2373.

Removes the webhooks extension as a separate sub-project and moves it to the core codebase.

Although it remains implemented as an extension, it is enabled at startup automatically so there's no need to explicitly enable it.

It now also implements `ServeEventListener` in addition to `PostServeAction` so both the old and new forms can be used.


oleg-nenashev:
Do you plan to submit a documentation pull request before merging? It looks like there is quite a lot of updates needed there

tomakehurst:
I guess I probably should 🙂

tomakehurst:
Docs: https://github.com/wiremock/wiremock.org/pull/174

src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java
```diff
@@ -132,6 +135,14 @@ private void configureTemplating() {
     }
   }
 
+  private void configureWebhooks() {
+    final List<WebhookTransformer> webhookTransformers =
+        ofType(WebhookTransformer.class).values().stream().collect(Collectors.toUnmodifiableList());
+
+    final Webhooks webhooks = new Webhooks(webhookTransformers, options.getProxyTargetRules());
+    loadedExtensions.put(webhooks.getName(), webhooks);
```
Mahoney, line 143: This would of course overwrite any existing extension of name `webhook`, but I guess you had thought of that and are OK with it.

src/main/java/org/wiremock/webhooks/Webhooks.java
Mahoney, line 1: I'm probably misunderstanding, but it looks like the WebHook will be sent twice per matched request? Once by this path:
```
Webhooks.doAction(ServeEvent, Admin, Parameters)  (org.wiremock.webhooks)
StubRequestHandler.triggerPostServeActions(ServeEvent)  (com.github.tomakehurst.wiremock.http)
StubRequestHandler.afterResponseSent(ServeEvent, Response)  (com.github.tomakehurst.wiremock.http)
```
and once by this:
```
Webhooks.triggerWebhook(ServeEvent, Parameters)  (org.wiremock.webhooks)
Webhooks.afterComplete(ServeEvent, Parameters)  (org.wiremock.webhooks)
ServeEventListener.onEvent(RequestPhase, ServeEvent, Parameters)  (com.github.tomakehurst.wiremock.extension)
StubRequestHandler.triggerListeners(RequestPhase, ServeEvent)(2 usages)  (com.github.tomakehurst.wiremock.http)
StubRequestHandler.afterResponseSent(ServeEvent, Response)  (com.github.tomakehurst.wiremock.http)
```

Mahoney: Oh, I see - it's just two different ways of configuring the WebHook.

It looks like `withServeEventListener` is now the preferred way, and `withPostServeAction` is deprecated. And if so, should we document that there is a new preferred way and add an `@Deprecated` annotation to `MappingBuilder.withPostServeAction`?

Mahoney (approved):
Two tiny queries but basically seems fine

src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java
tomakehurst: Yes, the idea is that you shouldn't have to do it yourself any more.

It takes the same network rules as the proxy from the options object and there's a new extension point for `WebhookTransformer` so no need to new it up yourself.

src/main/java/org/wiremock/webhooks/Webhooks.java
tomakehurst: Good point - the class is deprecated but not the method, so I'll do that in another PR.

PR #2377, opened 2023-09-14
Fixed #2364 - scenarios produced by recorder play back in reverse order

### LINKED ISSUE #2364: Scenarios unexpectedly play back in reverse order
### Proposal

**Expected behavior:** When recording a series of identical requests with the default `repeatsAsScenarios` setting (`true`), playing _back_ the scenario after recording has been stopped should play back the responses in the same order they were recorded.

**Actual behavior:** When recording a series of identical requests with the default `repeatsAsScenarios` setting (`true`), playing _back_ the scenario after recording has been stopped plays back the scenario in _reverse_ order. In other words, the mapping that _completes_ the scenario is incorrectly configured as the first one to run after the scenario state is `started`.

**Additional context:** In our case, we have an API that we poll when a test user's session is being created. When the session is still being created, the response of that API will be:

```json
{"id":"89f93e1f-28c6-452d-b102-ee4c74252de4","processing":true}
```

Once the session has been created, a new request to that endpoint will return the following:
```json
{"id":"89f93e1f-28c6-452d-b102-ee4c74252de4","processing":false, "username": "test_username", "password": "test_password"}
```

We've recorded this scenario, and what's puzzling us is that the _last_ request has the following configuration:

```json
"scenarioName": "scenario-1-demo-sessions-bf0b9386-ee34-487a-a9b0-645664cb4a77",
"requiredScenarioState": "Started",
"newScenarioState": "scenario-1-demo-sessions-bf0b9386-ee34-487a-a9b0-645664cb4a77-2"
```
In other words, the request/response mapping that occurred _last_ during record is being placed as the _first_ mapping in the scenario.

Aside from being able to tell that it is the last request's response from the fact that the body has `processing: false`, the `Date` header in the response also tells the same story. On the last response (but incorrectly first in the scenario), the `Date` is `Thu, 07 Sep 2023 16:06:47 GMT`, but the FIRST response (incorrectly _last_ in the scenario) has a `Date` of `Thu, 07 Sep 2023 16:06:28 GMT` (19 seconds _prior_ to the response being played back first).

### Reproduction steps

This should be easily reproducible using the `Date` header as described in the **Proposal** section above. In addition to our enterprise project, I was able to reproduce this with the sample project [here](https://jsonplaceholder.typicode.com/) which was originally created to reproduce https://github.com/wiremock/wiremock/issues/2361.

**Steps:**

1. Start up a wiremock server
2. Start recording with a `POST` to the Admin API `/recordings/start` endpoint with a body like:
```json
{
  "targetBaseUrl" : "https://jsonplaceholder.typicode.com",
  "persist" : true,
  "repeatsAsScenarios" : true
}
```
3. Use `curl` against the wiremock server to `GET` `http://localhost:<port>/todos/1` **more than one time.** Doing so will create a scenario
4. Stop recording via `/recordings/stop` Admin API endpoint
5. Hit the `/scenarios` Admin API endpoint to get all scenarios
6. Note the `Date` header of the FIRST mapping in the scenario (i.e the one with `requiredScenarioState` as `Started`) and see that it is _after_ the LAST mapping in the scenario (i.e the one that does not have a `newScenarioState`.

### References

_No response_

tomakehurst:


oleg-nenashev (approved):
We could tune performance later, looks like a valid fix

tomakehurst:
You think there's a performance issue @oleg-nenashev ?

oleg-nenashev:
@tomakehurst not an issue given this is not the recurrent functionality

Issue #2268, opened 2023-07-14
Build and provide 2.x maintenance releases


ryru:
### Summary

The latest version of wiremock-jre8 is 2.35 from November 2022. Can you please provide regular maintenance releases for the current stable version 2.x?

My request is specifically about the json-smart dependency, which has been fixed in https://github.com/wiremock/wiremock/pull/2252.

Many thanks for your work on wiremock!

### References

https://github.com/wiremock/wiremock/pull/2252

oleg-nenashev:
Hello @ryru . Please see https://github.com/wiremock/wiremock/issues/2135 and my note there that is yet to be added to the documentation. To understand the request better, what is the problem that needs addressing? Is it JSON Depth limit addressed in https://github.com/netplex/json-smart-v2/pull/156?

ryru:
Hi @oleg-nenashev, #2135 describes my request quite well. The wiremock version mentioned above only specifies the json-smart version 2.4.8 as a dependency. This forces dependency resolution to this very json-smart version in test packages. Which in turn triggers security scanners e.g. https://security.snyk.io/vuln/SNYK-JAVA-NETMINIDEV-3369748.

Because version 3 is still labeled 'beta', I think for the majority of wiremock users this is not an option at the moment. My suggestion is to regularly publish maintenance releases for the 2.x version branch to keep wiremock dependencies up to date. Or, as in this case with json-smart, maybe at least open the fixed version `2.4.8` dependency to a range e.g. `[2.4.8,2.4.11]`.

In my understanding this would also correspond to the ideas in #2135. Thanks in advance!

oleg-nenashev:
@ryru While the recorded CVSS 7.5 is quite high, in this case I would argue that it is much lower for a developer tool that is not a public web service. There is no confidentiality or integrity risk in this issue, so, wearing my open source maintainer hat, I would put the risk as low for the vast majority of WireMock use-cases. 

I will discuss the actions with @tomakehurst , but I think this issue alone does not justify the 2.x maintenance release and all the overheads of setting up the release flow for it

PR #2382, opened 2023-09-18
Update 2.x dependencies to latest patch releases

ryru:
The main motivation is to reduce the number of security dependency scanner findings. This adaptation is very conservative and only updates patch versions. It would also be conceivable to update minor versions of the dependencies. What do you think?

## References

- #2268

## Submitter checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,12 +26,12 @@ group = 'com.github.tomakehurst'
 project.ext {
   versions = [
     handlebars     : '4.3.1',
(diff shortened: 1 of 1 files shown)


ryru:
As discussed on [Slack](https://wiremock-community.slack.com/archives/C053Y3N34C8/p1695036640696469) dependency patches to 2.x are only applied on concrete security vulnerabilities. Since WireMock version 3 is not in beta anymore the better approach for most users would be to upgrade to the new release.

PR #2401, opened 2023-09-28
Favour host allow rules in network address rules

Mahoney:
If network address rules have conflicting IP deny and Host allow rules, the Host allow rule wins.

This allows blanket banning all private IP addresses, but then selectively allowing access by hostname.

This saves the configurer needing to know what IP address allowed services will resolve to.

Note - while there is no API change, the functionality has changed, that is unchanged configuration may now have different outcomes. I anticipate this will affect few if any existing users, but needs discussion.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/NetworkAddressRules.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/NetworkAddressRules.java
@@ -29,32 +29,29 @@ public class NetworkAddressRules {
     return new Builder();
   }
 
-  private final Set<NetworkAddressRange> allowed;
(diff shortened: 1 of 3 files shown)


Mahoney:
Just pushed a commit demonstrating that this approach will not have the desired effect.

By explicitly _allowing_ a hostname pattern, we implicitly _deny_ all other hostnames.

Which is not the desired behaviour - I was hoping to allow all hostnames as a basic case, then deny any hosts that resolved to private IP addresses, apart from a select few I would explicitly allow.

tomakehurst:
Obsolete

Issue #2397, opened 2023-09-27
Allow file names in mappings/ to end in .json5


masi:
### Proposal

WireMock already understands stubs in human-friendly JSON5 format. To make it easier for IDEs to do proper syntax highlighting allow files ending in .json5 within `mappings/` to clarify the file format.

What needs to be changed is that the file filter accepts both .json and .json5.

PR #2405, opened 2023-09-30
allow file names in mappings to end in .json5

samueldcs:
Addresses #2397.
WireMock already supports json5, but files ending in extensions other than `json` are ignored. I've changed this behavior to also load json5 files, both from remote and local sources.

## References

- TODO

## Submitter checklist

### DIFF
--- a/src/test/resources/classpath-filesource/mappings/subdir/new.json5
+++ b/src/test/resources/classpath-filesource/mappings/subdir/new.json5
@@ -1,0 +1,11 @@
+{
+  "request": {
+    "method": "GET",
+    "url": "/json5_test"
+  },
+  "response": {
+    // a JSON comment, which is only allowed in JSON5
+    "body": 'json5 testing', // a property in single quotes
+    "status": 200
+  }
+}
--- a/src/main/java/com/github/tomakehurst/wiremock/common/AbstractFileSource.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/AbstractFileSource.java
@@ -25,6 +25,7 @@ import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.List;
 import java.util.Optional;
 import java.util.function.Predicate;
@@ -191,6 +192,11 @@ public abstract class AbstractFileSource implements FileSource {
   }
 
   public static Predicate<BinaryFile> byFileExtension(final String extension) {
-    return input -> input.name().endsWith("." + extension);
+    return inputFile -> inputFile.name().endsWith("." + extension);
+  }
+
+  public static Predicate<BinaryFile> byFileExtensions(final Collection<String> extensions) {
+    return inputFile ->
+        extensions.stream().anyMatch(extension -> byFileExtension(extension).test(inputFile));
   }
 }
--- a/src/main/java/com/github/tomakehurst/wiremock/common/ContentTypes.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/ContentTypes.java
@@ -59,6 +59,8 @@ public class ContentTypes {
   public static final List<String> TEXT_FILE_EXTENSIONS =
       asList("txt", "json", "xml", "html", "htm", "yaml", "csv");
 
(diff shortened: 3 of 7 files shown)


src/test/resources/classpath-filesource/mappings/subdir/new.json5
tomakehurst, line 1: Could we expand this to confirm support for all the JSON5 documented [at json5.org](https://json5.org)?

samueldcs: Uh-oh! It doesn't seem every feature is supported with the [current ObjectMapper configuration](https://github.com/wiremock/wiremock/blob/aadd685927729db9d8c7b5797282d7e7aa0bb62d/src/main/java/com/github/tomakehurst/wiremock/common/Json.java).
A quick read through Jackson's documentation seems to indicate that most (though not all) JSON5 features are supported. Is it okay if I mess with WireMock's ObjectMapper to enable the missing options, and maybe document the current limitations?

tomakehurst: Yeah, I think let's write tests for each and enable everything we can. IIRC it's only one or two things that aren't supported by Jackson.

samueldcs: I've covered all the JSON5 features I could get working. A few notes:
- In accordance to [this StackOverflow answer](https://stackoverflow.com/a/76145609), I've found that quite a few JSON5 features are not yet supported by Jackson. I couldn't find any kind of roadmap suggesting that'll change anytime soon 
- Perhaps we should upgrade from configuring JsonParser.Feature to enabling JsonReadFeature [according to the jackson docs](https://github.com/FasterXML/jackson-core/wiki/JsonReadFeatures#:~:text=but%20recommendation%20is%20to%20use%20newer%20set%20of%20features%20to%20ease%20migration%20to%203.0), but I guess that's another PR

tomakehurst: I feel like these are likely to be lesser used features. I think it would be reasonable to make a note in the docs about these caveats but still claim support for JSON5.

tomakehurst:
Please can you do a `./gradlew spotlessApply` and push?

samueldcs:
> Please can you do a `./gradlew spotlessApply` and push?

Uhhh nothing appears to have changed.. might be related to https://github.com/wiremock/wiremock/issues/2333?

oleg-nenashev:
@samueldcs Maybe :(  I am not a huge fan of spotless, but fixing it until it passes seems to be the only way. Or one could add a github action that applies Spotless to an arbitrary pull request (with a risk of royal mess if it goes wrong)

src/test/java/com/github/tomakehurst/wiremock/matching/MatchesXPathPatternTest.java
```diff
@@ -320,7 +320,7 @@ void reportsErrorFromSubMatcher() {
             .match("<something>{ bad json</something>");
     checkJsonError(
         matchResult,
-        "Unexpected character ('b' (code 98)): was expecting double-quote to start field name\n at [Source: (String)\"{ bad json\"; line: 1, column: 4]");
+        "Unexpected character ('j' (code 106)): was expecting a colon to separate field name and value\n at [Source: (String)\"{ bad json\"; line: 1, column: 8]");
```
tomakehurst, line 323: I think we shouldn't change the way we parse user-supplied JSON in requests/responses, just stub mappings. I think maintaining a separate parser configuration and switching into it in the mappings loader (and possibly the API) might be the way forward here.

leeturner:
Hi @samueldcs I hope you are well.  I was just wondering if you are still open to working on this PR?  It seems that there is just one final bit left and that is Tom's final comment about maintaining a separate parser configuration.  I think this would be a great addition to WireMock especially when we are creating examples for users.  Being able to add comments to stub mappings would be great.

No worries if you don't currently have time.  If that is the case, would you be open to others picking this up and moving it forwards?

TWiStErRob:
I would love to have this. IDEAs don't nicely support mixing JSON and JSON5 content in .json files. :(

Issue #2393, opened 2023-09-21
A Stubmapping can be removed by ID


burcuakkayaa:
### Proposal

Hi,
I created multiple stubmappings with a same url and define spesific Id with using '.withId(specificStubMockId)' function on MappingBuilder.
Sometimes I need to remove stubs in function and if can we remove stubs with Id it can be more usable for it.
right now I used it 'WireMock.removeStub(WireMock.post("").withId(specificStubMockId))'  but I hope your team will be provide better experience. thank you

tomakehurst:
There's already an API call to do this (`DELETE /__admin/mappings/<uuid>`), but it's not exposed via the Java DSL so this would be a good thing to add.

It should be pretty easy to do if you feel like creating a PR?

PR #2408, opened 2023-10-02
A Stubmapping can be removed by ID #2393

burcuakkayaa:
## References

Each stubMapping can be removed by UUID

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -258,7 +258,8 @@ public class WireMockServer implements Container, Stubbing, Admin {
   public void removeStub(StubMapping stubMapping) {
     client.removeStubMapping(stubMapping);
   }
-
+  @Override
+  public void removeStub(UUID id) { client.removeStubMapping(id); }
   @Override
   public List<StubMapping> getStubMappings() {
     return client.allStubMappings().getMappings();
--- a/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
@@ -259,7 +259,10 @@ public class WireMockServer implements Container, Stubbing, Admin {
     client.removeStubMapping(stubMapping);
   }
   @Override
-  public void removeStub(UUID id) { client.removeStubMapping(id); }
+  public void removeStub(UUID id) {
+    client.removeStubMapping(id);
+  }
+
   @Override
   public List<StubMapping> getStubMappings() {
     return client.allStubMappings().getMappings();
--- a/src/main/java/com/github/tomakehurst/wiremock/junit/Stubbing.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit/Stubbing.java
@@ -38,6 +38,8 @@ public interface Stubbing {
 
   void removeStub(StubMapping mappingBuilder);
 
+  void removeStub(UUID id);
+
   List<StubMapping> getStubMappings();
 
   StubMapping getSingleStubMapping(UUID id);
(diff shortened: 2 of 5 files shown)


tomakehurst:
Can you add a couple of acceptance tests?

One happy path and one where the ID isn't found at a minimum would be good.

burcuakkayaa:
Hi @tomakehurst ,
I added to acceptance test, could you review them?
thanks!

src/main/java/com/github/tomakehurst/wiremock/WireMockServer.java
```diff
@@ -258,7 +258,8 @@ public void removeStub(MappingBuilder mappingBuilder) {
   public void removeStub(StubMapping stubMapping) {
     client.removeStubMapping(stubMapping);
   }
-
+  @Override
+  public void removeStub(UUID id) { client.removeStubMapping(id); }
```
oleg-nenashev, line 262:
```suggestion
  public void removeStub(UUID id) {
    client.removeStubMapping(id);
  }

```

src/main/java/com/github/tomakehurst/wiremock/junit/Stubbing.java
```diff
@@ -38,6 +38,8 @@ public interface Stubbing {
 
   void removeStub(StubMapping mappingBuilder);
 
+  void removeStub(UUID id);
```
oleg-nenashev, line 41: Without a default implementation, this is a binary compatibility breaking change.
You can add a default implementation by chaining `getSingleStubMapping(id)` and the existing removal method.

P.S: no strong opinion, but maybe the method should be actually returning the removed stubbing or `null` if it does not exist

oleg-nenashev (changes requested):
Hello, thanks for the pull request! I am requesting a change in the interface, because right now it introduces a compatibility breaking change that might impact downstream usages and extensions

src/test/java/com/github/tomakehurst/wiremock/RemoveStubMappingAcceptanceTest.java
```diff
@@ -135,6 +135,55 @@ public void removeStubThatDoesNotExists() {
...
+
+    assertThat(getMatchingStubCount("/stub-1", "/stub-2"), is(1));
+
+    removeStub(id2);
+
+    assertThat(getMatchingStubCount("/stub-1", "/stub-2"), is(0));
+  }
+
+  @Test
+  public void removeStubWithUUIDThatDoesNotExists() {
```
tomakehurst, line 164: Likewise I think this could be much simpler - really just:
1. Add a stub so that there's one in the list
2. Generate a random ID
3. Remove it
4. Assert that no exception was thrown
5. Assert that there's 1 stub in the list

oleg-nenashev:
My comment still needs to be addressed

src/main/java/com/github/tomakehurst/wiremock/junit/Stubbing.java
tomakehurst: Not worried about binary compatibility because it should only affect implementers of the `Stubbing` interface, and I can't think of a reason for anyone else to do that.

Agree returning the deleted stub or null would be a better interface design (albeit not consistent with the existing one), so let's do that.

tomakehurst:
Still a couple of outstanding comments to address:
1. Returning the deleted stub
2. Asserting that no exception is thrown when deleting a non-existent stub

Looking good other than that.

tomakehurst:
Hey @burcuakkayaa I think we're very close on this one. If you've got some capacity to look at the last couple of comments, then I'll be happy to merge this.

leeturner:
Hi @burcuakkayaa This is looking really nice and really close to being finished.  Is there anything I can support with to get this over the line ?

leeturner:
Hi @burcuakkayaa just to let you know that we are looking to push a new release of Wiremock next week so we might merge this PR as it is and then look to make the small update to the tests we have been discussing in this PR.  I hope that is OK with you?  Obviously if you wanted to make any of those changes before then we should be able to squeeze them in

PR #2423, opened 2023-10-07
Do not use deprecated jackson iso8601 class

SimonVerhoeven:
The currently used Jackson ISO8601 class is deprecated

## References

- TODO

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/admin/Conversions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/admin/Conversions.java
@@ -15,11 +15,12 @@
  */
 package com.github.tomakehurst.wiremock.admin;
 
-import com.fasterxml.jackson.databind.util.ISO8601DateFormat;
 import com.github.tomakehurst.wiremock.common.Errors;
 import com.github.tomakehurst.wiremock.common.InvalidInputException;
 import com.github.tomakehurst.wiremock.http.QueryParameter;
+
 import java.text.ParseException;
+import java.text.SimpleDateFormat;
 import java.util.Date;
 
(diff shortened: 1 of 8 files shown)


oleg-nenashev (approved):
LGTM, thanks!

src/main/java/com/github/tomakehurst/wiremock/admin/Conversions.java
```diff
@@ -30,7 +30,9 @@ public static Integer toInt(QueryParameter parameter) {
 
   public static Date toDate(QueryParameter parameter) {
     try {
-      return parameter.isPresent() ? new ISO8601DateFormat().parse(parameter.firstValue()) : null;
+      return parameter.isPresent()
+          ? new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss'Z'").parse(parameter.firstValue())
```
tomakehurst, line 34: Are we changing the contract here?

I think the class we're replacing will allow milliseconds whereas it looks like this won't.

SimonVerhoeven: Ah fair point, I looked upon the usages and based myself upon the usage for ParseDateHelper which does not use ms. Do you prefer a singular contract or optional ms?

tomakehurst:
How about we use `java.time` instead?

i.e. for formatting:

```java
DateTimeFormatter.ISO_ZONED_DATE_TIME.format(date.toInstant().atZone(ZoneId.of('Z')));
```

parsing:
```java
Date.from(ZonedDateTime.parse(isoDateTimeString).toInstant())
```

PR #2438, opened 2023-10-11
feat: add helper to generate fake data using data-faker

Shreya-7:
This helper will enable users to add fake, random data to their stub responses.

- Used https://github.com/datafaker-net/datafaker instead of https://github.com/DiUS/java-faker (as mentioned in the issue) as the former is a maintained fork of the latter.
- Used the `expression` method instead of `resolve` (as suggested in the issue) as the `resolve` method is not able to resolve nested references in the yml files. For example, in the [resource file for en-US](https://github.com/datafaker-net/datafaker/blob/main/src/main/resources/en-US.yml), `address.full_address` wasn't resolving because of the `#{street_address}` nested reference, but `address.postcode_by_state.AL` would work fine. `expression` is able to handle all cases.

## References

Closes https://github.com/wiremock/wiremock/issues/2412

## Submitter checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -98,6 +98,8 @@ dependencies {
 
   api 'com.networknt:json-schema-validator:1.0.87'
 
+  implementation 'net.datafaker:datafaker:2.0.2'
+
   testImplementation "junit:junit:4.13"
   testImplementation("org.junit.jupiter:junit-jupiter:$versions.junitJupiter")
   testImplementation("org.junit.platform:junit-platform-testkit")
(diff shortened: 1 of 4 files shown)


tomakehurst:
Hi @Shreya-7, thanks for submitting this.

My understanding was that datafaker is Java 17+, which would be an issue if true since WireMock supports Java 11+.

Also, was thinking we'd build this as an extension project initially, rather than a core feature due to the size of the faker library?

Shreya-7:
@tomakehurst I hadn't considered the Java version compatibility but datafaker has a build (1.x) to support Java 8 as well. We can switch over to the master build once we also go Java 17+. Does that sound good to you? Granted that 1.x is not maintained but it is in a better state than java-faker, which is also not maintained and seems the latest build for it is also broken due to the snakeyaml dep.

On the extensions part, yeah I checked the jar size and it came out to be 2.4 mb, will convert this to an extension.

tomakehurst:
I agree, using datafaker 1.x sounds like the best plan. Let's go with that.

Shreya-7:
Sounds good! Closing out this PR in favour of the new extension repository.

Issue #2414, opened 2023-10-04
GET_OR_HEAD method


tomakehurst:
### Proposal

A very common behaviour in real web servers is for a HEAD request to return the same header info as for the equivalent GET request, minus the body. This often includes the Content-Length.

Add a `GET_OR_HEAD` method, which will match either method and return the correct header information when `HEAD` is used.

tomakehurst:
Sure, please do!

Here are the contributing guidelines if you've not already read them:
https://github.com/wiremock/wiremock/blob/master/CONTRIBUTING.md

PR #2440, opened 2023-10-12
Implement Get or Head method

rafu01:
Implemented `GET_OR_HEAD` method as the requested issue. Feedback is welcome as this is my first contribution here, might need to change a few things.
#2414 

## Submitter checklist


src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
```diff
@@ -561,6 +563,15 @@ public static MappingBuilder head(UrlPattern urlPattern) {
     return new BasicMappingBuilder(RequestMethod.HEAD, urlPattern);
   }
 
+  public static MappingBuilder getOrHead(String method, UrlPattern urlPattern) {
```
tomakehurst, line 566: I don't think this is the right approach - we should be building the stub mapping with the GET_OR_HEAD request method, then when matching it should match requests with either method and serve the body or not as appropriate.

leeturner:
Hi, I think this was implemented in this PR - https://github.com/wiremock/wiremock/pull/2555

Perhaps we could close this one now ?

PR #2445, opened 2023-10-14
Add parsing for multi-valued query params in non standard format - Resolves #2224 [#hacktoberfest]

### LINKED ISSUE #2224: Expand support for various multi-value query parameter formats
### Proposal

Right now Wiremock only supports multi-value query parameters of the following format:
```
?id=1&id=2&id=3
```
That is, passing multiple values by using the same key.

However, since there is no standard dictating the format of multi-value parameters many services expect such parameters in different forms. A few examples:
```
?id[0]=1&id[1]=2&id[2]=3
?id=1,2,3
?id=1|2|3
?id=[1,2,3]
```
Support could be added so that, via a command-line argument, the user can choose either from a list of pre-existing formats or choose their own delimiters and separators. For instance:
```
# ?id[0]=1&id[1]=2&id[2]=3
java -jar wiremock-standalone.jar --multi-value-format index

# ?id=1,2,3
java -jar wiremock-standalone.jar --multi-value-format comma

# ?id=[1,2,3]
java -jar wiremock-standalone.jar --multi-value-format json

# ?id={1-2-3}
java -jar wiremock-standalone.jar --multi-value-parameter-separator "-" --multi-value-parameter-delimiters "{}"
```

Alternatively, for reasons of simplicity, some commonly used multi-value parameter formats such as the ones listed above could be supported by default in order to reduce the input needed by the user without the support for custom delimiters and separators.

### References

Documentation for the query-string TypeScript library taking a similar approach to multi-value query parameters: https://www.npmjs.com/package/query-string

Original request for support of multi-value parameters using duplicate keys: https://github.com/wiremock/wiremock/issues/398

kapishmalik:
Description:

This PR adds parsing for query params in non standard format. It will be able to parse following non-standard multi-valued query parameter formats along with standard one.

?id[0]=1&id[1]=2&id[2]=3
?id=1,2,3
?id=1|2|3
?id=[1,2,3]

## References
https://github.com/wiremock/wiremock/issues/2224

## Submitter checklist

#hacktoberfest

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
@@ -19,9 +19,6 @@ import static com.github.tomakehurst.wiremock.common.Exceptions.throwUnchecked;
 import static java.nio.charset.StandardCharsets.UTF_8;
 
 import com.github.tomakehurst.wiremock.http.QueryParameter;
-import com.google.common.collect.ImmutableListMultimap;
-import com.google.common.collect.ImmutableListMultimap.Builder;
-import com.google.common.collect.Maps;
 import java.net.MalformedURLException;
 import java.net.URI;
 import java.net.URL;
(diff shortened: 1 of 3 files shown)


Mahoney:
@kapishmalik did you see this comment on the original issue?
https://github.com/wiremock/wiremock/issues/2224#issuecomment-1595819768

kapishmalik:
@Mahoney  yup I saw that comment but if we are adding support for all without config does that harm any sense? IMO, it will be complete overhead of doing that. @tomakehurst  what do you want to say on this?

tomakehurst:
Looking good so far @kapishmalik, a few observations:
* Would be good to see some negative tests. I suspect there are some cases that'll make this code blow up e.g. a decimal point in the index.
* What about matching? The change to `Urls` should also permit matching query params split this way, but some acceptance test cases showing it would be valuable.
* Concur with @Mahoney's about https://github.com/wiremock/wiremock/issues/2224#issuecomment-1595819768

tomakehurst:
The main issue with just enabling this behaviour without a config flag is that it'll be a breaking change for anyone who depends on the current behaviour.

I think we could just make it boolean e.g. `enableAdvancedQueryParameterParsing`.

kapishmalik:
@tomakehurst getting hold of option enableAdvancedQueryParameterParsing from Diff class will be pain as method com.github.tomakehurst.wiremock.common.Urls#splitQuery(java.net.URI) is called from here as well and IMO, it will not be worth it to change and break this at multiple places. I believe, it will be more beneficial to ask end users to change as they are using this behaviour incorrectly. I am okay to go ahead with any of the approaches whatever you guys suggest.

tomakehurst:
What if we did the splitting in the `Request` implementation, rather than in the matching/templating system?

We could extend `QueryParameter` so that it explicitly modelled these cases. Then if the toggle is enabled we use the enhanced parsing in `WireMockHttpServletRequestAdapter` and everything that consumes query parameters gets this enhanced model.

PR #2448, opened 2023-10-16
Resolves #618 - Add ability to name specific wiremock instance

### LINKED ISSUE #618: Add ability to name specific Wiremock instances
It would be really useful to have the option to name a given Wiremock instance, so that error logs can become more descriptive.

For example:

```public WireMockRule customerService = new WireMockRule(options().name("Customer Service");```

which would result in this line in the logs:
```2017-03-01 13:23:53.651 ERROR 6984 --- [qtp260084831-27] WireMock Customer Service                                : Request was not matched:```

### PR DESCRIPTION
**Pull request by kapishmalik**
Added ability to name specific wiremock instance.
## References
https://github.com/wiremock/wiremock/issues/618

## Submitter checklist

- [ ] For new features, there's necessary documentation in this pull request or in a subsequent PR to [wiremock.org]

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
@@ -61,6 +61,8 @@ public class WireMockConfiguration implements Options {
 
   private long asyncResponseTimeout = DEFAULT_TIMEOUT;
   private boolean disableOptimizeXmlFactoriesLoading = false;
+
+  private String serverName = DEFAULT_SERVER_NAME;
   private int portNumber = DEFAULT_PORT;
   private boolean httpDisabled = false;
   private String bindAddress = DEFAULT_BIND_ADDRESS;
@@ -168,6 +170,12 @@ public class WireMockConfiguration implements Options {
 
   public WireMockConfiguration timeout(int timeout) {
     this.asyncResponseTimeout = timeout;
+    return this;
+  }
+
+  public WireMockConfiguration serverName(String serverName) {
+
+    this.serverName = serverName;
     return this;
   }
 
@@ -541,6 +549,12 @@ public class WireMockConfiguration implements Options {
   public WireMockConfiguration withMaxTemplateCacheEntries(Long maxTemplateCacheEntries) {
     this.maxTemplateCacheEntries = maxTemplateCacheEntries;
     return this;
+  }
+
+  @Override
+  public String getServerName() {
+
+    return serverName;
   }
 
   @Override
--- a/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/core/Options.java
@@ -32,6 +32,7 @@ import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.function.Function;
+import org.apache.commons.lang3.StringUtils;
 
 public interface Options {
 
@@ -46,6 +47,10 @@ public interface Options {
   int DEFAULT_TIMEOUT = 300_000;
   int DEFAULT_CONTAINER_THREADS = 25;
   String DEFAULT_BIND_ADDRESS = "0.0.0.0";
+
+  String DEFAULT_SERVER_NAME = StringUtils.EMPTY;
+
+  String getServerName();
 
   int portNumber();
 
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by tomakehurst** [comment]
This is looking good so far but let's add add a test that checks the name is actually emitted in logs.

Also, what might be interesting is to also have it affect the `Slf4jNotifier` that's used when starting programmatically. Specifically, we could modify line 23 to make it an instance field and set the logger name to the server name, something like:

```java
private final Logger log = LoggerFactory.getLogger(firstNonNull(serverName, "WireMock"));
```

src/main/java/com/github/tomakehurst/wiremock/common/Slf4jNotifier.java
```diff
@@ -15,19 +15,25 @@
  */
 package com.github.tomakehurst.wiremock.common;
 
+import static org.apache.commons.lang3.ObjectUtils.firstNonNull;
```
pks-1981, line 18: Please use ParameterUtils#getFirstNonNull

kapishmalik: Incorporated

pks-1981 (approved):

src/main/java/com/github/tomakehurst/wiremock/common/Slf4jNotifier.java
```diff
@@ -15,19 +15,25 @@
...
+  private Logger log = LoggerFactory.getLogger("WireMock");
 
   private final boolean verbose;
 
   public Slf4jNotifier(boolean verbose) {
     this.verbose = verbose;
   }
 
+  public void updateLogger(final String serverName) {
+    log = LoggerFactory.getLogger(getFirstNonNull(serverName, "WireMock"));
```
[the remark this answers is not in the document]
oleg-nenashev: TBH I am not site about making the logger non-static for that purpose. Maybe there is API would could invoke against a static field? Concurrency is not a problem here

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
```diff
@@ -171,6 +173,13 @@ public WireMockConfiguration timeout(int timeout) {
     return this;
   }
 
+  public WireMockConfiguration serverName(String serverName) {
+
+    this.serverName = serverName;
+    ((Slf4jNotifier) notifier).updateLogger(serverName);
```
oleg-nenashev, line 179: This unchecked cast impacts extensibility IMO. I suggest doing proper `instanceof` verification

oleg-nenashev (changes requested):
Looks good overall, requesting changes because of unverified lass cast, and I second @tomakehurst's note about the APIs though do not feel strongly about that

src/main/java/com/github/tomakehurst/wiremock/core/WireMockConfiguration.java
kapishmalik: done

oleg-nenashev (changes requested):
`Options#getServerName()` is still a breaking change, because there is no default implementation
https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html

kapishmalik:
> `Options#getServerName()` is still a breaking change, because there is no default implementation https://docs.oracle.com/javase/tutorial/java/IandI/defaultmethods.html

incorporated

oleg-nenashev (approved):
My comments are addressed

src/test/java/com/github/tomakehurst/wiremock/ServerNameInLogsAcceptanceTest.java
```diff
@@ -0,0 +1,44 @@
...
+  @BeforeAll
+  public static void setupServer() {
+    setupServer(options().serverName(SERVER_NAME));
+  }
+
+  @Test
+  public void checkServerNameIsPrintedInLogs() {
+
+    ByteArrayOutputStream byteArrayOutputStream = new ByteArrayOutputStream();
+    System.setOut(new PrintStream(byteArrayOutputStream));
```
tomakehurst, line 39: We need to un-set this in an @AfterEach block, otherwise it might suppress the output from other tests when running everything.

src/main/java/com/github/tomakehurst/wiremock/common/Slf4jNotifier.java
```diff
@@ -15,19 +15,34 @@
...
   public Slf4jNotifier(boolean verbose) {
+    this(verbose, null);
+  }
+
+  public Slf4jNotifier(boolean verbose, String loggerName) {
+    log = LoggerFactory.getLogger(getFirstNonNull(loggerName, "WireMock"));
     this.verbose = verbose;
   }
 
+  public void setLogger(String loggerName) {
```
tomakehurst, line 37: I still think we should try to get rid of this and make the logger final.

Issue #2420, opened 2023-10-06
FileSourceBlobStore throws FileNotFoundExceptions when it should return Optional.empty()


tomakehurst:
### Proposal

The `BlobStore` interface returns optionals when getting values, indicating that `Optional.empty()` should be returned when the item doesn't exist.

The `FileSourceBlobStore` violates this contract at the moment by throwing an exception. 

### Reproduction steps

Create a new `FileSourceBlobStore` pointing to an empty directory and call `get("any-key")`.

PR #2451, opened 2023-10-17
Fix contract for FileSourceBlobStore 

dkhozyainov:
Greetings, please look at my PR for #2420

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/store/files/FileSourceBlobStore.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/store/files/FileSourceBlobStore.java
@@ -20,6 +20,7 @@ import static com.github.tomakehurst.wiremock.common.LocalNotifier.notifier;
 import com.github.tomakehurst.wiremock.common.*;
 import com.github.tomakehurst.wiremock.store.BlobStore;
 import java.io.File;
+import java.io.FileNotFoundException;
 import java.io.InputStream;
 import java.util.Optional;
 import java.util.stream.Stream;
@@ -43,7 +44,10 @@ public class FileSourceBlobStore implements BlobStore, PathBased {
     try {
       return Optional.of(fileSource.getBinaryFileNamed(key).getStream());
     } catch (Exception exception) {
-      notifier().info("Error when working with FileSource:\n" + Json.write(exception.getMessage()));
+      if (!(exception instanceof FileNotFoundException)) {
+        notifier()
+            .info("Error when working with FileSource:\n" + Json.write(exception.getMessage()));
+      }
       return Optional.empty();
     }
   }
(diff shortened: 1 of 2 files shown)


oleg-nenashev (approved):
Looks okay though I would think about proper diagnostics logs, because now all exceptions would be silently suppressed. Would be also nice to add a human readable PR title for the changelog

dkhozyainov:
> Looks okay though I would think about proper diagnostics logs, because now all exceptions would be silently suppressed. Would be also nice to add a human readable PR title for the changelog

Can you tell me how nice it is to make diagnostic logs, maybe there is an example somewhere?

oleg-nenashev:
WireMock uses `com.github.tomakehurst.wiremock.common.LocalNotifier.notifier` for logging purposes, e.g.

```
notifier().info("Started recording with record spec:\n" + Json.write(spec));
```

I normally use stock logging frameworks, so not sure how much of debug logging we can make with it

dkhozyainov:
> @oleg-nenashev

@oleg-nenashev thanks for the hint, I added it in the next commit+ fixed code style

oleg-nenashev (approved):

src/main/java/com/github/tomakehurst/wiremock/store/files/FileSourceBlobStore.java
```diff
@@ -56,7 +63,12 @@ public Stream<String> getAllKeys() {
 
   @Override
   public Optional<byte[]> get(String key) {
-    return Optional.of(fileSource.getBinaryFileNamed(key).readContents());
+    try {
+      return Optional.of(fileSource.getBinaryFileNamed(key).readContents());
+    } catch (Exception exception) {
```
tomakehurst, line 68: How about we catch `FileNotFoundException` and don't log anything, since this is a normal event, then log at error for any other type of exception (which in practice would probably indicate a misconfigured file system)?

dkhozyainov: Done. I used instanceof because FileNotFoundException is cast to Throwable in the BinaryFile class via the throwUnchecked method

src/main/java/com/github/tomakehurst/wiremock/store/files/FileSourceBlobStore.java
```diff
@@ -38,7 +41,15 @@ public FileSourceBlobStore(FileSource fileSource) {
...
   public Optional<InputStream> getStream(String key) {
-    return Optional.of(fileSource.getBinaryFileNamed(key).getStream());
+    try {
+      return Optional.of(fileSource.getBinaryFileNamed(key).getStream());
+    } catch (Exception exception) {
+      if (!(exception instanceof FileNotFoundException)) {
+        notifier()
+            .info("Error when working with FileSource:\n" + Json.write(exception.getMessage()));
+      }
+      return Optional.empty();
```
tomakehurst, line 51: I think we should still throw the exception if it's not `FileNotFoundException`. Only return empty when it's exactly that exception.

We can probably do this with two catch blocks rather than `instanceof`.

The notifier call should be at `error` if it's not a not-found error.

tomakehurst:
Thanks!

PR #2453, opened 2023-10-17
Add ability to return application version from new admin endpoint and wiremock standalone cli

leeturner:
This PR provides the ability for the application to read the current version and return this version from a new admin endpoint and the cli for wiremock standalone.  

The current version is copied to a file in the `resources` folder using the same `filesWithVersion` mechanism as other files referenced in the build script.  The version is also added to the jar manifest files via the build script.

The PR adds a new `GET` admin endpoint at the following url - `__admin/version`.  By default this returns a simple json payload with the version information.  If the caller sends an `Accept` header with `text/plain` the version information is returned as a text response.

If used on the cli via `java -jar wiremock-standalone-3.2.0.jar --version` the version information will be output and the application will then exit.

One area that I am not sure about is how the open api files are updated so at the moment the new `version` admin endpoint is not referenced in there.

## Submitter checklist

- [] For new features, there's necessary documentation in this pull request or in a subsequent PR to [wiremock.org](https://github.com/wiremock/wiremock.org)


src/main/java/com/github/tomakehurst/wiremock/core/Version.java
```diff
@@ -0,0 +1,31 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.core;
+
+import java.io.IOException;
+import java.util.Properties;
+
+public class Version {
+  public static String getCurrentVersion() {
```
oleg-nenashev, line 22: How about caching it in a static variable? Java resource loading is heavy, and such unprotected endpoint may be exploited for various kinds of DDoS attempts

oleg-nenashev (approved):
LGTM, just one minor suggestion

src/main/java/com/github/tomakehurst/wiremock/core/Version.java
leeturner: Sounds great.  Thank you for the suggestion.  I have added simple caching of the version as suggested. If we add any other version information to the Version class we can add more sophisticated caching but for now this should do

oleg-nenashev (approved):
Needs Tom, fine with me

src/main/java/com/github/tomakehurst/wiremock/core/Version.java
tomakehurst: I've just merged a change with a `Lazy` utility class. This might be a nice way to do this (and is thread safe FWIW).

leeturner: Nice, I'll take a look at that and update the PR

leeturner: I have updated to the `Lazy` class.  Nice little utility class 👍

tomakehurst (approved):

Issue #2418, opened 2023-10-04
Async verification


tomakehurst:
### Proposal

WireMock supports a variety of `verify(...)` methods for checking that a request matching specific criteria were made to the server.

But these methods will only check whether a request matching the criteria has already been received. This is fine when testing synchronous code but when WireMock is called asynchronously it's useful to be able to wait for the request to arrive and fail after a timeout.

Although this can be achieved with external libraries such as Awaitility, incorporating it into WireMock's API would be a much slicker experience and would also greatly improve the experience for non-Java clients.

cjhetzle:
I see the method [WireMock::verifyThat](https://github.com/cjhetzle/wiremock/blob/43cfa3cfb23ca9d946fc0c35505ab06fd832714e/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java#L716) and it seems that we just need to add retries and an eventual timeout.

What would be an appropriate timeout?

cjhetzle:
Submitted a PR would love someone to help point me in how FixedDelay and RetryCount could be set by configuration settings.

PR #2456, opened 2023-10-19
Adding retry ability to WireMock Verify

cjhetzle:
Currently not using configuration settings
to set retry counts or fixed delay.
Add

## References

#2418 

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -718,21 +718,36 @@ public class WireMock {
     final RequestPattern requestPattern = requestPatternBuilder.build();
 
     int actualCount;
-    if (requestPattern.hasInlineCustomMatcher()) {
-      List<LoggedRequest> requests =
-          admin.findRequestsMatching(RequestPattern.everything()).getRequests();
-      actualCount = (int) requests.stream().filter(thatMatch(requestPattern)).count();
-    } else {
-      VerificationResult result = admin.countRequestsMatching(requestPattern);
-      result.assertRequestJournalEnabled();
-      actualCount = result.getCount();
-    }
+    int retryCount = 3;
+    long fixedDelay = 500;
 
-    if (!expectedCount.match(actualCount)) {
-      throw actualCount == 0
-          ? verificationExceptionForNearMisses(requestPatternBuilder, requestPattern)
-          : new VerificationException(requestPattern, expectedCount, actualCount);
-    }
+    do {
+      if (requestPattern.hasInlineCustomMatcher()) {
+        List<LoggedRequest> requests =
+            admin.findRequestsMatching(RequestPattern.everything()).getRequests();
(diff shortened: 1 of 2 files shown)


tomakehurst:
I think there's a better way of doing this rather than with retries (which could be expensive depending on the volume and match complexity), which is to create a `CompletableFuture` for each request to wait for a specific request.

Then you can call get on the future with a timeout and it'll block until the request arrive or the timeout is reached.

cjhetzle:
> I think there's a better way of doing this rather than with retries (which could be expensive depending on the volume and match complexity), which is to create a `CompletableFuture` for each request to wait for a specific request.
> 
> Then you can call get on the future with a timeout and it'll block until the request arrive or the timeout is reached.

Sure, I can look into doing that. I'm trying to get into concurrencies.

cjhetzle:
@tomakehurst alright, so I looked into 'CompletableFutures' but can't see how you mean for it to wait for a satisfying result. Once a FutureTask is complete I don't see any method to allow setting a desired result that automatically restarts the Task if failed. In which case, yeah, I don't see a way to do what you are describing without a Sleep call.

Could you further explain what you mean if you are sure it can be done?

tomakehurst:
@cjhetzle the `CompletableFuture.get(timeout)` method (defined by the `Future` interface) blocks until either the future has been marked complete or until the timeout expires. I think this should do what you need.

cjhetzle:
Let me take a look at this in the morning, should be straight forward to change if it works how I think.

cjhetzle:
@tomakehurst alright, so I am again looking at this issue and I need to explain why I am a bit confused by what you may think the `verifyThat` function is doing or doing wrong.

`verifyThat` will ask WireMock to return a full list of requests that it has received -> then it applies a search for a specific type and verifies the number that show in the list.
-> In this, there is *no point at which we are waiting* for results, which is where I can see the use of a `CompletableFuture timeout`.
    -> The issue I am trying to solve is WireMock giving us the wrong answer. So, it does return results and quickly. I can't see a way to solve this where we don't add a retry at some level.
    
The way you were describing CompletableFuture, I thought there was a method that you could tell it what result you want from the Task and it would maybe auto retry it for you until it returned what you were looking for or timedout.

tomakehurst:
I think we're on the wrong track with this PR.

What I had in mind is that we have a new set of verification methods like:

```java
waitForUpTo(20, SECONDS).until(getRequestedFor(urlPathEqualTo("/things")));
```

Calling this would create a `CompletableFuture` on the server, associate it with the match criteria then call `get(20, SECONDS)` on it. Then whenever a new request arrived it would be checked against the match criteria and if matching the future would be completed, releasing the call to `get(...)`.

Can I suggest you jump on the community Slack so we can discuss a bit more interactively?

tomakehurst:
I'm going to close this as I think we've ended up in quite different places in terms of how this should work. Happy to reopen and spend some more time on it if as and when you have the bandwidth.

Issue #2466, opened 2023-10-27
"truncateActual" results in null pointer exceptions


papiomytoglou:
### Proposal

Using a JSON mapping file which includes `truncateActual` in standalone results in null pointer exceptions for all requests with non-matching URIs or date/time values.

For example, when including the matcher below in a mapping
```
"queryParameters": {
  "datetime": {
    "equalToDateTime": "now",
    "truncateExpected": "first hour of day",
    "truncateActual": "first hour of day"
  }
}
```
all requests with matching URIs or `datetime` values will still be handled correctly but all other requests will result in the following null pointer exception:
```
java.lang.NullPointerException: Cannot invoke "java.time.ZonedDateTime.truncatedTo(java.time.temporal.TemporalUnit)" because "input" is null
```
Removing `truncateActual` stops the exception from being thrown but is not a solution since the two matchers are not equivalent.

Tested in standalone versions `3.0.1`, `3.1.0` and `3.2.0`

### Reproduction steps

Create a JSON file in `mappings` with the following data:
```
{
  "request": {
    "method": "GET",
    "urlPath": "/first",
    "queryParameters": {
      "datetime": {
        "equalToDateTime": "now",
        "truncateExpected": "first hour of day",
        "truncateActual": "first hour of day"
      }
    }
  },
  "response": {
    "status": 200,
    "body": "First stub"
  }
}
```
Create a second JSON file, also in `mappings`, with the following data:
```
{
  "request": {
    "method": "GET",
    "urlPath": "/second"
  },
  "response": {
    "status": 200,
    "body": "Second stub"
  }
}
```
Start `wiremock-standalone-3.2.0.jar` locally on port 80 only using the above two mappings then open a browser and navigate to the following URIs replacing `2023-10-27` with the current date:
```
http://localhost/first?datetime=2023-10-27T17:09:10.739Z
# Matches successfully

http://localhost/second
# Matches successfully

http://localhost/first
# Matching URI, absent date/time parameter, returns null pointer exception

http://localhost/third
http://localhost/nonexistent
http://localhost/12345
http://localhost/employee?id=42
# Non-matching URIs, all return null pointer exceptions

http://localhost/first?datetime=2047-01-01T17:09:10.739Z
# Matching URI, non-matching date/time parameter value, returns "Request was not matched" response

http://localhost/nonexistent?datetime=2023-10-27T17:09:10.739Z
# Non-matching URI, matching date/time parameter value, returns "Request was not matched" response
```

[...]

PR #2467, opened 2023-10-28
Add null check for actual date/time truncation (#2466)

papiomytoglou:
A missing null check in `DateTimeTruncation#truncate` results in null pointer exceptions whenever the user uses a matcher which includes `truncateActual`. This largely makes `truncateActual` unusable since it breaks functionality, especially when the mapping including `truncateActual` is added last.

## References
Fixes https://github.com/wiremock/wiremock/issues/2466

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/DateTimeTruncation.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/DateTimeTruncation.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Thomas Akehurst
+ * Copyright (C) 2021-2023 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


oleg-nenashev (approved):
Thanks! There is still an open issue with documentation/annotations and downstream use, but it definitely prevents NPE in this method. So it is net positive and should be merged IMO

papiomytoglou:
Thank you for your response @oleg-nenashev. I'd gladly add any required documentation, it's just not clear to me what it should be since neither the signature nor the return value of the method were affected. If you could point me to any relevant guidelines I will amend the PR.

oleg-nenashev:
@papiomytoglou Yes, this is actually the main problem. At the moment we do not have an agreed standard and guidelines on annotations, so it is not something one could address at the moment. Let's get it merged

PR #2468, opened 2023-10-29
chore: use List.of where possible if only one argument is passed

SimonVerhoeven:
Cleaned up some Arrays.asList that were never adapted/only one arguments, and removed them in case they weren't used at all.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/UniqueFilenameGenerator.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/UniqueFilenameGenerator.java
@@ -26,7 +26,7 @@ public class UniqueFilenameGenerator {
 
   public static String generate(String url, String prefix, String id, String extension) {
     String pathPart = Urls.urlToPathParts(URI.create(url));
-    pathPart = pathPart.isEmpty() ? "(root)" : sanitise(pathPart);
+    pathPart = pathPart.isBlank() ? "(root)" : sanitise(pathPart);
 
     if (pathPart.length() > 150) {
       pathPart = StringUtils.truncate(pathPart, 150);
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/tomakehurst/wiremock/common/UniqueFilenameGenerator.java
```diff
@@ -26,20 +26,13 @@ public static String generate(String url, String prefix, String id) {
 
   public static String generate(String url, String prefix, String id, String extension) {
     String pathPart = Urls.urlToPathParts(URI.create(url));
-    pathPart = pathPart.equals("") ? "(root)" : sanitise(pathPart);
+    pathPart = pathPart.isEmpty() ? "(root)" : sanitise(pathPart);
```
oleg-nenashev, line 29: Ideally should be `isBlank()` so that we handle more edge cases, but out of the scope for the PR title anyway

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JsonData.java
```diff
@@ -158,7 +158,7 @@ public boolean remove(Object o) {
     }
 
     public boolean containsAll(Collection<?> c) {
-      return data.containsAll(c);
+      return new HashSet<>(data).containsAll(c);
```
oleg-nenashev, line 161: Why?

oleg-nenashev (commented):
I am fine with a particular suggested refactoring, but many changes here are unrelated to it

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JsonData.java
SimonVerhoeven: HashSet.containsAll tends to be faster.

oleg-nenashev: Of course, but you would need to iterate through all the elements at least once to form the new collection. It defeats the purpose of this change for most of the cases

PR #2471, opened 2023-10-31
Delegate parsing of media types to library

jamietanna:
Instead of us handling any media-type related parsing, we should instead
delegate to a library for this.

---

This is a reimplementation of https://github.com/wiremock/wiremock/pull/1686 in part to freshly recreate it to avoid the merge conflicts, but also shamelessly to see if it can count for Hacktoberfest :jack_o_lantern:

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/ContentTypeHeader.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/ContentTypeHeader.java
@@ -17,21 +17,25 @@ package com.github.tomakehurst.wiremock.http;
 
 import com.github.tomakehurst.wiremock.common.Strings;
 import java.nio.charset.Charset;
+import java.util.Map;
 import java.util.Optional;
+import me.jvt.http.mediatype.MediaType;
 
 public class ContentTypeHeader extends HttpHeader {
 
   public static final String KEY = "Content-Type";
+  private static final String CHARSET_QUOTES = "\"([^\"]*)\"";
 
-  private String[] parts;
+  private final Optional<MediaType> maybeMediaType;
 
   public ContentTypeHeader(String stringValue) {
     super(KEY, stringValue);
-    parts = stringValue != null ? stringValue.split(";") : new String[0];
+    maybeMediaType = Optional.of(MediaType.valueOf(stringValue));
   }
 
   private ContentTypeHeader() {
     super(KEY);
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/tomakehurst/wiremock/http/ContentTypeHeader.java
```diff
@@ -17,21 +17,25 @@
...
 import java.util.Optional;
+import me.jvt.http.mediatype.MediaType;
 
 public class ContentTypeHeader extends HttpHeader {
 
   public static final String KEY = "Content-Type";
+  private static final String CHARSET_QUOTES = "\"([^\"]*)\"";
 
-  private String[] parts;
+  private final Optional<MediaType> maybeMediaType;
```
jamietanna, line 29: This has an IDE warning, but I believe it's the right thing to do to give us the ability to more easily handle the absense of the value. Thoughts? (I've not done Java in ~18 months so my knowledge is rusty!)

oleg-nenashev (dismissed):
The new library needs to be shaded

jamietanna:
Thanks Oleg, I _think_ I've done the shading right, but let me know!

tomakehurst:
Sorry @jamietanna in its current state I feel like the library isn't offloading enough of the work to be worth the addition of an extra dependency.

I agree in general that it would be good to outsource generic HTTP wrangling to libraries. Ideally I'd like to have a whole HTTP model (including a better URL implementation than Java's) extracted as a library, but that's a different sized job entirely...

jamietanna:
Hey @tomakehurst that's fair. Happy to leave this open, or close and we can look at adding in the future.

I know that one thing I was considering as a useful extension would be to allow Wiremock to perform content-negotiation - I've built some APIs in the past that will have an `Accept` header that is a bit awkward to statically mock up-front, but if we had i.e. `negotiatesTo: "application/vnd.api+json"` then it'd be possible to handle these requests more easily.

That being said, I'm not sure if it'd be that well-used - I don't see an issue raised for it, so happy to raise one and see if it's something the community would want?

tomakehurst:
I've had similar thoughts about content negotiation in the past but as you've said, there doesn't seem to be much demand for it.

I'll close this for now (in a futile effort at keeping things tidy), but happy to re-open it if we want to expand this use case a bit in the future.

Issue #1339, opened 2020-06-22
Enhancement: Require healthcheck that includes version, uptime and response time


rallygeetha:
It would be a nice feature to have a healthcheck endpoint that includes version, uptime and response time. Filing this as per request from Tom Akehurst's reply to my request.

## Bug reports

We're a lot more likely to look at and fix bugs that are clearly described and reproduceable. Please including the following details when reporting a bug:

- [x] Which version of WireMock you're using 

2.26.1

ChunjuanLi:
Hi @tomakehurst ,  I am working on this issue, and I would like to get your suggestion on how to get the version of wiremock. In gradle.build, version is presented, however I can't read the file and get the version value in java code as the config gradle.build won't be compiled. Based on that, I am thinking if we can hard code the version in heathcheck code such as result.setVersion("X.XX.X") or create a java config file to contain the version such that healthcheck can read the data from this java file, the downside of these solutions is that both of the gradle.build and java config/hard coded method are required to be updated together whenever there's a new version of wiremock. Can you please let me know your thoughts on it?

tomakehurst:
A typical solution to this is to write a version info file into the resources folder of the build output so that it's included in the JAR. Then in code you can read this file via the classpath and return the values from it in the response.

Here's an example from one of my projects:

```gradle
task setVersions(type: Copy) {

    def branch = 'git rev-parse --abbrev-ref HEAD'.execute().text.trim()
    def commitHash = 'git rev-parse --verify HEAD'.execute().text.trim()

    from 'src/main/resources/config/versions-template.json'
    into "$buildDir/resources/main/config/"
    rename { fileName ->
        fileName.replace('-template', '')
    }
    expand(
        GIT_BRANCH: branch,
        GIT_HASH: commitHash,
        VERSION: project.version
    )

}
```

And `versions-template.json`:

```json
{
  "version": "${VERSION}",
  "git":{
    "branch" : "${GIT_BRANCH}",
    "hash" :"${GIT_HASH}"
  }
}
```

ChunjuanLi:
@tomakehurst thanks! I will try that and make the code changes.

leeturner:
Hi, not sure if you have seen but we recently merged code from a separate PR to expose the version information. This doesn't include the git branch or hash but could be extended to do so - https://github.com/wiremock/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/core/Version.java

leeturner:
[...]

PR #2498, opened 2023-11-22
Version number add to the endpoint /health - healthcheck Issue Nro 1339

gorostiaga:
The version number was added  to the endpoint /health 
This partly covers the request on #1339 

## References
#1339 

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/admin/tasks/HealthCheckTaskTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/admin/tasks/HealthCheckTaskTest.java
@@ -26,6 +26,7 @@ import com.github.tomakehurst.wiremock.http.Request;
 import com.github.tomakehurst.wiremock.http.ResponseDefinition;
 import com.github.tomakehurst.wiremock.stubbing.ServeEvent;
 import java.net.HttpURLConnection;
+import org.hamcrest.Matchers;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/github/tomakehurst/wiremock/admin/tasks/HealthCheckTaskTest.java
```diff
@@ -46,5 +47,6 @@ public void healthy() {
         response.getStatusMessage(),
         equalTo(response.getReponseBody().asJson().get("message").asText()));
     assertThat(response.getReponseBody().asJson().get("status").asText(), is("healthy"));
+    assertThat(response.getReponseBody().asJson().get("version").asText(), Matchers.notNullValue());
```
leeturner, line 50: This all looks great. We keep a [version.properties](https://github.com/wiremock/wiremock/blob/master/src/test/resources/version.properties) file in the test classpath so we can actually validate the specific verion returned here - `is("X.X.X")`

gorostiaga: @leeturner the change has been made in sha: 88fb3b6dbdd700d4cb807f58c0eb825b8547b0d8

oleg-nenashev (approved):
+1, thanks! A good defensive code would be to ensure the health check does not fail if version cannot be retrieved by resources, e.g. by catching runtime exceptions and putting something like "unknown"

leeturner:
> +1, thanks! A good defensive code would be to ensure the health check does not fail if version cannot be retrieved by resources, e.g. by catching runtime exceptions and putting something like "unknown"

@oleg-nenashev That already happens in the `Version` class [here](https://github.com/wiremock/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/core/Version.java#L37)

PR #2504, opened 2023-11-24
Add support for custom logging by providing a Logback file as a CLI parameter for the standalone jar

LukeKynaston:
This change only impacts the standalone jar, and allows users to specify a `logback-config-file` CLI parameter to support custom logging.

i.e. `java -jar /path/wiremock-jre8-standalone-2.35.1-all.jar --logback-config-file=/path2/logback.xml`

Without this parameter, Logback is still used via the `Slf4jNotifier`, but the logging format is the same as it was before this change.

## References

- https://github.com/wiremock/wiremock/pull/1109
- Specifically this comment from @tomakehurst, was the desired outcome for this change: https://github.com/wiremock/wiremock/pull/1109#issuecomment-500991876 .

## Submitter checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -71,7 +71,7 @@ dependencies {
   api "org.ow2.asm:asm:9.4"
 
   implementation "org.slf4j:slf4j-api:1.7.36"
-  standaloneOnly "org.slf4j:slf4j-nop:1.7.36"
+  standaloneOnly "ch.qos.logback:logback-classic:1.2.12"
 
   api "net.sf.jopt-simple:jopt-simple:5.0.4"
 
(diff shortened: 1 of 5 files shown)


tomakehurst:
This isn't the design I want to go for here i.e. one that exposes a concrete dependency on a specific logging library.

Now that we've added sub-events I want to make this core to the logging strategy - significant info gets added a sub-events, then we log each event + children on each request and provide some startup options for whether the log is JSON, plain, HTTP line etc. and the level of detail.

More than happy to discuss if you'd like to work on this?

leeturner:
We now have a clearer idea of what we are moving towards with regards to logging.  There are some notes here - https://github.com/wiremock/wiremock/issues/1471#issuecomment-1907783172

As Tom mentioned, we are more than happy to support if you would like to work on this 👍

Issue #2447, opened 2023-10-15
Wiremock server sometimes responds with 500 when log limit is set


mykhaliev1:
### Proposal

Under high load wiremock server may respond with 500, when logs limit is set.
Reproducible with wiremock versions 3.0.2, 3.2.0
It happens ~0.01% of requests
<img width="1025" alt="image" src="https://github.com/wiremock/wiremock/assets/79195818/c3460f6f-5365-4550-9456-dc6b22b60c20">

### Reproduction steps

Steps to reproduce:
1) run standalone server from code: 
```
package org.example;
import static com.github.tomakehurst.wiremock.client.WireMock.*;

import com.github.tomakehurst.wiremock.WireMockServer;
import com.github.tomakehurst.wiremock.core.WireMockConfiguration;


public class Main {

    public static void main(String[] args) {
        WireMockConfiguration wireMockConfiguration = WireMockConfiguration.options()
                                                                            .maxRequestJournalEntries(10)
                                                                           .port(8080);
        
        WireMockServer wireMockServer = new WireMockServer(wireMockConfiguration);

        wireMockServer.addStubMapping(get(urlEqualTo("/echo"))
            .willReturn(aResponse()
                .withStatus(200)
                .withBody("echo")).build());
        
        wireMockServer.start();
    }
}
```

2) Create and run a jmeter test, that calling /echo endpoint. I reproduced it with test, that is using 20 threads and 10000 iterations

Wiremock log:
```
[main] INFO org.eclipse.jetty.server.Server - jetty-11.0.16; built: 2023-08-25T19:43:30.438Z; git: bedff458c4dd1a716d59e17b8cb0d2042eeab291; jvm 17.0.7+8-LTS-224
[main] INFO org.eclipse.jetty.server.handler.ContextHandler - Started o.e.j.s.ServletContextHandler@346939bf{/__admin,null,AVAILABLE}
[main] INFO org.eclipse.jetty.server.handler.ContextHandler.ROOT - RequestHandlerClass from context returned com.github.tomakehurst.wiremock.http.StubRequestHandler. Normalized mapped under returned 'null'
[main] INFO org.eclipse.jetty.server.handler.ContextHandler - Started o.e.j.s.ServletContextHandler@741b3bc3{/,null,AVAILABLE}
[...]

PR #2508, opened 2023-11-29
Do not return null objects in getAll of the InMemoryRequestJournalStore

coder-hugo:
Under heavy load it's possible that `serveEvents.get()` is called with an `UUID` that doesn't exist anymore in the `Map`. The reason for this is that the spliatator (required for `stream()`) of the `ConcurrentLinkedDeque` is weakly consistent. This means that it's not guaranteed that an element that is currently processed in a `Stream` still exists in the collection. So under heavy load it may happen, that one thread reads an element of the `ConcurrentLinkedDeque`. Then this thread gets paused and another thread gets activated which than completely runs the `remove` method of the `InMemoryRequestJournalStore`. Now the first thread gets activated again and calls the `get` method of the `serveEvents` with an `UUID` that doesn't exist anymore in the `Map` which results in a `null` value. As the remaining code doesn't expect `null` values in the `Stream` they should be filtered out. Another solution would be to `synchronize` the code of the `InMemoryRequestJournalStore` but this would slow down the whole thing dramatically.

I haven't added any tests for this as it's hard to reproduce a concurrency issue.

## References

- #2447 

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/store/InMemoryRequestJournalStore.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/store/InMemoryRequestJournalStore.java
@@ -36,7 +36,7 @@ public class InMemoryRequestJournalStore implements RequestJournalStore {
 
   @Override
   public Stream<ServeEvent> getAll() {
-    return deque.stream().map(serveEvents::get);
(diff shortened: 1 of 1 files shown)


tomakehurst:
Despite the lack of tests (which I agree are hard to reliably do) I think this looks safe to merge.

PR #2536, opened 2023-12-15
Adding build function for all kind of UrlPattern in RequestPatternBuilder

AlEmerich:
In order to be able to customize `url` with every kind of `UrlPattern` from a `RequestPattern` using the `like` function, it would be useful if `RequestPatternBuilder` allows to do so.

I added test only for `urlPath` because I did not judge that it was necessary for other (since it wil be exactly the same and lead to a lot of duplicated code, but feel free to tell me otherwise.

## References

- Related issue: https://github.com/wiremock/wiremock/issues/2531

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java
@@ -152,6 +152,26 @@ public class RequestPatternBuilder {
     return this;
   }
 
+  public RequestPatternBuilder withUrlPath(String url) {
+    this.url = WireMock.urlPathEqualTo(url);
+    return this;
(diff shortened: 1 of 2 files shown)


tomakehurst:
I think it might be more consistent to add a single `withUrl` overload that that takes a `UrlPattern` parameter, since this is how the main DSL works.

AlEmerich:
@tomakehurst Agree, should we still keep the `withUrl(String)` overload to ensure backward compatibility ?

AlEmerich:
@tomakehurst Hello :wave: I don't get why the build are failing, it seems to come from `spotless` but when I run `spotlessApply` locally, nothing changed, meaning that everything seems to be well formated.
Do you know what I need to do ?

PR #2550, opened 2023-12-30
Adding webhooks as extension externally logs a warning

### LINKED ISSUE #2528: UX: Prevent fatal failure when passing the Webhooks classpath to the CLI args
### Proposal

The webhooks extension was moved into the core and enabled by default in #2376, and after it some integration tests fail with `java.lang.NoSuchMethodException: org.wiremock.webhooks.Webhooks.<init>()`, for example in Docker integration tests until https://github.com/wiremock/wiremock-docker/pull/100 . It happens, because the initializer was removed.

To improve the user experiences for those who update top the recent 3.x versions, I suggest the following:

* Create an empty and `@Deprecated` `org.wiremock.webhooks.Webhooks.<init>()` entry point
* When the endpoint is called, log a warning that passing the class name is no longer required and may lead to compatibility issues in the future

### References

_No response_

prithvitewatia:
Closes #2528.
This PR adds a deprecated constructor for ```org.wiremock.webhooks.Webhooks``` class with logging that passing the class name is no longer required and may lead to compatibility issues in the future.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2023 Thomas Akehurst
+ * Copyright (C) 2011-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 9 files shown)


tomakehurst:
Thanks for creating this. Do we risk ending up with two instances of `Webhooks` being loaded by doing it this way?

prithvitewatia:
@tomakehurst Thanks for the review. Yes there is a risk of having two instance of ```Webhooks``` ( one formed with the constructor ```public Webhooks(
      WireMockServices wireMockServices,
      ScheduledExecutorService scheduler,
      List<WebhookTransformer> transformers)``` and one from the deprecated constructor). 
      
Do you think maybe we could throw an exception after logging from the deprecated constructor to solve this problem ?

tomakehurst:
I wonder if a better solution would be to look for that particular value in the extensions parameter, remove it before processing and log a warning.

prithvitewatia:
Hi @tomakehurst . Happy new year. I am sorry I didn't get you there. Are you suggesting that we should add a check while loading the extensions using [Extenstions' load](https://github.com/wiremock/wiremock/blob/b3874f041baad7c7466a11dfabc5125f802799b4/src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java#L224) method

tomakehurst:
I would probably make most sense to put the check somewhere it'll affect both CLI and programmatic initialisation of WireMock, so `Extensions.load()` or somewhere beneath.

PR #2551, opened 2023-12-31
Fix json-body not escaping special characters

G-Basak:
This pull request fixes json-body not escaping special characters.

When Response-Body object is constructed from JSON file, it is stored as byte-array (plain text). So, the escape sequences are never replaced with their correct representation. They remain as is.
This creates problem when the JSON has special meaning after escaping. Such as the issue reported in : https://github.com/wiremock/wiremock/issues/2547

## References

- `JsonNode` of Jackson offers a method `asText()` which return the actual value of `ValueNode`. This will return a escaped string for a `TextNode`
- [Javadoc for `JsonNode#asText()`](https://javadoc.io/doc/com.fasterxml.jackson.core/jackson-databind/2.10.0/com/fasterxml/jackson/databind/JsonNode.html#asText--)
- [Difference Between asText() and toString() in JsonNode](https://www.baeldung.com/java-jsonnode-astext-vs-tostring#special-characters)
- New member variable added to `Body` which stores the parsed JSON-nodes. This helps to avoid converting between escaped and non-escaped string many times.
- The constructors of `Body` are modified to store the escaped JSON string when necessary.
- Another Pull Requests(https://github.com/wiremock/wiremock/pull/2559) adds a few tests which is used in this PR to prove that the behavior is not changing for non-JSON body.

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/http/BodyTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/http/BodyTest.java
@@ -166,8 +166,7 @@ class BodyTest {
             + "    \"newline : \\n padding\",\n"
             + "    \"carriage return : \\r padding\",\n"
             + "    \"horizontal tab: \\t padding\",\n"
-            + "    \"hex digit: \\u12ab"
-            + " padding\"\n"
+            + "    \"hex digit: \\u12ab padding\"\n"
             + "  ]\n"
             + "}\n";
     byte[] jsonBytes = bytesFromString(jsonString);
--- a/src/test/java/com/github/tomakehurst/wiremock/http/BodyTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/http/BodyTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2015-2023 Thomas Akehurst
+ * Copyright (C) 2015-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


src/test/java/com/github/tomakehurst/wiremock/stubbing/JsonTest.java
```diff
@@ -110,6 +113,131 @@ public void counts1ForEmptyObject() {
...
+            .replace("\\n", "\n")
+            .replace("\\r", "\r")
+            .replace("\\t", "\t")
+            .replaceAll("\\\\u12ab", "\u12ab")
+            .getBytes(StandardCharsets.UTF_8);
+    assertThat(result, is(jsonCompressedAndEscaped));
+  }
+
+  @Test
+  public void test2() {
```
leeturner, line 236: Super small minor thing but I was wondering if we could add a more targetted name for this test?

G-Basak: This test is not really needed. I had it for my local testing. 😅

src/test/java/com/github/tomakehurst/wiremock/http/BodyTest.java
```diff
@@ -86,4 +211,12 @@ void hashCorrectly() {
 
     assertEquals(body.hashCode(), body2.hashCode());
   }
+
+  private String base64EncodeToString(String string) {
+    return new String(base64Encode(string), StandardCharsets.UTF_8);
+  }
+
+  private byte[] base64Encode(String string) {
+    return Base64.getEncoder().encode(string.getBytes(StandardCharsets.UTF_8));
```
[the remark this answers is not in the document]
G-Basak: Reused existing encoding functions.

G-Basak:
> I was also thinking it might be nice to add a few examples with this kind of escaping to the website here - https://github.com/wiremock/wiremock.org

@leeturner thank you for the suggestion.
Actually I am not really adding anything new. These escape sequence are part of standard JSON syntax.

But if you still think that documentation is needed, let me know where in the doc I should add.

leeturner:
Thank you for the updates.  We just need Tom to cast his eyes over it and then we can take it from there

leeturner (approved):
Looks great.  Really appreciate your contribution

aivus:
Just for the history. The changes were reverted to fix https://github.com/wiremock/wiremock/issues/2609

PR #2555, opened 2024-01-02
Added GET_OR_HEAD method.

### LINKED ISSUE #2414: GET_OR_HEAD method

prithvitewatia:
Closes #2414 . This PR adds a GET_OR_HEAD method. This stub method could be used for both get as well as head Http method. The response and http header information is returned accordingly

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2021 Thomas Akehurst
+ * Copyright (C) 2011-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,7 +33,7 @@ public class RequestMethod implements NamedValueMatcher<RequestMethod> {
   public static final RequestMethod HEAD = new RequestMethod("HEAD");
   public static final RequestMethod TRACE = new RequestMethod("TRACE");
   public static final RequestMethod ANY = new RequestMethod("ANY");
-
+  public static final RequestMethod GET_OR_HEAD = new RequestMethod("GET_OR_HEAD");
   private final String name;
 
   public RequestMethod(String name) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
```diff
@@ -75,7 +75,9 @@ public boolean equals(Object o) {
     if (o == null || getClass() != o.getClass()) return false;
 
     RequestMethod that = (RequestMethod) o;
-
+    if (name.equals("GET_OR_HEAD") && (that.name.equals("GET") || that.name.equals("HEAD"))) {
```
tomakehurst, line 78: I'm not sure doing the equals behaviour this way is a good idea. If we ever needed to put this e.g. as a map key I suspect this would prevent that from working properly.

src/test/java/com/github/tomakehurst/wiremock/client/WireMockClientAcceptanceTest.java
```diff
@@ -70,4 +73,60 @@ public void buildsMappingWithUrlOnyRequestAndResponseWithJsonBodyWithDiacriticSi
...
+
+    wireMock.register(
+        getOrHead(urlEqualTo(path))
+            .willReturn(
+                okJson("{\"key\": \"value\"}").withHeader("Content-Type", "application/json")));
+    WireMockResponse response = testClient.head(path);
+
+    assertThat(response.statusCode(), is(200));
+    assertThat(response.firstHeader("Content-Type"), is("application/json"));
+    assertThat(response.content(), is(emptyOrNullString()));
```
tomakehurst, line 103: Would be great to see an assertion about `Content-Length` matching the length of the response body.

Mahoney:
I was wondering if it would be worth just getting `HEAD` requests to match `GET` mappings by default, if no `HEAD` mapping matches? Obviously without returning the body...

tomakehurst:
> I was wondering if it would be worth just getting `HEAD` requests to match `GET` mappings by default, if no `HEAD` mapping matches? Obviously without returning the body...

I did wonder about this, but the `GET_OR_HEAD` approach felt a bit safer as it's something you're explicitly opting into rather than a subtle change to default behaviour (even if that change is in line with what a lot of web servers will do by default).

src/main/java/com/github/tomakehurst/wiremock/http/RequestMethod.java
prithvitewatia: @tomakehurst Thanks for the feedback. Requested changes have been made.

tomakehurst:
Thanks!

Issue #2557, opened 2024-01-05
Shading SLF4J breaks the notifier


toadzky:
### Proposal

*Expected:* The `Slf4jNotifier` should log requests, even in stand-alone mode.
*Actual:* The `Slf4jNotifier does *not* log requests in stand-alone mode.
*Cause:* The embedded/shaded slf4j implementation doesn't find any implementations on the classpath and therefore uses the `NOPLogger` as it's fallback implementation - which doesn't log anything anywhere.
![image](https://github.com/wiremock/wiremock/assets/2286682/59820cb6-743f-4a79-b485-60c9a1e78dac)

*Proposed Fix:* Stop shading SLF4J.

### Reproduction steps

Use the `Slf4jNotifier` while using the wiremock-standalone dependency (my current version is 3.3.1).

toadzky:
alternate solutions: 

- i don't think the slf4j api being used by this project changed between versions 1 and 2, so it could be made a provided dependency and it would pick up whatever version was already available on the classpath.
- have separate notifiers for slf4j1 and 2 and have both of them as provided dependencies on the classpath. i'm not sure how testing would work for this because slf4j didn't change packages with major versions so it's hard to say which implementation would be picked up for testing
- shade some slf4j implementations like logback so it would at least not use the noplogger implementation
- detect the noplogger and use JUL instead with the assumption that a) it's better than nothing and b) hopefully the actual project has the jul shim correctly installed so whatever slf4j they have wired up will pick up the logs correctly

i think the best options would be the first or last options: make it provided and let it use whatever it finds or detect noplogger and use JUL instead

toadzky:
https://github.com/wiremock/wiremock/pull/2558#issuecomment-1891795703

dlehammer:
Hi @toadzky ,

Just wanted to show my support for this particular suggestion 👍 

> i don't think the slf4j api being used by this project changed between versions 1 and 2, so it could be made a provided dependency and it would pick up whatever version was already available on the classpath.

---

I encountered a similar symptom, where no logs was propagated from WireMock as it unfortunately ended up utilizing: `NOPLoggerFactory`

```
SLF4J: Class path contains multiple SLF4J bindings.
[...]

PR #2558, opened 2024-01-05
Stop relocating slf4j

toadzky:
Stop relocating SLF4J in the standalone shadowjar

## References

- #2557 

## Submitter checklist

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -312,7 +312,6 @@ shadowJar {
   relocate "jakarta.servlet", "wiremock.jakarta.servlet"
   relocate "org.checkerframework", "wiremock.org.checkerframework"
   relocate "org.hamcrest", "wiremock.org.hamcrest"
-  relocate "org.slf4j", "wiremock.org.slf4j"
   relocate "joptsimple", "wiremock.joptsimple"
   exclude 'joptsimple/HelpFormatterMessages.properties'
   relocate "org.yaml", "wiremock.org.yaml"


tomakehurst:
Unfortunately shading it is the lesser of two evils. If we don't it causes issue when there's a different major version also on the classpath.

toadzky:
is there a different solution you would prefer? here are some options:

* i don't think the slf4j api being used by this project changed between versions 1 and 2, so it could be made a provided dependency and it would pick up whatever version was already available on the classpath.
* have separate notifiers for slf4j1 and 2 and have both of them as provided dependencies on the classpath. i'm not sure how testing would work for this because slf4j didn't change packages with major versions so it's hard to say which implementation would be picked up for testing
* shade some slf4j implementations like logback so it would at least not use the noplogger implementation
* detect the noplogger and use JUL instead with the assumption that a) it's better than nothing and b) hopefully the actual project has the jul shim correctly installed so whatever slf4j they have wired up will pick up the logs correctly

tomakehurst:
The direction we're taking is to avoid needing slf4j at all and instead attaching sub-events during processing as a way of reporting the kinds of things we'd usually log.

What log messages are you hoping to see?

toadzky:
i have wiremock log the matches and failed matches. i recently spent several days stepping though wiremock code trying to find the source of a 400 from my app in a test because i couldn't actually see what wiremock was returning since the slf4j notifier is broken in standalone mode. since regular wiremock is incompatible with spring, standalone mode is the only option and it completely breaks this important feature.

tomakehurst:
Would happily accept this as a contribution BTW, as I'm not sure anyone from the core team is going to be able to get to it immediately.

toadzky:
I'll see what I can do about implementing it.

tomakehurst:
Thanks @toadzky!

If you want to get started by creating the repository yourself we can transfer it into the wiremock org in due course.

PR #2561, opened 2024-01-08
Add extensionScanningEnabled option to @WireMockTest.

kyle-winkelman:
When using extensions that utilize service loading (e.g. [wiremock/wiremock-graphql-extension](https://github.com/wiremock/wiremock-graphql-extension)) it would be preferable to use the declarative option to enable extension scanning, rather than being forced to use the programmatic approach to manually define extensions.

## References

- https://github.com/wiremock/wiremock.org/pull/242

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/junit5/WireMockExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2023 Thomas Akehurst
+ * Copyright (C) 2021-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


kyle-winkelman:
Alternatively, I am curious why `extensionScanningEnabled` isn't true by default. Now that extensions are encouraged to [register themselves via service loading](https://wiremock.org/docs/extending-wiremock/#extension-registration-via-service-loading), it would make sense to default to picking them up.

kyle-winkelman:
I would be willing to attempt to get all the options available available in WireMockConfiguration also available in WireMockTest annotation if that would be preferred over only adding one option.

tomakehurst:
Extension scanning is disabled by default when setting up WireMock programmatically because you might want to run multiple instances with different combinations of extensions, or you might want to programmatically construct an extension for more control.

The `WireMockTest` annotation is meant as a convenience when you mostly want defaults. We recommend taking the `@RegisterExtension` approach when you want to customise significantly. However, adding the extension scanning flag to it makes sense I think.

tomakehurst:
Thanks!

Would you mind creating a documentation PR for this at https://github.com/wiremock/wiremock.org?

kyle-winkelman:
Here is the documentation PR:
https://github.com/wiremock/wiremock.org/pull/242

PR #2564, opened 2024-01-09
Supply lazy TemplateEngine, HttpClientFactory, and DefaultHttpClient.

kyle-winkelman:
Draft for issue in which ExtensionFactories are supplied a null TemplateEngine, HttpClientFactory, and DefaultHttpClient from Extensions. 

## References

- closes https://github.com/wiremock/wiremock/issues/2562

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/Extensions.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Thomas Akehurst
+ * Copyright (C) 2023-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


tomakehurst:
This sort of thing is a shortcoming of not having full DI (which I briefly tried by decided against the added complexity).

The approach we're intending to promote to solve this is the `Lazy` class e.g. https://github.com/wiremock/wiremock/blob/master/src/main/java/org/wiremock/webhooks/Webhooks.java#L44

kyle-winkelman:
What do you think of supplying lazy implementations as the PR currently reflects?

Do you know much about backwards compatibility? Because I contemplated turning TemplateEngine into an interface to make it easier to extend, but was unsure if that would maintain compatibility.

tomakehurst:
Hmmmm...I hadn't thought of doing this way but I like it, I think...

kyle-winkelman:
Haha I have a similar sentiment. I don't love this approach, but I can't think of a better way. Maybe explicitly creating classes named LazyTemplateEngine, LazyHttpClientFactory, and LazyHttpClient would help make it more clear?

Alternatively, we extensively document in [Extending WireMock via custom code](https://wiremock.org/docs/extending-wiremock/#factories) the limitations of using these methods.

kyle-winkelman:
Done!

PR #2574, opened 2024-01-15
Update wrapper scripts version to match gradle version.

### LINKED ISSUE #2572: gradle-wrapper.jar is inconsistent with current gradle version.
### Summary

# Problem Description

Currently wiremock is using gradle 8.0.1 as mentioned in gradle-wrapper.properties and the checksum of current wrapper.jar is 33ad4583fd7ee156f533778736fa1b4940bd83b433934d1cc4e9f608e99a6a89. However, as can be seen from official doc https://gradle.org/release-checksums/, the expected checksum of wrapper.jar 8.0.1 should be 91941f522fbfd4431cf57e445fc3d5200c85f957bda2de5251353cf11174f4b5. The inconsistency means that the wrapper.jar is stale and should be updated.

# Possible Solution 
As mentioned in gradle official document https://docs.gradle.org/current/userguide/gradle_wrapper.html#sec:upgrading_wrapper, you will need to run the wrapper task a second time to make sure the wrapper.jar is updated. You can run the below command in the root project. Automatically the wrapper scripts will be generated by gradle.
```
bash gradlew wrapper
```

### References

_No response_

HappyHacker123:
## References
Fix #2572. 

## Submitter checklist


tomakehurst:
Will need to check whether our release tasks still work after a major version bump, since this usually seems to be what breaks...

HappyHacker123:
> Will need to check whether our release tasks still work after a major version bump, since this usually seems to be what breaks...

Thanks for your reply :). It seems that wiremock is already using gradle 8.0.1 in current gradle.properties. 
```
distributionUrl=https\://services.gradle.org/distributions/gradle-8.0.1-bin.zip
```
But the gradle wrapper scripts haven't been updated, still in 7.2. So basically this pr only upgrades the scripts downloading gradle, not upgrades gradle version itself. So it's true that the risk exists, but i think it's relatively minor :).

PR #2592, opened 2024-01-25
Support custom content patterns and rework content pattern deserialization.

kyle-winkelman:
Allow users to define `ContentPatternExtension`s  to support custom content patterns. `Json` becomes an injectable service that is supplied where deserialization is required to handle these custom content patterns.

In order to allow these new `ContentPattern`s to support an arbitrary number and variety of parameters the `ContentPattern` uses `@JsonTypeInfo.Id.DEDUCTION` and registers the built-in classes. This means that the `@JsonCreator` of the subtypes will be called for instantiation.

This is superior to using [custom-matching](https://wiremock.org/docs/extensibility/custom-matching/):

- Multiple `bodyPatterns` are able to be supplied whereas only a single `customMatcher` can be supplied.
- The [diff](https://github.com/wiremock/wiremock/blob/master/src/main/java/com/github/tomakehurst/wiremock/verification/diff/Diff.java) for bodyPatterns are explicitly listed and easier to understand than the generic message for customMatcher.
- A `StringValuePattern` has more reusability with other components (e.g. `matchesJsonPath`, `matchesXPath`, `and`, `or`, `not`, etc.)

## References

-  https://github.com/wiremock/wiremock.org/pull/285

## Submitter checklist


src/main/java/com/github/tomakehurst/wiremock/common/Json.java
```diff
@@ -50,6 +51,7 @@ protected ObjectMapper initialValue() {
           objectMapper.configure(JsonParser.Feature.IGNORE_UNDEFINED, true);
           objectMapper.configure(JsonGenerator.Feature.WRITE_BIGDECIMAL_AS_PLAIN, true);
           objectMapper.configure(DeserializationFeature.USE_BIG_DECIMAL_FOR_FLOATS, true);
+          objectMapper.configure(MapperFeature.ACCEPT_CASE_INSENSITIVE_ENUMS, true);
```
kyle-winkelman, line 54: This is `@Deprecated` it recommends using `JsonMapper` instead of `ObjectMapper` when supplying `MapperFeature`s.

src/main/java/com/github/tomakehurst/wiremock/matching/ContentPatternDeserialiser.java
```diff
@@ -16,30 +16,79 @@
...
+  }
+
+  private ContentPatternDeserialiser(Class<?> vc) {
+    super(vc);
+  }
+
+  public static void registerPatterns(
+      Map<String, Class<? extends ContentPattern<?>>> contentPatterns) {
+    contentPatterns.forEach(PATTERNS::putIfAbsent);
+  }
```
kyle-winkelman: Looking for recommendations for `Extensions` to supply these in a different way. The only alternative I was able to come up with was to use a `ServiceLoader` here to go get them, but that would most likely require an additional `META-INF/services` file specifically for `ContentPatternExtension` and they wouldn't be able to be supplied from `ExtensionFactory`.

kyle-winkelman: My concern is that if multiple WireMock servers are created and only one registers the `ContentPatternExtension` all will still have access to use it.

kyle-winkelman (commented):
The changes represented in `AdminApiTest` are a change in behavior allowing any value that can be converted to String, not just `TextNode`s.

kyle-winkelman:
@tomakehurst, I see you self-assigned this. I was curious if you had any thoughts on this PR. My main motivation is for improving [wiremock-graphql-extension](https://github.com/wiremock/wiremock-graphql-extension) to expose a reusable EqualToGraphqlPattern.

tomakehurst:
It's a change I'd definitely like to make, but reviewing a 31 file PR requires a level of uninterrupted focus I'm not not getting much of at the moment.

Also the priority is sorting out this Jetty 12 compatibility issue as nearly everyone using Spring is complaining about it. After that's shipped I should be able to take a proper look at this.

src/main/java/com/github/tomakehurst/wiremock/matching/ContentPatternDeserialiser.java
```diff
@@ -16,30 +16,78 @@
...
+  public ContentPatternDeserialiser() {
+    super(ContentPattern.class);
+  }
+
+  private ContentPatternDeserialiser(Class<?> vc) {
+    super(vc);
+  }
+
+  public static void registerPatterns(Map<String, Class<? extends ContentPattern<?>>> patterns) {
+    patterns.forEach(PATTERNS::putIfAbsent);
```
tomakehurst, line 65: The issue with this approach is when there are multiple WM instances running in the same JVM. Most of the time it'll probably be fine, but produce weird edge cases when not.

So far we've kept global static state out of WireMock for precisely this reason and I'd be reluctant to start introducing it here.

From my analysis a while back, there are 2 other approaches that could be made to work per-instance:
1) Make `Json` a service that's injected everywhere it's needed, then custom serializers can be attached to the `ObjectMapper` it wraps. This would be the cleanest way to do it but is a pretty large amount of work.
2) Deserialize unknown matcher keys into a `Map` (e.g. via Jackson's `@JsonAnySetter` /  `@JsonAnyGetter`) then later associate them with matcher extensions that have been registered.

kyle-winkelman: I have just pushed some code for another option:
3. Use thread local state instead of global static state. Before performing any operations that deserializes a `StubMapping` or `StubMappingCollection`, use the `Extensions` of the currently running WM instance to supply `ContentPatternDeserialiser` with the extensions that have been registered.

kyle-winkelman: 1. Seems too complex of a change.
2. I may spend some time looking at this option, but I think it will look a lot like my current code. Where I make changes in the locations that deserialize `StubMapping` and `StubMappingCollection`, but instead of tampering with `ContentPatternDeserialiser`, I would post process the object to create concrete stub mappings from the generic map.

tomakehurst:
Hey, sorry for taking so long to review this.

I think this is an improvement over having to live with shared/static state, I still have a "there be dragons" feeling about using thread locals on this way, based on repeated previous experiences of edge cases, hard-to-debug problems and difficulty setting up tests correctly.

I do have an idea derived from yours that could get us to a better place however: modify the `Json` class so that it can be an instance injected as a service, but also set the threadlocal static instance to the one configured with the WireMock instance's settings. That way we can avoid having to do a big-bang cutover to the injected service approach, and migrate towards it over time, strangling the static use of `Json` over time.

kyle-winkelman:
I have one more simple idea I would love for you to review when you have the time. This approach will attempt to use the built-in patterns first, but will then fall back to looking for a `class` field and loading that custom pattern if no built-in is matched.

Pros:
- Easy to extend and understand
- Smaller code change
Cons:
- Doesn't follow existing extension methodology (if it is on the classpath it will work)

Example:
```json
{
    "class": "a.b.c.CustomStringValuePattern",
    "propertyOne": "a",
    "propertyTwo": "b"
}
```

tomakehurst:
I'm not really keen on this idea I'm afraid, as it means a) (as you say) there'd be two ways to add extensions, b) it wouldn't be possible to inject WireMock services or otherwise define your own construction logic.

kyle-winkelman:
Thanks @tomakehurst for your patience with this PR.

Currently running into issue around `HttpAdminClient` and `RemoteMappingsLoader` as there is currently no way for the `HttpAdminClient` to ask the remote server what `Extensions` it has.
- Should I add an Admin endpoint that allows to query for that?
- What should behavior be if local JVM doesn't have those `Extensions` on its classpath (for remote wiremock server)? For example, `HttpAdminClient#listAllStubMappings()` may need to parse some custom body matchers that it doesn't have.

tomakehurst:
I've refactored and mostly solved the client issue here, if you want to try pulling this:
https://github.com/wiremock/wiremock/tree/tom-custom-content-pattern

I still can't get the custom registration of `ContentPattern` subtypes to work at the JSON mapper factory level, but if we can figure that out then I think this starts to look like a decent solution.

src/main/java/com/github/tomakehurst/wiremock/extension/ClientExtensions.java
```diff
@@ -0,0 +1,53 @@
...
+
+import static java.util.stream.Collectors.toMap;
+
+import java.util.List;
+import java.util.function.Function;
+import java.util.stream.Stream;
+
+public class ClientExtensions extends Extensions {
+
+  public ClientExtensions(ExtensionDeclarations extensionDeclarations) {
```
kyle-winkelman, line 26: I believe `ClientExtensions` should accept a boolean to indicate `isExtensionScanningEnabled`, so that it can `loadExtensionsAsServices` similarly to `ServerExtensions`.

src/main/java/com/github/tomakehurst/wiremock/extension/ExtensionDeclarations.java
```diff
@@ -74,6 +74,10 @@ public List<ExtensionFactory> getFactories() {
     return factories;
   }
 
+  public boolean isEmpty() {
```
kyle-winkelman, line 77: This does not appear to be used, should it be removed?

src/main/java/com/github/tomakehurst/wiremock/extension/ExtensionFactory.java
```diff
@@ -15,8 +15,15 @@
...
 
+import static java.util.Collections.emptyList;
+
 import java.util.List;
 
 public interface ExtensionFactory {
+
   List<Extension> create(WireMockServices services);
+
+  default List<Extension> createForClient() {
```
kyle-winkelman, line 26: I am slightly concerned about this, as it makes it a little more difficult to properly build an `ExtensionFactory`. I am also curious if anyone would ever require `WireMockServices` to build a `ContentPatternExtension` (some really bad code that looks up the `getContentPatternClass` from an option or file).

kyle-winkelman (commented):
These are a couple things I would like your feedback on.

kyle-winkelman:
Thank you so much @tomakehurst, what you have is amazing. I think I fixed the issue you were running into around registration (it wasn't actually a registration issue, just that `StartsWithMatcher` didn't properly serialize to json).

kyle-winkelman:
For user debugging, I was wondering if in `HttpAdminClient` we should catch and wrap these two exceptions and provide additional information. For example, `There is a mismatch in the extensions registered on the remote WireMock server and the local client. Please ensure extensions are supplied to WireMockBuilder#extensions or that the extensions are loadable via Service Loading and it is enabled for the client (i.e., WireMockBuilder#extensionsViaServiceLoading).`

#### Remote WireMock server does not have an Extension that the client tried to provide:
```
com.github.tomakehurst.wiremock.common.InvalidInputException: {
  "errors" : [ {
    "code" : 10,
    "source" : {
      "pointer" : "/request/bodyPatterns/0"
    },
    "title" : "Error parsing JSON",
    "detail" : "Could not resolve subtype of [simple type, class com.github.tomakehurst.wiremock.matching.ContentPattern<java.lang.Object>]: Cannot deduce unique subtype of `com.github.tomakehurst.wiremock.matching.ContentPattern<java.lang.Object>` (20 candidates match)"
  } ]
}
	at com.github.tomakehurst.wiremock.common.ClientError.fromErrors(ClientError.java:38)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.throwParsedClientError(HttpAdminClient.java:596)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.safelyExecuteRequest(HttpAdminClient.java:561)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.executeRequest(HttpAdminClient.java:519)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.executeRequest(HttpAdminClient.java:495)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.addStubMapping(HttpAdminClient.java:157)
	at com.github.tomakehurst.wiremock.client.WireMock.register(WireMock.java:464)
	at com.github.tomakehurst.wiremock.client.WireMock.register(WireMock.java:459)
```

#### Client does not have an Extension that the remote WireMock server has:
```
com.github.tomakehurst.wiremock.common.JsonException: {
  "errors" : [ {
    "code" : 10,
    "source" : {
      "pointer" : "/mappings/0/request/headers/header"
    },
    "title" : "Error parsing JSON",
    "detail" : "Could not resolve subtype of [simple type, class com.github.tomakehurst.wiremock.matching.StringValuePattern]: Cannot deduce unique subtype of `com.github.tomakehurst.wiremock.matching.StringValuePattern` (20 candidates match)"
  } ]
}
	at com.github.tomakehurst.wiremock.common.JsonException.fromJackson(JsonException.java:52)
	at com.github.tomakehurst.wiremock.common.Json.readValue(Json.java:111)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.executeRequest(HttpAdminClient.java:521)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.executeRequest(HttpAdminClient.java:495)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.executeRequest(HttpAdminClient.java:485)
	at com.github.tomakehurst.wiremock.client.HttpAdminClient.listAllStubMappings(HttpAdminClient.java:187)
	at com.github.tomakehurst.wiremock.client.WireMock.allStubMappings(WireMock.java:484)
```

kyle-winkelman:
@tomakehurst, I have addressed all my own concerns and was hoping to get another review. Also, take a look at the corresponding PR for wiremock.org. Thanks for your time!

PR #2593, opened 2024-01-27
Add support of Jetty 12

reta:
Add support of Jetty 12

## References

- https://github.com/wiremock/wiremock/issues/2395

Road block (so far):
 - [x] ~`org.eclipse.jetty.server.MultiPartInputStreamParser` is gone for good~ replaced with MultiPartFormData.Parser but it does not have any special treatment to Base64 (as Jetty 11 had)
 - [x] Jetty 12 does not currently support cross context dispatch (workaround - use AdminXxxContext / MockXxxContext explicitly)

Jetty 12 limitations to document:
 -  [ ] when serving resource directory, always sends the redirect (if the path does not end with the `/`)
 -  [ ] does not allow to set reason / message along with set response status (HTTP 1.1 at least)
 -  [ ] does not decode Base64 (and other encodings) when using multipart form data

TODOs (from @tomakehurst):

Perhaps it might be worth considering a separate JAR after all. I've had a couple of ideas that might make it easier to live with:

- [x]  Still detect the Jetty version in the core JAR and if it's 12 exit with a message indicating that the user should add the jetty12 JAR to their dependencies.
- [x] Use Java service loading to delegate to the appropriate server factory so e.g. when there's no service loader impl available we default to the built-in Jetty 11 factory, but if one is present we use this. Then the Jetty 12 JAR would register its factory as a service implementation and this would get loaded automatically when the JAR was present on the classpath.

## Submitter checklist


reta:
@tomakehurst would love to hear your opinion before committing more time, the are 2 issues that I run into with implementing the support of both Jetty 11 and Jetty 12:
 - Jetty 12 needs JDK-17: easy to address by switching the build to JDK-17 but keeping JDK-11 baseline
 - Jetty 12 has backward incompatible changes with Jetty 11 for existing APIs (fe `HandlerCollection`, etc.): this is more difficult to deal with since we are having conflicting artifacts / classes but different Jetty versions

The way I was thinking to make it work is that (it should be more or less clean):
 - Jetty 11 stays as the default version (no change here)
 - build requires JDK-17 but keeps JDK-11 baseline (no user-visible changes here)
 - use multi-release JAR (Jetty 12 goes to `java17`), not all build tools deals with that well enough sadly 
 - use dynamic Jetty container detection at runtime

With multi-release JAR we should be able to replace Jetty 11 with Jetty 12 and compile the integration properly. The other option (without the multi-release jar) would be to have some kind of "stubs" at compile time to fill up the API discrepancies (haven't tried that yet but looks a bit ugly). 

Would appreciate the hints what direction you would steer towards, thank you.

tomakehurst:
I broadly agree with your proposed approach, but I think a multi-release JAR might not be such a great idea as it's the Jetty version rather than Java that we're switching on and we want to support the case of Java 17 + Jetty 11 (someone out there will definitely be doing this!).

What I suggest is that we try to find a way to put all of stuff where the Jetty API varies under an abstraction in `JettyHttpServer`, create `jetty11` and `jetty12` packages with implementations subclassing this and a default implementation that selects and delegates to the appropriate one after querying `Jetty.VERSION`.

reta:
Thanks a lot @tomakehurst 

> I broadly agree with your proposed approach, but I think a multi-release JAR might not be such a great idea as it's the Jetty version rather than Java that we're switching on and we want to support the case of Java 17 + Jetty 11 (someone out there will definitely be doing this!).

My apologies for confusion - the Jetty 12 won't be replacing Jetty 11, Jetty 11 would be default version to try independently of the JDK, the Jetty 12 would only be available on JDK-17+

> What I suggest is that we try to find a way to put all of stuff where the Jetty API varies under an abstraction in JettyHttpServer, create jetty11 and jetty12 packages with implementations subclassing this and a default implementation that selects and delegates to the appropriate one after querying Jetty.VERSION.

I will surely try that first, thank you.

tomakehurst:
I suggest we keep all sources under `src/main/java` and have `jetty11` and `jetty12` packages rather than having separate source trees for different Java versions. It should be fine to have a load of classes that depend on Jetty 12 provided they're only loaded when Jetty 12 is on the classpath.

tomakehurst:
Yeah, I guess that could be tricky. I wonder if we could put the compatibility tests in a sub-project that has Jetty 12 as its dependency, and keep Jetty 11 in the main project.

reta:
> Yeah, I guess that could be tricky. I wonder if we could put the compatibility tests in a sub-project that has Jetty 12 as its dependency, and keep Jetty 11 in the main project.

Exactly, I think we could do that by I haven't yet get to this point. If we find the way to phase off multi-release JAR, that would be easier, otherwise we would need to run Jetty 12 test suite against JAR (if nothing changed recently, none of build tools support running tests for multi-release artifacts if they are not packaged as JARs).

tomakehurst:
The Gradle team only seems to support multi-release very reluctantly.

And I really think we shouldn't be switching it on the Java version.

reta:
> And I really think we shouldn't be switching it on the Java version.

I was exploring the hosting the `jetty11` and `jetty12` under same source tree, but I haven't been able to pull off the compilation:  both `jetty11` and `jetty12` have many identical artifacts (fe `org.eclipse.jetty:jetty-server`, `org.eclipse.jetty:jetty-proxy`, version differs for sure) but absolutely incompatible APIs. So we may need stick to separate source sets.

However what I am exploring now, if we could merge the classes from both source sets (main + jetty12) once the compilation of both succeeds. In this case we don't need the multi-release JAR but single one. How does it sound to you?

tomakehurst:
Actually maybe having `src/jetty11` and `src/jetty12` source trees might be the solution. You'd comment out one of them while developing but then keep both, in addition to `src/java` when building so that both sets of classes end up in the JAR.

src/main/java/com/github/tomakehurst/wiremock/jetty/JettyHttpServerFactory.java
```diff
@@ -15,19 +15,63 @@
...
+import java.lang.invoke.MethodHandle;
+import java.lang.invoke.MethodHandles;
+import java.lang.invoke.MethodType;
+import java.lang.invoke.WrongMethodTypeException;
 
 public class JettyHttpServerFactory implements HttpServerFactory {
+  private static final MethodHandle SERVER_CONSTRUCTOR = getServerConstructor();
+
+  @SuppressWarnings("unchecked")
+  private static MethodHandle getServerConstructor() {
```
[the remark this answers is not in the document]
reta: > Can we avoid using reflection by querying Jetty.VERSION and then delegating to e.g. Jetty11HttpServerFactory or Jetty12HttpServerFactory appropriately?

Absolutely,  I haven't thought about that by I remember you suggesting it previously,  I will work on that part, thank you

tomakehurst:
Looking at your latest changes I'm starting to wonder if this is going to be a bit too painful to develop and maintain going forward.

Perhaps it might be worth considering a separate JAR after all. I've had a couple of ideas that might make it easier to live with:
1. Still detect the Jetty version in the core JAR and if it's 12 exit with a message indicating that the user should add the jetty12 JAR to their dependencies.
2. Use Java service loading to delegate to the appropriate server factory so e.g. when there's no service loader impl available we default to the built-in Jetty 11 factory, but if one is present we use this. Then the Jetty 12 JAR would register its factory as a service implementation and this would get loaded automatically when the JAR was present on the classpath.

We're planning to release WireMock on Thursday, although I'm guessing getting this in is a stretch at the moment.
Happy to assist with this if that would help.

BTW I appreciate you humouring me and trying to make this work the way I previously suggested!

reta:
Thanks for looking @tomakehurst, really appreciate it

> Perhaps it might be worth considering a separate JAR after all. I've had a couple of ideas that might make it easier to live with:

This would work (following the steps 1 and 2) but we have 2 choices here:
a) move project to multi-module (so jetty 11 & jetty 12 would be separate modules delegation to core)
b) have 2 jars (probably with qualifier?) out of the single build

> We're planning to release WireMock on Thursday, although I'm guessing getting this in is a stretch at the moment.
Happy to assist with this if that would help.

We are down to ~20 failing tests but large chunk of them is blocked by 2 issues (I also listed them on top but repeat it here for convinienc):
1. Jetty 12 does not allow to set reason / message along with set response status (HTTP 1.1 at least), it goes up to the raw response where reason / message is not pushed down (even if present)
2. org.eclipse.jetty.server.MultiPartInputStreamParser is gone for good

The 1st one is really difficult to get back - Jetty 12 literally phased out the status propagation from everywhere (if you need code hints, happy to share). For 2nd one we would need a replacement but I have not looked into it yet, if you have time to take it, would be awesome.

tomakehurst:
It looks like we can use `MultiPartFormData.Parser` in place of `MultiPartInputStreamParser` albeit with some work as the API isn't the same.

Agree status setting is tricky. I've been Googling a little bit to try to find out where and why they've done this, but without much luck so far so please share any links you've got to explanations of this.

reta:
> It looks like we can use `MultiPartFormData.Parser` in place of `MultiPartInputStreamParser` albeit with some work as the API isn't the same.

👍 (saw it, thank you)

> Agree status setting is tricky. I've been Googling a little bit to try to find out where and why they've done this, but without much luck so far so please share any links you've got to explanations of this.

I believe the context for it is Servlet 6 spec changes where `setStatus` could set only the status code. Under the hood, the `MetaData.Response` class is being used to carry over the status + reason BUT the `HttpChannelState` class, which prepares the response before sending it over the wire, passed `null` as the reason, always [1].

[1] https://github.com/jetty/jetty.project/blob/645e775114884b44c62a3df21088cef802254b33/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/internal/HttpChannelState.java#L1396

tomakehurst:
I suggest for now we prevent the reason setting tests from running when the Jetty version is 12.

We can document the fact it doesn't work with Jetty 12, and either drop support for it in WireMock 4.0 completely or see if we can persuade the Jetty team to restore this capability.

src/main/resources/META-INF/services/com.github.tomakehurst.wiremock.jetty.JettyHttpServerFactoryLoader
```diff
@@ -0,0 +1,2 @@
+com.github.tomakehurst.wiremock.jetty11.Jetty11HttpServerFactoryLoader
```
reta, line 1: Quick comment on why Jetty 11 and Jetty 12 bundled in the same service descriptor. Java uses lazy initialization of service loader providers, Jetty 11 comes first and as such, will be tried first.

tomakehurst:
Thanks for this latest update. I've found a simpler way to organise and build this. Will share a branch with you as soon as I've figured out why I've broken Archunit.

tomakehurst:
Here it is:
https://github.com/wiremock/wiremock/tree/reta-issue-2395-jetty-12

Are you in the Slack community BTW @reta ?

reta:
> Are you in the Slack community BTW @reta ?

Thanks a lot @tomakehurst , I am on vacation this week, should be back on March 7th (if you need me to take a look but please don't wait for me, I could review post merge), my apologies for inconvenience

reta:
> The main thing is I wanted to make sure you get properly credited by GitHub for all the work you've put in, which I'm not sure it'll do if I merge my branch.

I like your approach @tomakehurst  (I was a bit afraid to introduce subproject), it introduces much cleaner distribution and development models. Please feel free to go ahead, I have no issue with merging your branch, thanks a lot for working with me on that!

tomakehurst:
How about you pull my branch into this PR, then I'll squash and merge? That way you'll get properly credited by the OSS gods.

reta:
> How about you pull my branch into this PR, then I'll squash and merge? That way you'll get properly credited by the OSS gods.

Sure! No objections either, on it, thanks a lot @tomakehurst !

reta:
🟢 , we should be all set  (the documentation update would be a separate pull request), I had to remove JDK-11 from GA actions since we cannot build Jetty 12 subproject anymore with it

tomakehurst:
I think we should create separate GA workflows for the main project and Jetty12. Ensuring 11 support is too important to drop it completely from the matrix.

reta:
> I think we should create separate GA workflows for the main project and Jetty12. Ensuring 11 support is too important to drop it completely from the matrix.

Got it, will do that shortly, thanks @tomakehurst !

.github/workflows/release.yml
```diff
@@ -14,7 +14,7 @@ jobs:
       - name: Set up Java
         uses: actions/setup-java@v3
         with:
-          java-version: '11'
+          java-version: '17'
```
reta, line 17: Releasing with JDK-17 so to have wiremock-jetty12 included

settings.gradle
```diff
@@ -1 +1,4 @@
 rootProject.name = 'wiremock'
+if (JavaVersion.current() >= JavaVersion.VERSION_17) {
```
reta, line 2: @tomakehurst making `wiremock-jetty12` subproject conditional (based on JDK version)

reta:
> This is looking good. Is it ready to merge as far as you're concerned?

No concerns, thanks a lot @tomakehurst ! (resolved the conflicts)

reta:
@tomakehurst wondering if you have something in mind for this pull request that holds it off, thank you.

tomakehurst:
It's ready I think. Going to try to merge and release it today, along with a few other things.

Issue #1389, opened 2020-10-03
[Proxy] Empty Login page(s)


margasan:
# Bug report
Configuration:
- wiremock-jre8:2.27.2
- java 11.0.5 2019-10-15 LTS
- macOS Catalina 10.15.6 (host in which Wiremock server is launched)
- mobile browsers: Safary, Chrome, Android Webview, etc.
- Wiremock configuration:
`wireMockServer = new WireMockServer(wireMockConfig()
                .enableBrowserProxying(true)
                .trustAllProxyTargets(true)
                .port(Integer.parseInt(wireMockHttpPort))
                .bindAddress(wireMockAddress));`

Steps to reproduce:
1. Launch Wiremock as a browser proxy
2. Set proxying on mobile devices (configure ip:port of wiremock proxy, add and trust security certificate
3. Open in browser https://login.live.com/

Actual result:
Empty white page without login, password text fields, and other UI elements appears

Expected result:
Login page with login and password text fields and other UI elements appears

Notes:
1. This happens while open another login/entry point pages, for example, https://mobile.twitter.com
2. This issue is not reproduced with other proxies, like, Charles, Browsermob, etc.

![IMG_0023](https://user-images.githubusercontent.com/18184847/94987092-51af9380-056c-11eb-821d-679ce505ec9b.png)
)

margasan:
It looks like the issue is still not fixed. Reproduced using:
wiremock-jre8 version: '2.32.0'
Could you have a look?

tomakehurst:
Looks like it's related to #1543 this time

tomakehurst:
@margasan OK, looks like it isn't about the CORS headers after all.

When you tested this did you trust the WireMock root certificate you can grab from http://localhost:8080/__admin/recorder ?

margasan:
> When you tested this did you trust the WireMock root certificate you can grab from http://localhost:8080/__admin/recorder ?
Sure, the root CA is installed and trusted on the mobile device, so can to proxy all https traffic.

tomakehurst:
OK, having tested it again I've reverted to my original belief that it's about CORS headers.

Try setting `--enable-stub-cors` or `.stubCorsEnabled(true)` on WireMock's startup.

margasan:
> Try setting `--enable-stub-cors` or `.stubCorsEnabled(true)` on WireMock's startup.
Tried. No luck.

tomakehurst:
[...]

PR #2597, opened 2024-01-31
Disable connection reuse only for proxy clients not for the admin client

coder-hugo:
We are using WireMock for our integration tests. In one of our projects the test suite does ~ 29000 requests to the WireMock for stubbing, verifying, resetting ... requests. With the current implementation for each and every request a new TCP connection is created. This causes a lot of useless overhead. On powerful systems the test suite takes just 1 minute and a few seconds and it looks like we now reached a limit where some kind of ratelimiting (haven't yet found the reason for) on the network layer blocks the creation of new TCP connections to the WireMock. If I enable the connection reuse everything works as expected.
I just had a look through the history and it looks like disabling the connection reuse was introduced due to a bug that occurs when the WireMock server is running in proxy mode. Therefore this feature should only be disabled in these cases. (Btw. I don't think it's a good idea to mix-up the configuration of clients that are used for completely different use cases.)

To provide a test for this a added a small dummy server that just accepts a single connection. If you disable the connection reuse feature you will see it failing.

## References

- #1389

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/http/HttpClientFactory.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/http/HttpClientFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2023 Thomas Akehurst
+ * Copyright (C) 2011-2024 Thomas Akehurst
  *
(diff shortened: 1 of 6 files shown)


tomakehurst:
Thanks!

PR #2603, opened 2024-02-05
Disable XML External Entities

Pr0methean:
If an untrusted XML file is passed to one of the Xml class's public methods, XML External Entity injection attacks are possible. This PR fixes that by disabling all known features that would cause the DocumentBuilderFactory or TransformerFactory to load external entities and doctypes. It also uses `ThreadLocal::withInitial` to fix failing cases of `UnusedCodeTest`; note that this doesn't work for `InheritableThreadLocal` because that class has no equivalent method.

## References

- [OWASP XML External Entity Prevention Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/XML_External_Entity_Prevention_Cheat_Sheet.html)

## Submitter checklist


tomakehurst:
Thanks!

PR #2613, opened 2024-02-19
Revert "Fix json-body not escaped properly" 

tomakehurst:
This produces badly formed JSON where there are escaped quotes present. Given the potential for other undiscovered edge-cases, we'll release a patch version without this fix for now.


G-Basak:
@tomakehurst can you share an example when you get time?

And, let me know if you want me to take another try at this.

PR #2616, opened 2024-02-21
Added a System.Logger Notifier implementation and deprecated the Slf4 notifier

tomakehurst:
This attempts to address #2557 and more general concerns about slf4j versions and shading.

By removing WireMock's direct dependency on slf4j this frees the user to make their own arrangements and use `org:slf4j:slf4j-jdk-platform-logging` with their own library versions if needed.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/logging/JavaSystemNotifier.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/logging/JavaSystemNotifier.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014-2024 Thomas Akehurst
+ * Copyright (C) 2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


src/main/java/com/github/tomakehurst/wiremock/logging/JavaSystemNotifier.java
```diff
@@ -0,0 +1,49 @@
...
+
+  private final boolean verbose;
+
+  public JavaSystemNotifier(boolean verbose) {
+    this.verbose = verbose;
+  }
+
+  @Override
+  public void info(String message) {
+    if (verbose) {
```
ciscoo, line 35: Why not:

```java
if (logger.isLoggable(Level.INFO)) {
    // ...	
}
```

Let the user configure the logging system however they want it.

tomakehurst: If there was a lot of processing to happen inside the `if` block it'd be worth it, but since we just log the parameter value we're given there's no performance gain to be had really.

Having said that `if (verbose)` is basically redundant since that parameter is only used in standalone mode which defaults to the console notifier, so I'll remove that.

PR #2618, opened 2024-02-23
Fixed xPath concurrency issues

BublG:
Using InheritableThreadLocal means that entity can be accessible by some threads at the same time, so access should be synced or entity should be read-only. In the XmlDocument class there is a not synced write operation in xPath object that is InheritableThreadLocal.

I encountered the following problem. A am running a lot of requests to wiremock and mappings have different namespaces. And sometimes (random and rarely) at the evaluating xPath expression time (XmlDocument.findNodes() method) another thread (parent or child) is changing namespace context of current xPath object and error "Prefix must resolve to a namespace" appears, but namespaces in mapping file are correct. So I suggest to use simple ThreadLocal here. This change fixed my problem.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/xml/XmlNode.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/xml/XmlNode.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2020-2023 Thomas Akehurst
+ * Copyright (C) 2020-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


src/main/java/com/github/tomakehurst/wiremock/common/xml/XmlNode.java
```diff
@@ -39,14 +39,12 @@
...
-        }
-      };
+  protected static final ThreadLocal<XPath> XPATH_CACHE =
+      ThreadLocal.withInitial(
+          () -> {
+            final XPathFactory xPathfactory = XPathFactory.newInstance();
+            return xPathfactory.newXPath();
+          });
 
   protected static final InheritableThreadLocal<Transformer> TRANSFORMER_CACHE =
```
tomakehurst, line 49: Can we make the same change here too? Guessing this is also at risk from the same issue.

BublG: Yes let's do it. Transformer class is not thread-safe also according documentation

PR #2622, opened 2024-03-02
Replace org.apache.commons.lang3.StringUtils

pks-1981:
Proposal: Replace org.apache.commons.lang3.StringUtils


leeturner:
@pks-1981 Thanks for another contribution 👍

PR #2624, opened 2024-03-03
Replace net.jockx:littleproxy by JDK

pks-1981:
Replace net.jockx:littleproxy by JDK


leeturner:
Great stuff, thanks for the PR. Looks great

PR #2629, opened 2024-03-08
JSONPath performance tuning

tomakehurst:
Improves match performance when using matchesJsonPath in multiple stubs by per-request caching parsed JSON documents and values resolved via evaluated JSONPath expressions.

Also includes some other tunings based on profiler output.


src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java
```diff
@@ -102,6 +103,8 @@ public void handle(Request request, HttpResponder httpResponder, ServeEvent orig
 
     serveEvent.afterSend();
     afterResponseSent(serveEvent, response);
+
+    RequestCache.onRequestEnd();
```
dieppa, line 107: There is the risk that if, for any reason, an exception is thrown, the cache never gets to be cleaned.

Issue #2654, opened 2024-03-28
Provide a summary of logged requests so that a user can assert on them


tomasbjerre:
### Proposal

Given an application that that has several integrations and you are using Wiremock to mock those integrations.

A test case may do some stuff in an integration test and assert on how the integrations were invoked, with something like:

# Example 1

```java
    assertThat(WireMock.findAll(WireMock.anyRequestedFor(WireMock.anyUrl())))
      .hasSize(2);

    LoggedRequest aaaaaaRequest = WireMock.findAll(WireMock.anyRequestedFor(WireMock.urlEqualTo("/aaaaaa"))).get(0);
    assertThat(aaaaaaRequest.getBodyAsString())
      .isEqualToIgnoringWhitespace("whatever");

    LoggedRequest ooooooRequest = WireMock.findAll(WireMock.anyRequestedFor(WireMock.urlEqualTo("/oooooo"))).get(0);
    assertThat(ooooooRequest.getBodyAsString())
      .isEqualToIgnoringWhitespace("whatever");
```

There might be an `IndexOutOfBoundsException` on `WireMock.findAll(WireMock.anyRequestedFor(WireMock.urlEqualTo("/oooooo"))).get(0)` because the `2` requests might both be on `/aaaaaa`.

# Example 2

Perhaps you invoke your application in a way where it is important that integration `/aaaaaa` is not invoked. And you have a test like:

```java
    assertThat(WireMock.findAll(WireMock.anyRequestedFor(WireMock.urlEqualTo("/aaaaaa"))))
      .isEmpty();
```

And you implement a new feature where it is now important that `/oooooo` is never invoked and `/aaaaaa` is removed from the application. Now this assertion on `/aaaaaa` will still pass, it will pass forever because there is no such integration in the application any more. And it is very likely that the developer will not be aware of the assertion and not change it when replacing `/aaaaaa` with `/oooooo`.

# Proposed solution

I would like to assert on a summary of how Wiremock was invoked. Like this:

```java
    assertThat(findAll(anyRequestedFor(anyUrl())).requestSummary())
        .isEqualToIgnoringWhitespace("""
                1	| GET /aaaaaa
                1	| PUT /aaaaaa
                1	| DELETE /oooooo
                2	| GET /oooooo
                1	| DELETE /zzzzzz
                """);
```

That will make me aware of exactly what integrations were invoked and help avoid unwanted regressions.

I can make this assertion first, and after that make assertions on specific URL:s being invoked.

### References

[...]

PR #2657, opened 2024-03-29
feat: provide summary of HttpHeaders to be used in assertions (refs #2654)

tomasbjerre:
I want a compact summary of the headers that I can use in assertions.

I see there is a test case that serializes to JSON and asserts on that. I think that is a bit too verbose and not as readable as this alternative.

See motivation in the similar PR: #2658

## References

- #2654
- #2658

## Submitter checklist


Mahoney:
Could you explain why it's needed as well as the existing `toString()`, which returns something very similar?

tomasbjerre:
> Could you explain why it's needed

From the description: "I want a compact summary of the headers that I can use in assertions."

> as well as the existing `toString()`, which returns something very similar?

- I don't think it is obvious that the `toString` is intended to be used in assertions.
- I might be wrong, but the headers and their values are not sorted in the `toString`-method. I sorted the string in the summary to avoid having toggling test cases.

PR #2655, opened 2024-03-28
feat: summary of logged requests (refs #2654)

tomasbjerre:
## References

#2654 

## Submitter checklist

- [x] Recommended: Join [WireMock Slack](https://slack.wiremock.org/) to get any help in `#help-contributing` or a project-specific channel like `#wiremock-java`
- [x] The PR request is well described and justified, including the body and the references
- [x] The PR title represents the desired changelog entry
- [x] The repository's code style is followed (see the contributing guide)
- [x] Test coverage that demonstrates that the change works as expected
- [x] For new features, there's necessary documentation in this pull request or in a subsequent PR to [wiremock.org](https://github.com/wiremock/wiremock.org)

PR #2658, opened 2024-03-29
feat: provide summary of Cookies, Query and Form to be used in assertions (refs #2654)

tomasbjerre:
I want to assert on cookies, query and form in a compact, consistent and readable way that simplifies my test cases and makes them easier to maintain. I want clear errors when those test cases fail.

With this change I can do my assertions i a compact, consistent and readable way:

```java
    assertThat(findAll(anyRequestedFor(anyUrl())).get(0).getFormParameters().summary())
        .isEqualTo("...");

    assertThat(findAll(anyRequestedFor(anyUrl())).get(0).getCookies().summary())
        .isEqualTo("...");

    assertThat(findAll(anyRequestedFor(anyUrl())).get(0).getQueryParams().summary())
        .isEqualTo("...");
```

When they fail, the asserted string is changed which will give a clear error with a diff of how the string looked before and how it looks now.

## References

- #2654
- #2655
- #2657

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/requestfilter/RequestWrapper.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/requestfilter/RequestWrapper.java
@@ -178,7 +178,7 @@ public class RequestWrapper implements Request {
   }
 
   @Override
-  public Map<String, Cookie> getCookies() {
+  public MultiValues<Cookie> getCookies() {
     Map<String, Cookie> cookieMap = new HashMap<>();
     for (Map.Entry<String, Cookie> entry : delegate.getCookies().entrySet()) {
       Cookie newCookie =
@@ -193,7 +193,7 @@ public class RequestWrapper implements Request {
 
     cookieMap.putAll(additionalCookies);
 
-    return Collections.unmodifiableMap(cookieMap);
+    return new MultiValues<>(cookieMap);
   }
 
   @Override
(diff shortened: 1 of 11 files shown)


Mahoney:
This is a breaking change to the API of `Request`, which is a big deal.

What's wrong with the `toString()` on the existing `Map<String, Cookie>`?

tomasbjerre:
> Thanks for your contribution. Wrapping the cookies in a n object is a nice idea, but unfortunately it's a breaking change to the API of `Request`, which is a big deal.

- Note that it is still implements `Map<String, Cookie>`.
- I added another test case showing the serialization, looks like before.

How is it not backwards compatible?

> What's wrong with the `toString()` on the existing `Map<String, Cookie>`?

I might be wrong but I don't think that map is sorted. Not deterministic. Would lead to toggling test cases.

Mahoney:
> > Thanks for your contribution. Wrapping the cookies in a n object is a nice idea, but unfortunately it's a breaking change to the API of `Request`, which is a big deal.
> 
> * Note that it is still implements `Map<String, Cookie>`.
> * I added another test case showing the serialization, looks like before.
> 
> How is it not backwards compatible?

Apologies, missed it implementing `Map<String, Cookie>`.

tomasbjerre:
https://github.com/wiremock/wiremock/pull/2655#issuecomment-2033890656

PR #2671, opened 2024-04-06
JDK17 default

pks-1981:
JDK17 default


pks-1981:
@tomakehurst Maybe we should switch to 17 by default?

tomakehurst:
The plan is to keep Java 11 support until the next major version so I'm going to close this for now.

PR #2672, opened 2024-04-07
Remove deprecated java.security.AccessControl

pks-1981:
Remove deprecated java.security.AccessControl

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
@@ -24,21 +24,17 @@ import com.github.tomakehurst.wiremock.common.SilentErrorHandler;
 import java.io.IOException;
 import java.io.StringReader;
 import java.io.StringWriter;
-import java.security.AccessController;
-import java.security.PrivilegedAction;
 import javax.xml.XMLConstants;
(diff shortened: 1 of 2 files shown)


dieppa:
Thanks for your contribution.

Having a look to the official [JEP 411: Deprecate the Security Manager for Removal](https://openjdk.org/jeps/411), which says explicitly :

>Deprecate the Security Manager for removal in a future release. The Security Manager dates from Java 1.0. It has not been the primary means of securing client-side Java code for many years, and it has rarely been used to secure server-side code. To move Java forward, we intend to deprecate the Security Manager for removal in concert with the legacy Applet API ([JEP 398](https://openjdk.java.net/jeps/398)).


It provides an extensive explanation on the motivations and implications, but a summary is:
> The Security manager's goal was to protect agains two types of vulnerabilities: malicious intent by remote code and the threat of accidental vulnerabilities in local code. The threat of the first was mainly a concern with applets. Now applet are not supported and , so the protection from the Security Manager is no longer significant in this matter. 
Regarding the threat of accidental vulnerabilities in local code, it's almost impossible to address with the Security Manager. Many of the claims that the Security Manager is widely used to secure local code do not stand up to scrutiny; it is used far less in production than many people assume.


From all the above, we have decide to give green light to this change 😃

PR #2689, opened 2024-04-18
Fix some Sonar or/and "IDEA inspect code" issues (partly)

pks-1981:
Fix some Sonar or/and "IDEA inspect code" issues (partly)


dieppa (approved):
Thanks

PR #2703, opened 2024-05-02
Fix parsing of date/time query parameter values with an offset ahead of UTC

papiomytoglou:
According to the documentation on request matching, Wiremock should be able to handle offsets in date/time query parameter values in all of the following three cases:
```
/resource?date=2021-06-24T13:40:27+01:00
/resource?date=2021-06-24T13:40:27Z
/resource?date=2021-06-24T13:40:27-01:00
```
In other words, it should be able to parse a date/time query parameter value with an offset ahead of UTC (`+HH:mm`), an offset of zero (`Z`) and an offset behind UTC (`-HH:mm`).

Right now only the last two work. Offsets ahead of UTC found in query parameters cannot be matched due to the `+` also having semantics in URL encoding. That is, in URL encoded strings `+` represents a space and as a result `2021-06-24T13:40:27+01:00` is decoded to `2021-06-24T13:40:27 01:00` which results in a `DateTimeParseException`.

To avoid that, this PR adds a check to see whether the parameter's value is an ISO-8601 formatted date/time value and, if so, returns it without decoding it first. That way `2021-06-24T13:40:27+01:00`  remains as is while `hello+world` is still decoded to `hello world`. In other words, the fix only applies to strings that can successfully be parsed by `ISO_OFFSET_DATE_TIME`, everything else is treated the same as before.

A helpful discussion of the problem above is available in the note found [here](https://developer.apple.com/documentation/foundation/nsurlcomponents/1407752-queryitems). The context of the note is Apple technologies, not Java, but this problem is independent of the technologies used.

A unit test which breaks with the current implementation has also been added.

## References
Wiremock documentation where the expected behaviour is described: https://wiremock.org/docs/request-matching/#local-vs-zoned

## Submitter checklist

- [ ] For new features, there's necessary documentation in this pull request or in a subsequent PR to [[wiremock.org](http://wiremock.org/)](https://github.com/wiremock/wiremock.org)

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
@@ -18,6 +18,7 @@ package com.github.tomakehurst.wiremock.common;
 import static com.github.tomakehurst.wiremock.common.Exceptions.throwUnchecked;
 import static com.github.tomakehurst.wiremock.common.Strings.ordinalIndexOf;
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.time.format.DateTimeFormatter.ISO_OFFSET_DATE_TIME;
 
 import com.github.tomakehurst.wiremock.http.QueryParameter;
 import com.google.common.collect.ImmutableListMultimap;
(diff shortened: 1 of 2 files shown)


dieppa (approved):

dieppa:
Thanks @papiomytoglou ,

Reviewed and merged!

PR #2717, opened 2024-05-14
Fixes #2696 - xPath helper returns error when result is primitive value

### LINKED ISSUE #2696: xPath helper doesn't return result when XPath expression returns a single string item
### Proposal

Expressions that return a single string node, but the xPath helper expects a collection and currently treats anything else as an invalid expression.

### Reproduction steps

xPath expressions like `local-name(/*/*/*[1])` trigger this issue.

### References

_No response_

tomakehurst:


leeturner (approved):
👍

PR #2725, opened 2024-05-17
Join - Handlebar helper

dieppa:
Requested issue: https://github.com/wiremock/wiremock/issues/2710

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
@@ -29,7 +29,7 @@ public class JoinHelper extends HandlebarsHelper<Object> {
       return handleError("Separator parameter must be a String");
     }
 
-    String separator = context.toString();
+    String separator = (String)context;
 
     Object param = options.param(0, null);
     if (param == null || !(Iterable.class.isAssignableFrom(param.getClass()))) {
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
@@ -15,19 +15,27 @@
  */
 package com.github.tomakehurst.wiremock.extension.responsetemplating.helpers;
 
-import com.github.jknack.handlebars.Helper;
 import com.github.jknack.handlebars.Options;
 import com.github.jknack.handlebars.TagType;
 import java.io.IOException;
 import java.util.List;
 
-public class JoinHelper implements Helper<Object> {
+public class JoinHelper extends HandlebarsHelper<Object> {
 
   @Override
   public Object apply(Object context, Options options) throws IOException {
 
+    if (context == null || !(context instanceof String)) {
+      return handleError("Separator parameter must be a String");
+    }
+
     String separator = context.toString();
-    List<Object> items = options.param(0, null);
+
+    Object param = options.param(0, null);
+    if (param == null || !(Iterable.class.isAssignableFrom(param.getClass()))) {
+      return handleError("The parameters must be list");
+    }
+    List<Object> items = (List<Object>) param;
     if (items == null || items.isEmpty()) {
       return "";
     }
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2019-2021 Thomas Akehurst
+ * Copyright (C) 2019-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
```diff
@@ -0,0 +1,79 @@
...
+import com.github.jknack.handlebars.TagType;
+import java.io.IOException;
+import java.util.List;
+
+public class JoinHelper extends HandlebarsHelper<Object> {
+
+  @Override
+  public Object apply(Object context, Options options) throws IOException {
+
+    if (context == null || !(context instanceof String)) {
```
leeturner, line 28: Don't technically need the `context == null` here with the other condition

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
```diff
@@ -0,0 +1,79 @@
...
+      return "";
+    }
+    String result;
+    if (options.tagType == TagType.SECTION) {
+      result = processSection(options, separator, items);
+    } else {
+      result = processWithoutSection(separator, items);
+    }
+
+    return result.replaceAll("\n", "");
```
leeturner, line 49: I wasn't sure what the purpose of this was.  Is there a reason we wouldn't want to let people have new line characters in their output?  For example, this test would fail and I am not sure why it is an issue? 

```java
  @Test
  public void joinWithArrayOfStringsWithNewLineCharacters() {
    String result = transform("{{join ',' (array 'One\n' 'Two\n' 'Three\n')}}");

    assertThat(result, equalToCompressingWhiteSpace("One\n,Two\n,Three\n"));
  }
```

leeturner:
This all looks super cool.  Really like it.  Very useful helper.  While I was reviewing I noticed that the tests were limited to joining only with `,` which might be the biggest usecase for this helper to be fair.  I wondered wheather adding a few more variants would be good:

```java
  @Test
  public void joinWithArrayOfInts() {
    String result = transform("{{join '*' (array 1 2 3)}}");

    assertThat(result, equalToCompressingWhiteSpace("1*2*3"));
  }

  @Test
  public void joinWithBlankSeparator() {
    String result = transform("{{join ' ' (array 'WireMock' 'Rocks')}}");

    assertThat(result, equalToCompressingWhiteSpace("WireMock Rocks"));
  }

  @Test
  public void joinWithEmptySeparatorAndArrayOfChars() {
    String result = transform("{{join '' (array 'W' 'i' 'r' 'e' 'M' 'o' 'c' 'k' ' ' 'R' 'o' 'c' 'k' 's')}}");

    assertThat(result, equalToCompressingWhiteSpace("WireMock Rocks"));
  }

  @Test
  public void joinWithArrayOfStringsAndStringSeparator() {
    String result = transform("{{join \" - * - \" (array 'One' 'Two' 'Three')}}");

    assertThat(result, equalToCompressingWhiteSpace("One - * - Two - * - Three"));
  }

  @Test
  public void joinWithArrayOfStringsAndSeparatorFromHandlebars() {
    String result = transform("{{join (pickRandom ':') (array 'One' 'Two' 'Three')}}");

    assertThat(result, equalToCompressingWhiteSpace("One:Two:Three"));
  }
```

And also, because one of the usecases we want from this is to be able to more easily generate valid `json` I would probably update this test just to make it output valid json:

```java
  @Test
  public void joinWithObjectBody() {
    String result =
        transform(
            "{{#parseJson 'myThings'}}\n"
                + "[\n"
                + "  { \"id\": 1, \"name\": \"One\" },\n"
                + "  { \"id\": 2, \"name\": \"Two\" },\n"
                + "  { \"id\": 3, \"name\": \"Three\" }\n"
                + "]\n"
                + "{{/parseJson}}"
                + "[{{#join ',' myThings as |item|}}"
                + "{\n"
                + "\"name{{item.id}}\": \"{{item.name}}\"\n"
                + "}\n"
                + "{{/join}}]");

    assertThat(result, equalToCompressingWhiteSpace("[{\"name1\": \"One\"},{\"name2\": \"Two\"},{\"name3\": \"Three\"}]"));
  }
```

tomakehurst:
Agree with @leeturner that some other variations in the tests would be good. Otherwise I'm happy with this.

dieppa:
Yeah, I agree, adding some variations in the separator 👌🏼

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/JoinHelper.java
dieppa: On the other hand, if we do this:
```
{{#join ',' myData |as| item}}
{
  "name": "{{item.name}}"
}
{{/join}}
```

We would like to obtain this `{name2:One},{name2:Two},{name2:Three}` , but instead, we obtain this:
```
{
name2:One
}
,
{
name2:Two
}
,
{
name2:Three
}
```

dieppa: Actually, for some reason I was assuming it should be in one line, but checking what `each` does, it prints what it's in there without removing anything, printing it in multiple lines. So, removing it 😃

dieppa:
I added all the test for multiple separators in a unique test method. Normally I don't like this, but the code looks clear and concise 😃

tomakehurst:
Can we have a better name for the PR before merging, for the sake of the release note?

PR #2737, opened 2024-05-23
Disabled DNS filtering when no custom proxy security rules are set

egerix:
## References

attempt to fix https://github.com/wiremock/wiremock/issues/2736

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/DefaultNetworkAddressRules.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/DefaultNetworkAddressRules.java
@@ -15,7 +15,7 @@
  */
 package com.github.tomakehurst.wiremock.common;
 
-import static com.github.tomakehurst.wiremock.common.NetworkAddressRange.ALL;
+import static com.github.tomakehurst.wiremock.common.NetworkAddressRange.ALL_RANGES;
 import static com.github.tomakehurst.wiremock.common.NetworkAddressUtils.isValidInet4Address;
 import static java.util.stream.Collectors.toSet;
 
@@ -37,7 +37,7 @@ public class DefaultNetworkAddressRules implements NetworkAddressRules {
                     networkAddressRange ->
                         !(networkAddressRange instanceof NetworkAddressRange.DomainNameWildcard))
                 .collect(toSet()),
-            Set.of(ALL));
+            ALL_RANGES);
     this.allowedHostPatterns =
         defaultIfEmpty(
             allowed.stream()
@@ -45,7 +45,7 @@ public class DefaultNetworkAddressRules implements NetworkAddressRules {
                     networkAddressRange ->
                         (networkAddressRange instanceof NetworkAddressRange.DomainNameWildcard))
                 .collect(toSet()),
-            Set.of(ALL));
+            ALL_RANGES);
     this.denied =
         denied.stream()
             .filter(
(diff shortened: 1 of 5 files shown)


src/main/java/com/github/tomakehurst/wiremock/common/DefaultNetworkAddressRules.java
```diff
@@ -80,4 +80,12 @@ public boolean isAllowed(String testValue) {
           && deniedHostPatterns.stream().noneMatch(rule -> rule.isIncluded(testValue));
     }
   }
+
+  @Override
+  public boolean isAllowedAll() {
+    return allowed.equals(Set.of(ALL))
+        && allowedHostPatterns.equals(Set.of(ALL))
+        && denied.equals(Set.of())
```
Mahoney, line 88: Just `denied.isEmpty()`?
Same in the line below.

src/main/java/com/github/tomakehurst/wiremock/common/DefaultNetworkAddressRules.java
```diff
@@ -80,4 +80,12 @@ public boolean isAllowed(String testValue) {
           && deniedHostPatterns.stream().noneMatch(rule -> rule.isIncluded(testValue));
     }
   }
+
+  @Override
+  public boolean isAllowedAll() {
+    return allowed.equals(Set.of(ALL))
```
Mahoney, line 86: I wonder if we should pull out `Set.of(ALL)` as a constant, we've now got it 5 times... not sure what to call it though, since `ALL` is taken.

Mahoney (approved):
Nitpicking aside, looks like a safe workaround to me until we can implement security rules for IPv6.

src/main/java/com/github/tomakehurst/wiremock/common/DefaultNetworkAddressRules.java
egerix: Thank you for the suggestion. I agree that extracting Set.of(ALL) as a constant would improve the readability of the code.
Regarding the placement of the constant, would you prefer it to be in the `NetworkAddressRange` class or should we keep it within the `DefaultNetworkAddressRules` class?

 I was thinking we could name it `ALL_RANGES = Set.of(ALL);`

egerix: > I wonder if we should pull out `Set.of(ALL)` as a constant, we've now got it 5 times... not sure what to call it though, since `ALL` is taken.

Thank you for the suggestion. I agree that extracting Set.of(ALL) as a constant would improve the readability of the code.
Regarding the placement of the constant, would you prefer it to be in the NetworkAddressRange class or should we keep it within the DefaultNetworkAddressRules class?

 I was thinking we could name it `ALL_RANGES = Set.of(ALL);`

egerix:
@Mahoney, could you approve the workflow? Or should it be done later?

src/main/java/com/github/tomakehurst/wiremock/common/DefaultNetworkAddressRules.java
tomakehurst: I think this is a good name and I suggest we put it in `NetworkAddressRange`

egerix: e450d5ecebe8e96912d491a01de0397156345905

tomakehurst:
Thanks!

PR #2741, opened 2024-05-28
Add support for request matching with dynamic year and dynamic year/month values

papiomytoglou:
Currently Wiremock supports request matching with dynamic date/time values but not with dynamic year values. In other words, given the following matcher:
```
"request": {
  "method": "GET",
  "urlPath": "/resource",
  "queryParameters": {
    "year": {
      "equalToDateTime": "now",
      "truncateExpected": "first day of year"
    }
  }
}
```
A GET request to the following URI does not match (assuming the current year is 2024):
```
/resource?year=2024
```
Even though a GET request to the following three URIs matches (assuming the current year is 2024 and the system timezone is UTC):
```
/resource?year=2024-01-01T00:00:00.000Z
/resource?year=2024-01-01T00:00:00.000
/resource?year=2024-01-01
```
That is because the current implementation attempts to parse the actual date/time value using  `ZonedDateTime`, `LocalDateTime` and `LocalDate` consecutively and none of these classes can parse a year value in the `yyyy` format.

This PR introduces a very small fix which allows the actual value to be parsed using `Year` in case parsing fails with all three aforementioned classes. A unit test which fails with the current implementation is also added.

## References
Original enhancement request for request matching with dynamic dates: https://github.com/wiremock/wiremock/issues/1475

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/DateTimeParser.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/DateTimeParser.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021 Thomas Akehurst
+ * Copyright (C) 2021-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -111,6 +111,22 @@ public class DateTimeParser {
   public LocalDate parseLocalDate(String dateTimeString) {
     if (dateTimeFormatter != null) {
       return LocalDate.parse(dateTimeString, dateTimeFormatter);
+    }
+
+    return null;
+  }
+
+  public YearMonth parseYearMonth(String dateTimeString) {
+    if (dateTimeFormatter != null) {
+      return YearMonth.parse(dateTimeString, dateTimeFormatter);
+    }
+
+    return null;
+  }
+
+  public Year parseYear(String dateTimeString) {
+    if (dateTimeFormatter != null) {
+      return Year.parse(dateTimeString, dateTimeFormatter);
     }
 
     return null;
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/AbstractDateTimePattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/AbstractDateTimePattern.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2022 Thomas Akehurst
+ * Copyright (C) 2021-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,9 +22,7 @@ import com.github.tomakehurst.wiremock.common.DateTimeOffset;
 import com.github.tomakehurst.wiremock.common.DateTimeParser;
 import com.github.tomakehurst.wiremock.common.DateTimeTruncation;
 import com.github.tomakehurst.wiremock.common.DateTimeUnit;
-import java.time.LocalDate;
-import java.time.LocalDateTime;
-import java.time.ZonedDateTime;
+import java.time.*;
 import java.time.format.DateTimeParseException;
 import java.util.List;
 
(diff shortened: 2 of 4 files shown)


RafeArnold:
this looks like a more broad limitation of the datetime matchers. currently, any datetime supplied needs to be in a format that can be parsed to a full `LocalDateTime` or `ZonedDateTime`. thus, any dates in the following formats or similar will not be able to be matched: `yyyy`, `yyyy-MM`, `yyyy-MM-dd`. a possible solution for this would be to use the truncation value to determine what datetime type to parse the datetimes into. for instance, if the truncation value for the actual datetime is `FIRST_DAY_OF_THE_YEAR`, we know we only need to parse the actual value to a `Year`, whereas if it's `FIRST_DAY_OF_MONTH`, we only need to parse the actual value to a `YearMonth`. i may experiment with this when i have a moment.

papiomytoglou:
Hello @RafeArnold, thank you for your response.

You are correct to point out that my proposal doesn't take under account `yyyy-MM` for which `YearMonth` would be needed. I will amend this PR.

On the other hand, you are incorrect as far as `yyyy-MM-dd` is concerned. The current implementation supports that format via the `LocalDate` class. As I mentioned in the original message. the current sequence of parsing attempts is as follows:
```
ZonedDateTime > LocalDateTime > LocalDate
```
I think adding `YearMonth` and `Year` to the above covers all cases. So the final sequence should be:
```
ZonedDateTime > LocalDateTime > LocalDate > YearMonth > Year
```
As for using the truncation to determine which parser to use, I don't think it would work since a truncation is not guaranteed to be present. Imagine we only want to match a specific date, a specific month or a specific year. No truncation would be needed in that case:
```
"request": {
  "method": "GET",
  "urlPath": "/resource",
  "queryParameters": {
    "month": {
      "equalToDateTime": "2024-07"
    }
  }
}
```
This is the matcher for July 2024 and only July 2024. A truncation isn't needed in this case.

RafeArnold:
> Hello @RafeArnold, thank you for your response.
> 
> You are correct to point out that my proposal doesn't take under account `yyyy-MM` for which `YearMonth` would be needed. I will amend this PR.
> 
> On the other hand, you are incorrect as far as `yyyy-MM-dd` is concerned. The current implementation supports that format via the `LocalDate` class. As I mentioned in the original message. the current sequence of parsing attempts is as follows:
> 
> ```
> ZonedDateTime > LocalDateTime > LocalDate
> ```
> 
> I think adding `YearMonth` and `Year` to the above covers all cases. So the final sequence should be:
> 
> ```
> ZonedDateTime > LocalDateTime > LocalDate > YearMonth > Year
> ```
> 
> As for using the truncation to determine which parser to use, I don't think it would work since a truncation is not guaranteed to be present. Imagine we only want to match a specific date, a specific month or a specific year. No truncation would be needed in that case:
> 
> ```
> "request": {
>   "method": "GET",
>   "urlPath": "/resource",
>   "queryParameters": {
>     "month": {
>       "equalToDateTime": "2024-07"
>     }
>   }
> }
> ```
> 
> This is the matcher for July 2024 and only July 2024. A truncation isn't needed in this case.

Yes, you're correct on all fronts. Using try/catch for control flow is really unfortunate. I wonder if there is a nicer way to achieve the same result.

papiomytoglou:
I don't think the try/catch can be avoided but it can be extracted in a separate method returning a boolean. That's what I did in [this earlier PR](https://github.com/wiremock/wiremock/pull/2703/files#diff-7f83e1a3ccd51da01b797c6b602cc2356ef4be192b4e6d1fc9f0b7f4c117d832R109-L102).

papiomytoglou:
I tried extracting separate methods to hide the try/catch blocks inside. So something along those lines:
```
private static Optional<LocalDateTime> parseYearMonth(String dateTimeString) {
  try {
    return Optional.of(Year.parse(dateTimeString).atDay(1).atStartOfDay());
  } catch (DateTimeParseException ignored) {
    return Optional.empty();
  }
}

private static Optional<LocalDateTime> parseYear(String dateTimeString) {
  try {
    return Optional.of(YearMonth.parse(dateTimeString).atDay(1).atStartOfDay());
  } catch (DateTimeParseException ignored) {
    return Optional.empty();
  }
}
```
But this solution soon started getting overengineered so I reverted back to using try/catch within the `parseLocalOrNull` method. It's not ideal but it's much more succinct.

Please let me know if you have an alternative in mind.

RafeArnold:
looks good. can support be added for custom formats too, like we already do for localdatetime and localdate?

papiomytoglou:
Good point, @RafeArnold, totally missed that. I have included support for custom formatters in the latest iteration. Also added two more unit tests.

RafeArnold:
can you add tests to EqualToDateTimePatternTest.java for the custom formats, to assert the new parseYear and parseYearMonth methods are being used by the pattern matcher?

papiomytoglou:
Addressed in latest iteration, @RafeArnold.

RafeArnold (approved):
looks great. thanks for your contribution @papiomytoglou !

PR #2742, opened 2024-05-29
Prevent creating multiple stub mappings with the same ID

RafeArnold:
This pull request implements the fix outlined in [this issue](https://github.com/wiremock/wiremock/issues/2734) to prevent the creation of multiple stub mappings with the same ID.

## References

https://github.com/wiremock/wiremock/issues/2734

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/StubbingAcceptanceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/StubbingAcceptanceTest.java
@@ -34,8 +34,10 @@ import static org.hamcrest.Matchers.containsString;
 import static org.hamcrest.Matchers.hasItem;
 import static org.hamcrest.Matchers.is;
 import static org.hamcrest.Matchers.notNullValue;
+import static org.hamcrest.Matchers.nullValue;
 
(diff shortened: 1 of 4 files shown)


src/test/java/com/github/tomakehurst/wiremock/StubbingAcceptanceTest.java
```diff
@@ -1221,6 +1224,31 @@ void queryParamCanBeMatchedAsNotAbsent() {
     assertThat(testClient.get("/search").statusCode(), is(404));
   }
 
+  @Test
+  void cannotCreateMultipleStubMappingsWithTheSameId() {
```
leeturner, line 1228: Is it also worth adding a coresponding test to the `AdminApiTest` to show the errors we get in the admin API when we use a duplicate id ?

RafeArnold: Yes, I think that's a good idea. This test already makes its create request via the admin HTTP client, so is basically already making its assertions on the errors returned by the admin API, but if you think a more explicit test would be worth it, I will add it.

leeturner (approved):

PR #2747, opened 2024-06-05
Ignore order of xml nodes on same level when matching request body

viktormelnychuk:
Ignores the order of nodes on the same level by applying secondary sorting by the text content of the node. This allows to match requests where there can be multiple same tags in arbitrary order
Given example configuration:
```java
    .withRequestBody(equalToXml("<body>" +
            "   <entry>1</entry>" +
            "   <entry>2</entry>" +
            "</body>", false, true))
```
Will match
```xml
<body>
    <entry>2</entry>
    <entry>1</entry>
</body>
```
and match 
```xml
<body>
    <entry>1</entry>
    <entry>2</entry>
</body>
```

Added property to enable secondary sort by node text content. The property name is `ignoreOrderOfSameNode`. Tried to follow pattern similar to `equalToJson` implementation. Update existing `equalToXml` static methods to pass `null` value as the parameter to preserve existing behavior. Added two additional `equalToXml` overloads that accept the `ignoreOrderOfSameNode` parameter:
where we don't use placeholder delimiters
```java
  public static EqualToXmlPattern equalToXml(
      String value, boolean enablePlaceholders, boolean ignoreOrderOfSameNode) {
    return new EqualToXmlPattern(value, enablePlaceholders, ignoreOrderOfSameNode);
  }
```
and 
```java
  public static EqualToXmlPattern equalToXml(
      String value,
      boolean enablePlaceholders,
      String placeholderOpeningDelimiterRegex,
      String placeholderClosingDelimiterRegex,
      boolean ignoreOrderOfSameNode) {
    return new EqualToXmlPattern(
        value,
        enablePlaceholders,
        placeholderOpeningDelimiterRegex,
        placeholderClosingDelimiterRegex,
        null,
        ignoreOrderOfSameNode);
  }
```
where we do use placeholder delimiters. 

Docs are also updated and PR raised to reflect changes: 
## References

- TODO

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -256,6 +256,7 @@ public class WireMock {
   public static EqualToXmlPattern equalToXml(String value, boolean enablePlaceholders) {
     return new EqualToXmlPattern(value, enablePlaceholders, null, null, null, false);
   }
+
   public static EqualToXmlPattern equalToXml(
       String value, boolean enablePlaceholders, boolean ignoreOrderOfSameNode) {
     return new EqualToXmlPattern(value, enablePlaceholders, ignoreOrderOfSameNode);
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -254,7 +254,11 @@ public class WireMock {
   }
 
   public static EqualToXmlPattern equalToXml(String value, boolean enablePlaceholders) {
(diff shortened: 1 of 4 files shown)


tomakehurst:
This is a breaking change as it stands. I think we'd need to introduce a parameter similar to the ignore order one in `equalToJson` to avoid this.

viktormelnychuk:
@tomakehurst  Hi!
I changed the implementation per your recommendations. Added a parameter and tried to follow the `equalToJson` implementation

tomakehurst:
Please can you run `./gradlew spotlessApply` and push the results?

viktormelnychuk:
Run it, sorry

PR #2755, opened 2024-06-16
Jetty 12: Restore Cross-Context Dispatch

reta:
While implementing Jetty 12 support (https://github.com/wiremock/wiremock/pull/2593), it was necessary to change the way `NotFoundHandler` was configured due to the fact that cross-context dispatch was removed. However, since the  Jetty `12.0.8` release, the cross-context dispatch was reintroduced. 

## References

- https://webtide.com/back-to-the-future-with-cross-context-dispatch/0/
- https://github.com/wiremock/wiremock/pull/2593

## Submitter checklist


tomakehurst:
What's the benefit of returning to this approach, given that you got it to work without?

reta:
> What's the benefit of returning to this approach, given that you got it to work without?

Thanks for looking @tomakehurst , I think functionally-wise there is no benefits really, but implementation-wise the code becomes cleaner: the mock context explicitly needed an admin context for initialization (so we could simulate cross context dispatch), that is gone now.

PR #2757, opened 2024-06-17
Add request id to the template model

leeturner:
This PR adds the unique serveEvent Id to the `request` template model and also exposes the same id in the webhook `originalRequest` template model.

## References

resolves https://github.com/wiremock/wiremock/issues/2756

## Submitter checklist

### DIFF
--- a/src/main/java/org/wiremock/webhooks/Webhooks.java
+++ b/src/main/java/org/wiremock/webhooks/Webhooks.java
@@ -26,7 +26,6 @@ import com.github.tomakehurst.wiremock.extension.Parameters;
 import com.github.tomakehurst.wiremock.extension.PostServeAction;
 import com.github.tomakehurst.wiremock.extension.ServeEventListener;
 import com.github.tomakehurst.wiremock.extension.WireMockServices;
-import com.github.tomakehurst.wiremock.extension.responsetemplating.RequestTemplateModel;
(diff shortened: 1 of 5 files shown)


RafeArnold (approved):

src/main/java/org/wiremock/webhooks/Webhooks.java
```diff
@@ -128,7 +128,7 @@ private WebhookDefinition applyTemplating(
         webhookDefinition.getExtraParameters() != null
             ? webhookDefinition.getExtraParameters()
             : Collections.<String, Object>emptyMap());
-    model.put("originalRequest", RequestTemplateModel.from(serveEvent.getRequest()));
+    model.put("originalRequest", RequestTemplateModel.from(serveEvent));
```
tomakehurst, line 131: Could we inject the template engine and call that to build the model instead?

That way we also get the extra model elements provided by extensions.

leeturner: Sounds good.  Will take a look

leeturner: @tomakehurst this would add the `request` key to the model along side the `originalRequest`.  I assume this wasn't added originally for a reason.  Would we want to remove this from the model that the webhooks use?

PR #2766, opened 2024-06-19
2759 report webhook errors and results as sub events

leeturner:
This PR adds `INFO` sub events for the successful webhook request/response as well as `ERROR` sub events for any error conditions that occur. Requests and Responses are logged as a `LoggedRequest` and `LoggedResponse` where the `LoggedResponse` body limit is taken from the configured `maxLoggedResponseSize`

## References

impements https://github.com/wiremock/wiremock/issues/2759

## Submitter checklist

### DIFF
--- a/src/main/java/org/wiremock/webhooks/Webhooks.java
+++ b/src/main/java/org/wiremock/webhooks/Webhooks.java
@@ -38,6 +38,8 @@ import java.util.concurrent.ScheduledExecutorService;
 @SuppressWarnings("deprecation") // maintaining PostServeAction for backwards compatibility
 public class Webhooks extends PostServeAction implements ServeEventListener {
 
+  private static final String WEBHOOK_REQUEST_SUB_EVENT_NAME = "WEBHOOK_REQUEST";
+
   private final ScheduledExecutorService scheduler;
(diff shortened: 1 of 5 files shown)


src/main/java/org/wiremock/webhooks/Webhooks.java
```diff
@@ -80,8 +84,12 @@ private void triggerWebhook(ServeEvent serveEvent, Parameters parameters) {
       }
       definition = applyTemplating(definition, serveEvent);
       request = buildRequest(definition);
+
+      serveEvent.appendSubEvent(SubEvent.INFO, LoggedRequest.createFrom(request));
     } catch (Exception e) {
-      notifier().error("Exception thrown while configuring webhook", e);
+      final String msg = "Exception thrown while configuring webhook";
+      notifier().error(msg, e);
```
tomakehurst, line 91: Suggest we don't add a notifier() call here since we're intending to phase this out and instead log the current request's sub-events instead.

leeturner: After further discussion we decided to leave the notifer in place as it was there prior to this PR

Issue #2758, opened 2024-06-18
Add self-description callback(s) to MatchResult for use in Diff


tomakehurst:
### Proposal

Currently Diff and its related classes knows far too much about how to render specific matchers' info and results.

Additionally, custom matchers can't report their reason for not matching into a Diff so the best we can get is their names.

Propose we add self-description (a la Hamcrest?) to MatchResult and refactor Diff to use this info when rendering a report.

PR #2771, opened 2024-06-21
2758 self description matchresult

leeturner:
This PR introduces a new capability to `MatchResult` class, enabling it to hold self-description class to be used in diff reporting. A new `DiffDescriptionLine` class has been introduced to allow a gradual opt in to diff reporting using the new `DiffDescription` class.

This enhancement impacts the `WeightedAggregateMatchResult`, `MemoizingMatchResult` and `EagerMatchResult` classes as they now can take an optional `DiffDescription` class.  All existing APIs should have remained the same.

This feature is in Beta while we try this out with custom matchers to determine whether we want to roll this out to the internal WireMock matchers

## References

implements #2758 

## Submitter checklist


src/main/java/com/github/tomakehurst/wiremock/matching/MemoizingMatchResult.java
```diff
@@ -43,6 +43,11 @@ public Boolean get() {
   private final MatchResult target;
 
   public MemoizingMatchResult(MatchResult target) {
+    this(target, null);
+  }
+
+  public MemoizingMatchResult(MatchResult target, DiffDescription diffDescription) {
```
tomakehurst, line 49: This doesn't look like it should be here to me. This memoizing class is just a wrapper so the specifics of the match result should be entirely defined by the object being wrapped I think.

src/main/java/com/github/tomakehurst/wiremock/matching/WeightedAggregateMatchResult.java
```diff
@@ -31,6 +31,12 @@ public class WeightedAggregateMatchResult extends MatchResult {
   private final Lazy<Pair<Boolean, List<SubEvent>>> resultAndEvents;
 
   public WeightedAggregateMatchResult(List<WeightedMatchResult> matchResults) {
+    this(matchResults, null);
+  }
+
+  public WeightedAggregateMatchResult(
+      List<WeightedMatchResult> matchResults, DiffDescription diffDescription) {
```
tomakehurst, line 38: I'm wondering if we should aggregate all the diff descriptions somehow rather than passing a single one in, as I'm not sure how we'd assign a single diff description to an aggregate. Possibly warrants some further discussion.

src/main/java/com/github/tomakehurst/wiremock/verification/diff/DiffDescriptionLine.java
```diff
@@ -0,0 +1,67 @@
...
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.verification.diff;
+
+import com.github.tomakehurst.wiremock.matching.MatchResult;
+import com.github.tomakehurst.wiremock.matching.MatchResult.DiffDescription;
+import com.github.tomakehurst.wiremock.matching.NamedValueMatcher;
+import org.wiremock.annotations.Beta;
+
+@Beta(
```
tomakehurst, line 23: Don't think we need this here as this is an internal class anyway.

src/main/java/com/github/tomakehurst/wiremock/matching/WeightedAggregateMatchResult.java
tomakehurst: Maybe we can get away with returning null here for now since aggregate results are never used in diff reports.

PR #2788, opened 2024-07-02
Add support for reversing order of offsetting and truncating expected date/time values

papiomytoglou:
Given the matcher below:
```
"request": {
  "method": "GET",
  "urlPath": "/resource",
  "queryParameters": {
    "date": {
      "equalToDateTime": "now +1 months",
      "truncateExpected": "last day of month"
    }
  }
}
```
One would expect for the `last day of month` truncation to be applied to the `now +1 months` expected date/time value.

Instead, what currently happens is the `last day of month` truncation is applied to `now` and the result is then offset by `+1 months`.

This is very counterintuitive and not what someone would expect by looking at the matcher above.

This PR reverses the order of offsetting and truncating expected date/time values with the truncations now being applied last.

## References
A more in depth discussion of the issue can be found in the enhancement proposal: #2789

## Submitter checklist

- [ ] For new features, there's necessary documentation in this pull request or in a subsequent PR to [[wiremock.org](http://wiremock.org/)](https://github.com/wiremock/wiremock.org)

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/AbstractDateTimePattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/AbstractDateTimePattern.java
@@ -35,9 +35,10 @@ public abstract class AbstractDateTimePattern extends StringValuePattern {
   private DateTimeOffset expectedOffset;
   private DateTimeTruncation truncateExpected;
   private DateTimeTruncation truncateActual;
+  private boolean applyTruncationLast = false;
 
   protected AbstractDateTimePattern(String dateTimeSpec) {
-    this(dateTimeSpec, null, (DateTimeTruncation) null, null, null, null);
+    this(dateTimeSpec, null, (DateTimeTruncation) null, null, false, null, null);
   }
 
   protected AbstractDateTimePattern(
@@ -62,6 +63,7 @@ public abstract class AbstractDateTimePattern extends StringValuePattern {
       String actualDateFormat,
       String truncateExpected,
       String truncateActual,
+      boolean applyTruncationLast,
       Integer expectedOffsetAmount,
       DateTimeUnit expectedOffsetUnit) {
     this(
@@ -69,6 +71,7 @@ public abstract class AbstractDateTimePattern extends StringValuePattern {
         actualDateFormat,
         truncateExpected != null ? DateTimeTruncation.fromString(truncateExpected) : null,
         truncateActual != null ? DateTimeTruncation.fromString(truncateActual) : null,
+        applyTruncationLast,
         expectedOffsetAmount,
         expectedOffsetUnit);
   }
@@ -78,6 +81,7 @@ public abstract class AbstractDateTimePattern extends StringValuePattern {
       String actualDateFormat,
       DateTimeTruncation truncateExpected,
       DateTimeTruncation truncateActual,
+      boolean applyTruncationLast,
       Integer expectedOffsetAmount,
       DateTimeUnit expectedOffsetUnit) {
     super(dateTimeSpec);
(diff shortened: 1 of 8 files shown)


tomakehurst:
This would be a breaking change so I don't think we can merge it. Also, I considered this at the time and I don't think either way round is any more intuitive than the other.

papiomytoglou:
Hello @tomakehurst. I understand it's a breaking change so not merging it is absolutely understandable. But I disagree on the issue of intuitiveness. `now +1 months` is more easily seen as a single unit, I believe. The breaking of `now +1 months` to `now` and `+1 months` with the truncation being applied in between the two is not something that I would consider obvious.

Other than that, I agree. Breaking existing matchers is an issue.

tomakehurst:
I agree it's be useful to be able to switch the order. Perhaps an additional, optional boolean parameter to change the behaviour would be the right approach?

papiomytoglou:
Hello again @tomakehurst. I have pushed the latest iteration taking under account your suggestion for an optional boolean parameter (I named it `applyTruncationLast` for the time being). I have also converted this PR into a draft until we agree on the best approach.

The PR includes a test (`testApplyTruncationLast`) which demonstrates a scenario in which this parameter would be needed.

Please pay attention to the fact that I added a private `Clock` field to `AbstractDateTimePattern` along with the corresponding setter just for testing purposes. I tried using a Mockito spy first but wasn't successful. Will look further into whether the dependency injection approach I took is the only/best one once I have more time.

tomakehurst:
This looks good I think.

Please can we also have a test cases ensuring that the new attribute will be serialised and deserialised correctly?

papiomytoglou:
Hello @tomakehurst,

You were correct, a getter needed for the serialization process was missing, I added `getApplyTruncationLast` which returns null if `applyTruncationLast` is false so that it is excluded from serialization when it has the default value, similar to the logic of the `stringOrNull` method.

I have also added the `deserialisesFromJsonWithApplyTruncationLast`​ test which tests both the deserialization and the feature logic in order to avoid code duplication in the unit tests.

Finally, I removed the `clock` field which I had originally introduced solely for testing purposes and replaced its functionality with using Mockito's `MockStatic` with the `CALLS_REAL_METHODS` argument for partial mocking of static methods.

The Mockito documentation has one [note](https://javadoc.io/static/org.mockito/mockito-core/5.12.0/org/mockito/Mockito.html#CALLS_REAL_METHODS) advising against partial mocking:
> As usual, you are going to read the partial mock warning: Object oriented programming is more-or-less tackling complexity by dividing the complexity into separate, specific, SRPy objects. How does partial mock fit into this paradigm? Well, it just doesn't... Partial mock usually means that the complexity has been moved to a different method on the same object. In most cases, this is not the way you want to design your application.

And another [note](https://javadoc.io/static/org.mockito/mockito-core/5.12.0/org/mockito/Mockito.html#mockStatic(java.lang.Class)) advising against mocking static methods of classes in the standard library:
>We recommend against mocking static methods of classes in the standard library or classes used by custom class loaders used to execute the block with the mocked class. A mock maker might forbid mocking static methods of know classes that are known to cause problems. Also, if a static method is a JVM-intrinsic, it cannot typically be mocked even if not explicitly forbidden.

Still, I think this approach might be better than introducing a field solely for the needs of unit testing. Please let me know what you think.

tomakehurst:
I think this approach to controlling the clock is fine in this case since it's so narrowly scoped. If we were doing lots of this kind of thing I'd be tempted to refactor the code in a way that supports clock injection as standard (rather than just as a testing hook) but for the sake of one test case it's not worth it.

tomakehurst:
One more request - please could you raise a docs PR for this also?

papiomytoglou:
Sorry for the delay, @tomakehurst, here is the documentation PR: https://github.com/wiremock/wiremock.org/pull/318

PR #2803, opened 2024-07-16
update MatchesJsonSchemaPattern to handle invalid json schemas.

RafeArnold:
and include an explanatory sub event.

previously invalid json schemas were rejected on creation, but this is not guaranteed to be correct as external schema refs can change and make previously invalid schemas valid and vice versa.

## Submitter checklist


tomakehurst:
Have we considered the option of eagerly validating the schema (so that it fails when constructing the matcher, rather than at match time)?

PR #2812, opened 2024-07-18
Pick random helper enhancements

tomakehurst:
* Return objects instead of `.toString()` to support further processing.
* Allow an optional `count` to be specified, returning a collection of items with no duplicates.

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/PickRandomHelper.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/PickRandomHelper.java
@@ -30,33 +30,24 @@ public class PickRandomHelper extends HandlebarsHelper<Object> {
           "Must specify either a single list argument or a set of single value arguments.");
     }
 
-    List<Object> candidateValueList = new ArrayList<>();
+    List<Object> valueList = new ArrayList<>();
     if (Iterable.class.isAssignableFrom(context.getClass())) {
-      ((Iterable<Object>) context).forEach(candidateValueList::add);
+      ((Iterable<Object>) context).forEach(valueList::add);
     } else {
-      candidateValueList.add(context);
-      candidateValueList.addAll(Arrays.asList(options.params));
+      valueList.add(context);
+      valueList.addAll(Arrays.asList(options.params));
     }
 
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/PickRandomHelper.java
```diff
@@ -40,20 +40,8 @@ public Object apply(Object context, Options options) throws IOException {
...
-        while (index == -1 || usedIndexes.contains(index)) {
-          index = ThreadLocalRandom.current().nextInt(candidateValueList.size());
-        }
-
-        outputValueList.add(candidateValueList.get(index));
-        usedIndexes.add(index);
-      }
-
-      return outputValueList;
+      Collections.shuffle(candidateValueList);
```
tomakehurst, line 43: Didn't do it this way due to performance. Shuffling a list of 100s just to pick 10 at random seems inefficient.

RafeArnold: good point. ive pushed another commit that only shuffles the list up to the desired length. also added a check to make sure the helper is returning a list of the correct length, because i realised we werent actually doing that. will revert again if you still dont think this is more efficient.

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/PickRandomHelper.java
```diff
@@ -30,33 +30,24 @@ public Object apply(Object context, Options options) throws IOException {
...
-        int index = -1;
-        while (index == -1 || usedIndexes.contains(index)) {
-          index = ThreadLocalRandom.current().nextInt(candidateValueList.size());
-        }
-
-        outputValueList.add(candidateValueList.get(index));
-        usedIndexes.add(index);
+      int desiredLength = Math.min(valueList.size(), count);
+      for (int i = 0; i < desiredLength; i++) {
+        Collections.swap(valueList, i, ThreadLocalRandom.current().nextInt(i, valueList.size()));
```
tomakehurst, line 45: Nice, wouldn't have considered doing it this way

PR #2813, opened 2024-07-20
Add fail fast option to EqualToJson

leeturner:
In a recent release of JsonUnit, a `FAIL_FAST` option was added to improve performance.  This allows the diff comparison to exit out when the first difference is found instead of processing all the differences.  This PR adds a new `failFast` property to the `EqualToJsonPattern` to allow us to turn this property on or off.  To maintain backwards compatibility this is turned off (`false`) by default. 

## References

- https://github.com/lukas-krecan/JsonUnit/pull/785

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
@@ -224,12 +224,17 @@ public class WireMock {
   }
 
   public static StringValuePattern equalToJson(String value) {
-    return new EqualToJsonPattern(value, null, null);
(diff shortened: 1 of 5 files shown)


tomakehurst:
I'm wondering if it would make more sense to always use this in `isExactMatch()` but never in `getDistance()`, rather than making it configurable. That way it's fast when matching and accurate when producing a diff report.

This is what we're already doing in `equalToXml` BTW.

leeturner:
Interesting.  I'll look at implementing that change.  I might create a new PR for that instead of reverting all the configuration changes from this PR.

PR #2819, opened 2024-07-31
Allow Admin implementations to clean up resources

Mahoney:
By making `Admin` extend `Autocloseable` we are able to allow
`HttpAdminClient` to call close on its `CloseableHttpClient`, releasing
the resources.

In the process we can also allow `WireMockServer` & `WireMockApp` to
work with try-with-resources.

## References

- TODO

## Submitter checklist


tomakehurst:
Would be good to see a test case for try-with-resources + `WireMockServer`, although I suspect this might be a bit clunky due to the fact that the server doesn't get started in the constructor. Maybe adding a `WireMockServer.createStarted()` type method would be a good idea?

PR #2823, opened 2024-08-07
Enable customization of cipher suites and TLS protocol with WiremockConfiguration

laffonso:
Adds the ability to customize cipher suites and tls protocols in `WireMockServer` using `WiremockConfiguration`.
With this PR it will be possible to simulate cases where external services deprecates certain tls protocols or cipher suites to guarantee the tested application will continue to work with the remaining supported tls protocol or cipher suites.

When instantiating an WireMockServer will be possible to configure the cipher suites and tls protocol like below:

```java
private static final WireMockServer wireMockServer = new WireMockServer(options()
        .needClientAuth(true)
        .httpDisabled(true)
        .httpsPort(6666)
        .tlsProtocols(LIST_OF_TLS_PROTOCOLS)
        .cipherSuites(LIST_OF_CIPHERS));
```

## References
https://wiremock-community.slack.com/archives/C03N1E6HFPY/p1720564292990239
https://groups.google.com/g/wiremock-user/c/YbgTErqt5ts

- TODO

https://github.com/wiremock/wiremock/issues/2808

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/HttpsSettings.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/HttpsSettings.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2013-2023 Thomas Akehurst
+ * Copyright (C) 2013-2024 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


leeturner:
Hi, many thanks for your contribution.  There are a few things that I was wondering if you could take a look at. 

You have added the new options to the `WireMockConfiguration.java` class but not the Options interface.  This means the new options you have added cannot be configured from the command line for users running WireMock standalone.

I see you have updated the `SslContexts.java` class in the jetty11 package.  We also have support for jetty12 in WireMock so we will need to make sure that your changes are supported across both versions.

I was also wondering if there are any tests you could write to validate the changes you have made?

Issue #2359, opened 2023-09-06
Webhook extension - allow to access response object for webhook request body


AlexShuraits:
### Proposal

Currently webhook extension allows to access only the original request object, but usually, we want to use in webhook body the data generated in the response body.

Example:

```
{
  "request": {
    "urlPathTemplate": "/charge",
    "method": "POST",
    "bodyPatterns": [
      {
        "contains": "source=ba_"
      }
    ]
  },
  "response": {
    "status": 200,
    "bodyFileName": "body.hbs",
    "headers": {
      "Server": "nginx"
    }
  },
  "serveEventListeners": [
    {
      "name": "recordState",
      "parameters": {
        "context": "{{jsonPath response.body '$.id'}}",
        "state": {
          "body": "{{{jsonPath response.body '$'}}}"
        }
      }
    }
  ],
  "postServeActions": [
    {
      "name": "webhook",
      "parameters": {
        "method": "POST",
        "url": "{{systemValue type='ENVIRONMENT' key='WIREMOCK_API_URL'}}/api/webhook",
        "body": "{{{originalResponse.body}}}"
      }
    }
  ]
}
```

oleg-nenashev:
Indeed, it would be a good enhancement

MFrat:
Any schedule?

Diegovictorbr:
Missing this feature :(

adelaune-unyc:
Hello, is there some progress? I would be very interested too

pylavaigun:
It would be really great to implement this, I believe in you!

warrenc5:
I believe this is re-solved in https://github.com/wiremock/wiremock/pull/2128

See standalone test case for wiremock 4.0.0-beta-14 + PR at https://github.com/warrenc5/wiremock-test

PR #2838, opened 2024-08-29
Allow access to response body from webhook request

magzot:
### Access response body in webhook requests.

With this PR, you can now reference the **response.body** inside your callback.

```
{
  "request": {
    "method": "POST",
    "url": "/add-job"
  },
  "response": {
    "status": 202,
    "body": "{\"jobId\":\"{{randomValue length=10 type='ALPHANUMERIC'}}\"}",
    "headers": {
      "Content-Type": "application/json"
    }
  },
  "serveEventListeners": [
    {
      "name": "webhook",
      "parameters": {
        "method": "POST",
        "url": "http://hostname:8080/add-job-callback",
        "headers": {
          "Content-Type": "application/json"
        },
        "body": "{ \"message\": \"success\", \"jobId\": \"{{jsonPath response.body '$.jobId'}}\" }"
      }
    }
  ]
}

```

If you think this implementation is on the right track, I'll proceed to finalise it with tests and documentation.

Cheers,
Magnus

## References

I think this is related to issue #2359.

- TODO

## Submitter checklist


leeturner:
Hi, many thanks for your contribution.  There are a couple of things that I was wondering if you could take a look at.  When we use the original request in the webhook we specifically map this to the `originalRequest` key.  To maintain consistency we should map the response template model to `originalResponse`.

Also, there are no tests in your PR.  We don't merge PRs without an acceptable level of test coverage of your change.   If you could add tests that would be great.

warrenc5:
I see the problem here - the patch does not apply cleanly and rebase is annoying - I will work on this to resolve it.

PR #2853, opened 2024-09-25
chore: changing group of shadow to GradleUp

tomasbjerre:
It was moved here:
https://github.com/GradleUp/shadow


leeturner (approved):
LGTM

PR #2866, opened 2024-10-10
Update external dependency handlebars.java to 4.4.0

gruenich:


leeturner:
Hi, many thanks for the PR but we can't merge this.  The baseline for the `4.4.x` stream is java 17 and we are still supporting java 11 as our baseline.  This is definitely on our radar and we will be updating all dependencies when we move our baseline to java 17

gruenich:
Thanks for the feedback. I missed the required Java version.  
Looking forward to get a version with more recent version of Java and other dependencies!

PR #2897, opened 2024-11-11
Bugfix: Fix list reference behavior in RequestPatternBuilder.like()

narcoticfresh:
## References

I recently was very confused when I did this:

```java
RequestPattern requestPatternA =
    RequestPatternBuilder.newRequestPattern()
        .withRequestBody(WireMock.equalTo("foo"))
        .build();

RequestPattern requestPatternB =
  RequestPatternBuilder.like(requestPatternA)
    .withRequestBody(WireMock.equalTo("bar"))
    .build();
```

What happens is, is that now **both** `RequestPattern` instances contain the **same** `ContentPattern` instances. 

image

Now, I think this is counter intuitive - I would expect that any changes `requestPatternB` are not reflecting back to `requestPatternA` after using `like()`. Would you share that assumption?

This pull request changes the current behavior by making sure that new lists are created in `like()`, thus not referencing the lists in `requestPatternA` anymore.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilder.java
@@ -100,28 +100,28 @@ public class RequestPatternBuilder {
     builder.url = requestPattern.getUrlMatcher();
     builder.method = requestPattern.getMethod();
     if (requestPattern.getHeaders() != null) {
-      builder.headers = requestPattern.getHeaders();
+      builder.headers = new LinkedHashMap<>(requestPattern.getHeaders());
     }
     if (requestPattern.getPathParameters() != null) {
-      builder.pathParams = requestPattern.getPathParameters();
+      builder.pathParams = new LinkedHashMap<>(requestPattern.getPathParameters());
     }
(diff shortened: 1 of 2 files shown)


leeturner:
Many thanks for this PR.  At the moment this is a breaking change for all the people who might be relying on the existing functionality.  Is there any way to do this without it being a breaking change?

narcoticfresh:
@leeturner 
i fully understand the point - although it's a "fix", it changes behavior and breaks bc. 

i think one may decide (i'm not sure how those decisions are formed within the wiremock project) which one should be the "correct" behavior of `RequestPatternBuilder.like()`.

if it shall be changed (and thus only then it will be behaving like widely used form of the `Builder` pattern and immutability), then i guess it would be best to document it and just change it with the next major release.

if that is not desired, another approach could be to just introduce for example `RequestPatternBuilder.copyOf(RequestPatternBuilder)` which then does this "real" property transfer.

let me know if i should add this to the pr..

tomakehurst:
Could we make the name explicitly indicate that the body patterns would be merged rather than replaced? Or add an overload of `like()` taking a parameter to indicate this?

PR #2919, opened 2024-12-20
Fix serialization of nested AdvancedPathPattern where @JsonUnwrapped is used

leeturner:
Ths PR fixes https://github.com/wiremock/wiremock/issues/2910 where nested AdvancedPathPattern matches are not serialized correctly due to the use of the `@JsonUnwrapped` annotation.  When a custom serializer is used, as is the case with `MatchesXPathPattern` and `MatchesJsonPathPattern`, you effectively need to make sure your serializers handle the unwrapping

## References

- https://michael-simons.github.io/simple-meetup/unwrapping-custom-jackson-serializer
- https://github.com/FasterXML/jackson-databind/issues/3007

## Submitter checklist


src/test/java/com/github/tomakehurst/wiremock/AdvancedPathPatternSerializationTest.java
```diff
@@ -0,0 +1,188 @@
...
+    String expectedJson =
+        "{\n"
+            + "        \"matchesXPath\" : {\n"
+            + "            \"expression\" : \"//AccountId\",\n"
+            + "            \"equalTo\" : \"123\"\n"
+            + "        }\n"
+            + "}";
+    StringValuePattern pattern = matchingXPath("//AccountId", equalTo("123"));
+    String serializedPattern = Json.write(pattern);
+    JSONAssert.assertEquals(expectedJson, serializedPattern, true);
```
tomakehurst, line 50: Can I suggest using JSONUnit for these assertions rather than JSONAssert? I think this is the best of the JSON matching libraries we have in play so we should be moving towards using it for everything.

tomakehurst (changes requested):
Aside from the comment about JSON assertions, looks good 👍

src/test/java/com/github/tomakehurst/wiremock/AdvancedPathPatternSerializationTest.java
leeturner: Awesome.  I have updated the asserts to use jsonUnit - `assertThat(Json.write(pattern), jsonEquals(expectedJson));`

tomakehurst (approved):

Issue #2936, opened 2025-01-15
EqualToXmlPattern ignores namespaces


michaelomichael:
### Proposal

There are a number of cases where XML documents with different namespaces are, wrongly, considered to be exactly equal by `EqualToXmlPattern`.

### Example 1

Document 1a:
```xml
<?xml version="1.0" encoding="utf-8"?>
<a:Root xmlns:a="http://whatever/">
    <a:Child/>
</a:Root>
```

Document 1b:
```xml
<?xml version="1.0" encoding="utf-8"?>
<a:Root xmlns:a="http://whatever/">
    <Child/>
</a:Root>
```

These two documents are not identical because the `Child` element with namespace prefix is completely different, semantically, from the one with no prefix.

However, both of the following return `true` today, when they should really return `false`:
* `equalToXml(document1a).match(document1b).isExactMatch()`
* `equalToXml(document1b).match(document1a).isExactMatch()` 

### Example 2

Document 2a:
```xml
<?xml version="1.0" encoding="utf-8"?>
<a:Root xmlns:a="http://whatever/">
    <Child/>
</a:Root>
```

Document 2b:
```xml
<?xml version="1.0" encoding="utf-8"?>
<a:Root xmlns:a="http://whatever/">
    <Child xmlns="http://somethingnew/"/>
</a:Root>
```

We get mixed results this time:
* `equalToXml(document2a).match(document2b).isExactMatch()` -> correctly returns `false` 😄 
* `equalToXml(document2b).match(document2a).isExactMatch()` -> should return `false` but returns `true` 😦 

### Invalid Test

The [`namespaceComparisonCanBeExcluded` test](https://github.com/wiremock/wiremock/blob/master/src/test/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPatternTest.java#L440) is passing today, but the XML documents are completely different due to their differing namespace URIs so, really, it should fail. 

If the `exemptingComparisons(NAMESPACE_URI)` method were to be incorporated (as is the case in the subsequent `namespaceComparisonCanBeExcluded2` test) then I would expect the test to pass.

### Notes
This was tested in 3.10.0, but I think the issue been around for a while.

### Reproduction steps

The tests added to [this branch](https://github.com/wiremock/wiremock/commit/b1a87da4ffafb06a3cbeeb49913b9283e1a65c18) demonstrate the problem:
* `doesNotReturnExactMatchWhenActualHasMissingNamespaceUri`: passes ✅ 
[...]

PR #2937, opened 2025-01-15
Use namespace-aware XML parser in EqualToXmlPattern

michaelomichael:
The `DocumentBuilderFactory` used by `EqualToXmlPattern` does not have its "namespace-aware" feature enabled, so certain combinations of missing or added namespaces in the XML documents being compared are being incorrectly ignored. (This is described in detail in issue #2936.)

The changes in this PR include creation of a subclass of `Xml.SkipResolvingEntitiesDocumentBuilderFactory` to maintain a separate namespace-aware `DocumentBuilderFactory`, following the same `ThreadLocal`-based caching as the original.

A contributing factor is the logic in `IgnoreUncountedDifferenceEvaluator.evaluate()`, which ignores any comparisons where the "control" value is `null`:
```java
      if (finalCountedComparisons.contains(comparison.getType())
          && comparison.getControlDetails().getValue() != null) {
        return outcome;
      }

      return ComparisonResult.EQUAL;
```

I've removed the second line, so the tests now pass, but I imagine that the logic was put in for a reason, so it would be good to hear about any edge cases from the past, for which we should add new test cases.

Overall, this change will make the `isEqualToXml` pattern more strict, and I'm aware that different people will have different requirements regarding just how strict the equality should be, so if there's a different approach you think we should take here just let me know. I will say that this bug caught my team by surprise: we expected that our tests would fail if the XML our app was writing was missing some namespaces, but that wasn't the case, and the bug in the app was only discovered later during integration testing.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2016-2024 Thomas Akehurst
+ * Copyright (C) 2016-2025 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


leeturner:
Hi, many thanks for the PR.  Given you say this PR will make `isEqualToXml` more strict, am I correct in saying this could break users tests that are currently working using the current version of WireMock?  If so, this would be a breaking change and we would need to be careful about how we release this. 

We would idealy keep any breaking changes to WireMock version 4 where breaking changes are going to be expected.  Is there any way we could release this while keeping the existing behaviour as the default but allowing your change to be opt in?

michaelomichael:
Hi @leeturner, Yes, that's a good point about breaking changes. 

I'm going to cancel this PR, as the basic namespace-aware aspect has just been implemented separately by @RafeArnold in #2945. Thanks @RafeArnold! 😄

PR #2939, opened 2025-01-17
feat: add `originalResponse` to be available in the templating

Mohsin-Ul-Islam:
## References

- Implements functionality for #2359

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
@@ -1,0 +1,39 @@
+/*
+ * Copyright (C) 2016-2025 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.extension.responsetemplating;
+
+import com.github.tomakehurst.wiremock.common.ListOrSingle;
+import com.github.tomakehurst.wiremock.http.Body;
+import java.util.Map;
+
+public class ResponseTemplateModel {
+
+  private final Body body;
+  private final Map<String, ListOrSingle<String>> headers;
+
+  public ResponseTemplateModel(Map<String, ListOrSingle<String>> headers, Body body) {
+    this.body = body;
+    this.headers = headers;
+  }
+
+  public String getBody() {
+    return body.asString();
+  }
+
+  public Map<String, ListOrSingle<String>> getHeaders() {
+    return headers;
+  }
+}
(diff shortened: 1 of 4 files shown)


src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
```diff
@@ -0,0 +1,39 @@
...
+  private final Body body;
+  private final Map<String, ListOrSingle<String>> headers;
+
+  public ResponseTemplateModel(Map<String, ListOrSingle<String>> headers, Body body) {
+    this.body = body;
+    this.headers = headers;
+  }
+
+  public String getBody() {
+    return body.asString();
```
leeturner, line 33: Hi, many thanks for your PR.  I think we will need to take a look at this and how it is going to work when the `Body` isn't a string.   The body can be returned as base64 which we will need to factor in here I think.  At least allow the other body representations to be returned from the ResponseTemplateModel

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
```diff
@@ -0,0 +1,39 @@
...
+import com.github.tomakehurst.wiremock.common.ListOrSingle;
+import com.github.tomakehurst.wiremock.http.Body;
+import java.util.Map;
+
+public class ResponseTemplateModel {
+
+  private final Body body;
+  private final Map<String, ListOrSingle<String>> headers;
+
+  public ResponseTemplateModel(Map<String, ListOrSingle<String>> headers, Body body) {
```
leeturner, line 27: I was just wondering if there is any value to adding the response status here for a more complete response model?

leeturner:
Many thanks for the PR.  I have added a few comments for you to take a look at.  Also, could we add a new test to validate the headers are transferred to the model OK and maybe what happens when there are no headers set in the response etc

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
Mohsin-Ul-Islam: Yes, I think it is better to have response status in the model as well.

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
Mohsin-Ul-Islam: makes sense

Mohsin-Ul-Islam:
> Many thanks for the PR. I have added a few comments for you to take a look at. Also, could we add a new test to validate the headers are transferred to the model OK and maybe what happens when there are no headers set in the response etc

@leeturner I have updated the test to assert for headers, status & body. Please re-review.

Thanks

Mohsin-Ul-Islam:
> Also, if you have time, it would be great if you could add a PR to the docs site documenting the new addition to the webhook page - https://github.com/wiremock/wiremock.org/blob/main/_docs/webhooks-and-callbacks.md

@leeturner here is the PR containing documentation about this feature: https://github.com/wiremock/wiremock.org/pull/334

src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/ResponseTemplateModel.java
```diff
@@ -0,0 +1,45 @@
...
+    this.status = status;
+    this.headers = headers;
+  }
+
+  public int getStatus() {
+    return status;
+  }
+
+  public String getBody() {
+    return body.asBase64();
```
tomakehurst, line 39: I don't think this is what we want to expose to the templates.

The best approach would be to return the string representation when it's a textual body and null otherwise IMO.

bernalvarela:
Hello, this feature looks really interesting.

It seems that this PR completes the work from the previously merged one: https://github.com/wiremock/wiremock/issues/2359.

Is there anything that can be done to unblock this PR and get it merged?

Thank you all for your work and best regards!!

Mohsin-Ul-Islam:
@leeturner @tomakehurst can we take this forward. Thanks for your support.

leeturner:
We have now pushed the 4.x beta branch to main so would it be possible to rebase on the latest ?

Issue #2931, opened 2025-01-08
Allow passing a custom HTTP client into the HttpAdminClient constructor


ascopes:
### Proposal

We have a use case where we need to be able to access WireMock over a port-forwarded connection, and there is unfortunately no room to be able to change requirements around networking.

To ensure that the system hosting the WireMock server can be hit by the client, we have to ensure that SNI is handled correctly due to constraints around the hosting platform. The use of port forwarding will mean that SNI is incorrectly handled due to the fact hitting `https://localhost:5000` would be forwarded to the target network.

The workaround for this is that we need to be able to customize the Apache HttpComponents HTTP client with some custom configuration to support this setup. 

Right now, HttpAdminClient does not provide a public constructor that allows passing a pre-built CloseableHttpClient into the class. Our use case would be able to be supported via the addition of a new constructor in that class that can take a `CloseableHttpClient` as a parameter.

The alternative right now would be to implement the `Admin` type entirely from scratch, since the [safelyExecuteRequest](https://github.com/wiremock/wiremock/blob/9f394706c59e84fee2d237d5f5f3055f288bffd7/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java#L543) method is marked as private, so we cannot just subclass `HttpAdminClient` and override this method.

Would it be possible to implement this functionality to unblock this use case?

dnsmkl2:
Hi,

I'm also very much interested in this enhancement.

Question to Wiremock maintainers:
- Does Wiremock have any specific needs from the `CloseableHttpClient`? (didn't see anything wiremock specific from glancing at `HttAdminClient/HttpClientFactory` implementation)
- Would you accept PR, which would allow to pass custom `CloseableHttpClient`?

Arguments for it:

- quite a lot of projects that use wiremock already are constructing `CloseableHttpClient`. (Apache `HttpClient` is still used a lot). So  just reusing same `CloseableHttpClient` would make things simple.
- it would solve https://github.com/wiremock/wiremock/issues/2774 "Support mTLS (client certificate) for standalone WireMock client"

Argument against it could be:

- it would tie implementation to Apache httpClient. (But that's relevant only if there is a plan/wish to change to different flavor of http client)

viktorijabb:
Hi, I can pick this up and implement it asap as it is of interest to me as well. Happy for it to be assigned to me. 

[...]

PR #2941, opened 2025-01-21
GH-2931 Allow passing a custom HTTP client into the HttpAdminClient constructor 

viktorijabb:
## References

Fixes GH-2931

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
@@ -92,6 +92,17 @@ public class HttpAdminClient implements Admin {
     this(scheme, host, port, urlPathPrefix, hostHeader, null, 0, noClientAuthenticator());
   }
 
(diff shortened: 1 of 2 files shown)


leeturner:
Many thanks for this PR.  In the issue we [talked about](https://github.com/wiremock/wiremock/issues/2931#issuecomment-2590330560) using the WireMock `HttpClient` interface as the new parameter to the constructor instead of the `CloseableHttpClient` to help stop the Apache classes leak into our own.  Is that something you would be up for implementing.  It will require a little rework of the `HttpAdminClient` but it would be a nice addition to this PR if you are up for it?

viktorijabb:
Yep, Im happy to pick that up!

tomakehurst:
This is the right way to start, but it doesn't compile at the moment as the code that calls the HTTP client also needs to be updated.

I suggest running `./gradlew spotlessApply check` before pushing new commits.

viktorijabb:
Hi @leeturner @tomakehurst , this needs some attention, when possible.

PR #2944, opened 2025-01-24
improve performance of EqualToXmlPattern.

RafeArnold:
removes use of DiffBuilder.ignoreComments which performs an xslt transform on the provided XML documents which is very memory intensive. now comments are ignored at read/parse time.

as part of this change, the Xml utility class was also refactored to (hopefully) make it more reusable.

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPatternTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPatternTest.java
@@ -568,4 +568,12 @@ public class EqualToXmlPatternTest {
     MatchResult result = pattern.match("<body><entry/><entry/></body>");
     assertTrue(result.isExactMatch());
   }
+
+  @Test
+  void matchesIfCommentsDiffer() {
+    EqualToXmlPattern pattern =
+        new EqualToXmlPattern("<body><!-- Comment --><entry/><entry/></body>", false, true);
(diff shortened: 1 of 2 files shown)


src/test/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPatternTest.java
```diff
@@ -450,7 +450,8 @@ public void namespaceComparisonCanBeExcluded() {
             + "    <st:thing>Match this</st:thing>\n"
             + "</stuff>";
 
-    MatchResult matchResult = equalToXml(expected).match(actual);
+    MatchResult matchResult =
```
tomakehurst, line 453: I think fixing this test and enabling the namespace flag in order to make it pass again is the right thing to do, but it should be in a separate PR. I suggest you create one for that first, we'll get it merged then you can slim this one down so it just fixes the comments performance issue.

tomakehurst: Just adding the namespace flag causes quite a few tests to fail so I think we need to examine these carefully to see if we need to preserve old behaviour in at least some cases.

RafeArnold: ive put the Xml refactor and namespace aware change into a separate pr here https://github.com/wiremock/wiremock/pull/2945

tomakehurst (approved):

PR #2953, opened 2025-02-03
Fix OpenAPI 3.0 schemas

ascopes:
Spotted a few more issues with the OpenAPI docs over the weekend, so have put a PR together to try and address them in bulk. Hope it is all okay.

- Fix `equal-to-json` to document that using a raw object is also valid, and add example.
- Rewrite schema for `matches-json-schema` as it was just copied from the `matches-xpath` schema.
- Removed undocumented `xPathNamespaces` attributes from `matches-json-schema`.
- Update description of `not` to match `or` and `and`.
- Fix example in `request-pattern` that was not valid JSON.
- Fix headers example in `logged-request`.
- Correctly trim whitespace out of `equal-to-pattern` example.
- Remove spurious whitespace from admin API descriptors.
- Convert raw JSON examples to raw YAML objects to make the example format clear in any generated Swagger/OA3 HTML pages.

### DIFF
--- a/src/main/resources/swagger/schemas/logged-request.yaml
+++ b/src/main/resources/swagger/schemas/logged-request.yaml
@@ -15,19 +15,19 @@ properties:
   headers:
     description: 'Header patterns to match against in the <key>: { "<predicate>": "<value>" } form'
     type: object
-    example: |-
-      {
-        "Connection": { "equalTo": "keep-alive" },
-        "Host": { "equalTo": "localhost:56738" },
-        "User-Agent": { "equalTo": "Apache-HttpClient/4.5.1 (Java/1.7.0_51)" }
-      }
+    example:
+      Connection:
+        equalTo: keep-alive
+      Host:
+        equalTo: "localhost:56738"
+      User-Agent:
+        equalTo: "Apache-HttpClient/4.5.1 (Java/1.7.0_51)"
   cookies:
     description: 'Cookie patterns to match against in the <key>: { "<predicate>": "<value>" } form'
     type: object
-    example: |-
-      {
-        "user_id": { "equalTo": "tom1234" }
-      }
+    example:
+      user_id:
+        equalTo: tom1234
   body:
     description: Body string to match against
     type: string
--- a/src/main/resources/swagger/schemas/logged-request.yaml
+++ b/src/main/resources/swagger/schemas/logged-request.yaml
@@ -7,22 +7,27 @@ properties:
   url:
     description: The path and query to match exactly against
     type: string
-    example: "/received-request/2"
+    example: "/received-request/2?foo=bar"
   absoluteUrl:
     description: The full URL to match against
     type: string
-    example: "http://localhost:56738/received-request/2"
+    example: "http://localhost:56738/received-request/2?foo=bar"
   headers:
     description: 'Header patterns to match against in the <key>: { "<predicate>": "<value>" } form'
     type: object
-    example:
-      Connection: keep-alive
-      Host: localhost:56738
-      User-Agent: Apache-HttpClient/4.5.1 (Java/1.7.0_51)
+    example: |-
+      {
+        "Connection": { "equalTo": "keep-alive" },
+        "Host": { "equalTo": "localhost:56738" },
+        "User-Agent": { "equalTo": "Apache-HttpClient/4.5.1 (Java/1.7.0_51)" }
+      }
   cookies:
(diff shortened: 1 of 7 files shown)


src/main/resources/swagger/schemas/matches-json-schema-pattern.yaml
```diff
@@ -3,23 +3,39 @@ type: object
...
+              "name"
+            ],
+            "properties": {
+            "name": {
+              "type": "string"
+            },
+            "tag": {
+              "type": "string"
+            }
+          }
```
RafeArnold, line 33:
```suggestion
            "properties": {
              "name": {
                "type": "string"
              },
              "tag": {
                "type": "string"
              }
            }
          }
```

src/main/resources/swagger/schemas/logged-request.yaml
```diff
@@ -7,22 +7,27 @@ properties:
   url:
     description: The path and query to match exactly against
     type: string
-    example: "/received-request/2"
+    example: "/received-request/2?foo=bar"
   absoluteUrl:
     description: The full URL to match against
     type: string
-    example: "http://localhost:56738/received-request/2"
+    example: "http://localhost:56738/received-request/2?foo=bar"
```
tomakehurst, line 14: If we've got a query parameter here and on line 10 then it also needs to be under the `queryParams` map.

src/main/resources/swagger/schemas/logged-request.yaml
```diff
@@ -7,22 +7,27 @@ properties:
...
   headers:
     description: 'Header patterns to match against in the <key>: { "<predicate>": "<value>" } form'
     type: object
     example:
-      Connection: keep-alive
-      Host: localhost:56738
-      User-Agent: Apache-HttpClient/4.5.1 (Java/1.7.0_51)
+      Connection:
+        equalTo: keep-alive
```
tomakehurst, line 20: This isn't correct - we're logging the values of the headers not defining matchers. Suggest we just revert this to the previous version.

src/main/resources/swagger/schemas/logged-request.yaml
```diff
@@ -7,22 +7,27 @@ properties:
...
+        equalTo: "localhost:56738"
+      User-Agent:
+        equalTo: "Apache-HttpClient/4.5.1 (Java/1.7.0_51)"
   cookies:
     description: 'Cookie patterns to match against in the <key>: { "<predicate>": "<value>" } form'
     type: object
-    example: {}
+    example:
+      user_id:
+        equalTo: tom1234
```
tomakehurst, line 30: Same as for headers - this should be a simple value, not a matcher.

src/main/resources/swagger/schemas/matches-json-schema-pattern.yaml
```diff
@@ -3,23 +3,40 @@ type: object
 properties:
   matchesJsonSchema:
     oneOf:
```
tomakehurst, line 5: I don't think this `oneOf` is needed at all, could be removed.

src/main/resources/swagger/schemas/matches-json-schema-pattern.yaml
```diff
@@ -3,23 +3,40 @@ type: object
...
+                "type": "string"
+              }
+            }
+          }
 
-  xPathNamespaces:
-    type: object
-    additionalProperties:
-      type: string
+  schemaVersion:
```
tomakehurst, line 36: This could be an enum of the following values:
V4
V6
V7
V201909
V202012

tomakehurst (changes requested):

src/main/resources/swagger/schemas/logged-request.yaml
ascopes: Reverted!

src/main/resources/swagger/schemas/matches-json-schema-pattern.yaml
ascopes: Removed

src/main/resources/swagger/schemas/matches-json-schema-pattern.yaml
ascopes: Good shout, updated with the change

ascopes:
I will check out the failing tests when I am at my PC next. Hope thats okay.

tomakehurst:
Thanks for this contribution, and sorry it's taken so long to merge!

ascopes:
No problem, thanks!

PR #2954, opened 2025-02-04
feat: add namespace parameter to equalToXml.

RafeArnold:
Previously, the behaviour of EqualToXmlPattern in regards to XML namespaces was inconsistent and hard to predict. This PR attempts to define this behaviour and add additional options to be more explicit about how namespaces should be handled.

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPattern.java
@@ -45,7 +45,6 @@ public class EqualToXmlPattern extends StringValuePattern {
           SCHEMA_LOCATION,
           NO_NAMESPACE_SCHEMA_LOCATION,
           NODE_TYPE,
-          NAMESPACE_PREFIX,
           NAMESPACE_URI,
           TEXT_VALUE,
           PROCESSING_INSTRUCTION_TARGET,
@@ -102,9 +101,6 @@ public class EqualToXmlPattern extends StringValuePattern {
     Set<ComparisonType> comparisonsToExempt = new HashSet<>();
     if (exemptedComparisons != null) {
       comparisonsToExempt.addAll(exemptedComparisons);
-    }
-    if (namespaceAwareness == NamespaceAwareness.LEGACY) {
-      comparisonsToExempt.add(NAMESPACE_PREFIX);
     }
     this.countedComparisons =
         COUNTED_COMPARISONS.stream()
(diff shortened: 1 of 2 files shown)


RafeArnold:
docs https://github.com/wiremock/wiremock.org/pull/335

src/main/java/com/github/tomakehurst/wiremock/matching/EqualToXmlPattern.java
```diff
@@ -45,6 +45,7 @@ public class EqualToXmlPattern extends StringValuePattern {
           SCHEMA_LOCATION,
           NO_NAMESPACE_SCHEMA_LOCATION,
           NODE_TYPE,
+          NAMESPACE_PREFIX,
```
tomakehurst, line 48: Do we actually care about whether the prefix is the same?

My understanding is that as long as two prefixes refer to the same URI they're semantically equivalent.

RafeArnold: i think it depends on what the user wants. i figure putting it in will allow the user to do prefix comparison but also allow them to opt out via the exemptedComparisons attributes if they're not interested. that way they can configure their matcher how they please

tomakehurst: This could end up breaking a lot of folks' tests for something that's not a meaningful difference. I suggest we pull this one out for the time being, or at least make it opt-in rather than out.

RafeArnold: done

src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java
```diff
@@ -282,13 +286,30 @@ public static EqualToXmlPattern equalToXml(
       String placeholderOpeningDelimiterRegex,
       String placeholderClosingDelimiterRegex,
       boolean ignoreOrderOfSameNode) {
+    return equalToXml(
+        value,
+        enablePlaceholders,
+        placeholderOpeningDelimiterRegex,
+        placeholderClosingDelimiterRegex,
+        ignoreOrderOfSameNode,
+        null);
```
tomakehurst, line 295: Why null rather than NONE?

tomakehurst: OK, that's a good reason to leave it null I think.

tomakehurst (approved):

PR #2957, opened 2025-02-05
Support optional removal of null-valued attributes in jsonMerge helper

tomakehurst:
Adds an optional `removeNulls` parameter to the `jsonMerge` helper. When true this removes any attributes from the resulting JSON that have null values in the JSON to merge.


RafeArnold (approved):

tomakehurst:
OSS doc PR: https://github.com/wiremock/wiremock.org/pull/339

PR #2961, opened 2025-02-11
Use java system proxy properties for forward proxy mode

mnemeceditel:
This PR enhances Wiremock with a capability to read java system proxy settings (i.e. -Dhttps.proxyHost & -Dhttps.proxyPort) when no --proxy-via option is specified.

## References

- no references

## Submitter checklist


mnemeceditel:
Hi @tomakehurst, thank you for the feedback.

The idea was that JRE proxy settings should be honored in general (i.e. my feeling was that they were simply ignored in this scenario). If one configures those properties on JRE level, her expectation should, of course, be that it applies on all outgoing http traffic from that java application.

The suggestion from you is definitely fine too - i.e. it also solves the described problem. I could imagine a flag like --proxy-via-exclude or something similar. What do you think?

Issue #2913, opened 2024-12-08
FormatXmlHelper constructor throws exception when TransformerFactory does not support attribute


vkuzel:
### Proposal

Version: 3.10.0

In the `FormatXmlHelper` constructor, there are a few calls of the `TransformerFactory.setAttribute()` method.

https://github.com/wiremock/wiremock/blob/7df345b9065b2b7ba75ede12998603a262d8ff6c/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/FormatXmlHelper.java#L68-L70

The method may throw an `IllegalArgumentException` if the attribute is not supported by the underlying implementation of the `TransformerFactory`.

The underlying implementation is selected by classes present on the classpath. The issue may manifest for the `oracle.xml.jaxp.JXSAXTransformerFactory`

### Workaround:

Set the `TransformerFactory` implementation explicitly via System.property. E.g., `System.setProperty("javax.xml.transform.TransformerFactory", "com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl");`

### Suggested solution

Instead of searching for the `TransformerFactory` implementation via `TransformerFactory.newInstance()` use custom algo which will explicitly prefer supported known implementations present on the classpath.

### Reproduction steps

1. Add Oracle XMLparserv2 dependency to the project: `com.oracle.database.xml:xmlparserv2:23.5.0.24.07`
2. Create instance of the `com.github.tomakehurst.wiremock.extension.responsetemplating.helpers.FormatXmlHelper`

### References

1. [`TransformerFactory.setAttribute()` documentation](https://docs.oracle.com/javase/8/docs/api/javax/xml/transform/TransformerFactory.html#setAttribute-java.lang.String-java.lang.Object-).

EstelaGarcia:
We are getting this problem with Saxon as the XML parser too.

PR #2975, opened 2025-02-27
Always use the system default `DocumentBuilderFactory`, `TransformerFactory` and `XPathFactory`

### LINKED ISSUE #2454: Cannot render XML if another TransformerFactoryImpl exists
### Proposal

To support in-process integrations with applications that make use of libraries like Saxon

Since we're using JDK 17, Wm cannot use the internal JDK class, and because a different library brings in Saxon then Wm fails because Saxon does not support the indent attribute.

### Reproduction steps

wiremock v2.35.0
camel-spring-ws v3.20.2 which depends on 
Saxon-HE  v11.4 which doesn't support "indent-number" attribute

### References

Related issues:

#849 
#1072 
#1695

A possible solution from another project
https://github.com/nextgenhealthcare/connect/issues/1361#issuecomment-626880525
Implemented here
https://github.com/nextgenhealthcare/connect/commit/f1a31208bed28c8bd28225b3f0a442438362afd7

vkuzel:
Closes #2454
Related #2913

Followup to recently merged https://github.com/wiremock/wiremock/pull/2918 which solved the issue partly.

The `newInstance()` method of aforementioned classes may return a different implementation based on classes present on classpath. For example when `oracle.xml.jaxp.JXDocumentBuilderFactory` is present on the classpath, the use use of the `http://xml.org/sax/features/validation` attribute in the `Xml.java` file causes `ParserConfigurationException`.

The `newDefaultInstance()` method ensures consistency across different environment by using the same implementation irrespective to the classpath.

## References

- TODO

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
@@ -48,8 +48,8 @@ public class Xml {
 
   public static void optimizeFactoriesLoading() {
     try {
-      String transformerFactoryImpl = TransformerFactory.newInstance().getClass().getName();
-      String xPathFactoryImpl = XPathFactory.newInstance().getClass().getName();
+      String transformerFactoryImpl = TransformerFactory.newDefaultInstance().getClass().getName();
+      String xPathFactoryImpl = XPathFactory.newDefaultInstance().getClass().getName();
 
       System.setProperty(TransformerFactory.class.getName(), transformerFactoryImpl);
       System.setProperty(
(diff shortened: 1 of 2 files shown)


src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java
```diff
@@ -116,7 +116,7 @@ private static TransformerFactory createTransformerFactory() {
                   .getDeclaredConstructor()
                   .newInstance();
```
jpraet, line 117: On JDK17 this throws `java.lang.IllegalAccessException: class com.github.tomakehurst.wiremock.common.xml.Xml cannot access class com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl (in module java.xml) because module java.xml does not export com.sun.org.apache.xalan.internal.xsltc.trax to unnamed module @96532d6`.

Can't it be removed? Similar to how it's done here:

https://github.com/wiremock/wiremock/blob/fbf904da993830e6f16f4fb8d6d8e694d6645dae/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/FormatXmlHelper.java#L68-L69

PR #2990, opened 2025-03-12
allow admin calls when standalone instance is behind haproxy/loadbala…

vggeorgas:
…ncer/service mesh

Existing Wiremock implementation does not allow to do admin calls if standalone instance is behind a load balancer/haproxy/service mesh etc. This is because it forces you to specify port (although in this case it is implied to be 443 if scheme is https for example) and the resulting url is https://domain:port/suffix. It is very often the case the routing in above loadbalancers, relies on the host header which would include the :port part, and therefore it does not work. With this change, we allow users to create a Wiremock instance without specifying the port, letting http protocol take care of that.

## References

- TODO

I did not create an issue but I had the following chat in the slack channel:
https://wiremock-community.slack.com/archives/C03N1E6HFPY/p1741623042096729

## Submitter checklist

- [x ] Recommended: Join [WireMock Slack](https://slack.wiremock.org/) to get any help in `#help-contributing` or a project-specific channel like `#wiremock-java`

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
@@ -125,13 +125,14 @@ public class HttpAdminClient implements Admin {
       String proxyHost,
       int proxyPort,
       ClientAuthenticator authenticator) {
-    this(scheme,
-            host,
-            port,
-            urlPathPrefix,
-            hostHeader,
-            authenticator,
-            HttpClientFactory.createClient(createProxySettings(proxyHost, proxyPort)));
+    this(
+        scheme,
+        host,
+        port,
+        urlPathPrefix,
+        hostHeader,
+        authenticator,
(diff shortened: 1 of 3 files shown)


src/main/java/com/github/tomakehurst/wiremock/client/HttpAdminClient.java
```diff
@@ -132,6 +137,25 @@ public HttpAdminClient(
     httpClient = HttpClientFactory.createClient(createProxySettings(proxyHost, proxyPort));
   }
 
+  public HttpAdminClient(
+      String scheme,
+      String host,
```
RafeArnold, line 142: can the constructor above delegate down to this new one.

```java
    this(
        scheme,
        host,
        port,
        urlPathPrefix,
        hostHeader,
        authenticator,
        HttpClientFactory.createClient(createProxySettings(proxyHost, proxyPort)));
```
should do the trick.

RafeArnold (approved):
lgtm. need to run `spotlessApply` gradle task then we can merge it.

RafeArnold:
Thanks!

PR #3004, opened 2025-03-26
feat: added `captureHost` option to `RecordSpec`

### LINKED ISSUE #1288: Adding the host and protocol to record/playback mode
My team has been experimenting with [record and playback](http://wiremock.org/docs/record-playback/), and we noticed that the `url` field in the produced JSON is populated only with the path that is called but not with the host or protocol. It would be nice as a feature to add this information to the generated JSON. If this feature already exists, is it possible to document it? If not, I could take a look into making a PR to implement it.

quintesse:
Fixes #1288

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/recording/RecordSpec.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/recording/RecordSpec.java
@@ -35,6 +35,9 @@ public class RecordSpec {
   // matcher
   private final Map<String, CaptureHeadersSpec> captureHeaders;
 
+  // Include the host pattern in the stub mapping
+  private final Boolean captureHost;
+
   // Factory for the StringValuePattern that will be used to match request bodies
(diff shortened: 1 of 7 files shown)


quintesse:
I only created this as a draft PR because even though this (probably) fixes the original issue as defined in #1288 I'm not really happy with this solution.

Why? Because what if somebody says "hey, now I want to add the port number too!" and if that, why not the protocol too?  Etc etc. Meaning we we have to keep adding new flags to `RecordSpec` and keep passing all of them down the call chain into `SnapshotStubMappingGenerator` and `RequestPatternTransformer`.

So I'm wondering if it would not be a better option to have an extension point where a user can provide their own implementation of a `SnapshotStubMappingGenerator`. Wdyt?

tomakehurst:
I think a flag for each is the way to do it, but you could change `SnapshotStubMappingGenerator` so that it just takes the whole `RecordSpec` object as a parameter instead of having to unpack the individual attributes.

PR #3014, opened 2025-03-30
Expose native matcher for client ip address

lhcopetti:
This PR adds a new native matcher for client IP address that is available in wiremock's request ([see here](https://github.com/wiremock/wiremock/blob/7257dd665facf7329a656d14987ccfecff33606d/src/main/java/com/github/tomakehurst/wiremock/servlet/WireMockHttpServletRequestAdapter.java#L121) for reference where the actual value is loaded from). This improvement came up during a slack discussion available below. 

Usage looks like this:

```
wm.stubFor(get(anyUrl()).withClientIp(equalTo("127.0.0.1")));
```

As a native matcher, it benefits from a improved diff output in case of errors:

image

This new matcher is basically syntactic sugar for:

```
wm.stubFor(get(anyUrl()).andMatching(r -> equalTo("127.0.0.1").match(r.getClientIp())));
```

And the diff output without the native matcher would look like below (no extra info in case of failure):

image

## References

Slack conversation: https://wiremock-community.slack.com/archives/C04HHJ7F20G/p1741182115686839

## Submitter checklist

### DIFF
--- a/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
+++ b/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2011-2024 Thomas Akehurst
+ * Copyright (C) 2011-2025 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,6 +46,7 @@ public class RequestPattern implements NamedValueMatcher<Request> {
   private final String scheme;
   private final StringValuePattern host;
   private final Integer port;
+  private final StringValuePattern clientIp;
   private final UrlPattern url;
   private final RequestMethod method;
   private final Map<String, MultiValuePattern> headers;
@@ -66,6 +67,7 @@ public class RequestPattern implements NamedValueMatcher<Request> {
       final String scheme,
       final StringValuePattern host,
       final Integer port,
+      final StringValuePattern clientIp,
       final UrlPattern url,
       final RequestMethod method,
       final Map<String, MultiValuePattern> headers,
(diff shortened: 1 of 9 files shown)


src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
```diff
@@ -299,6 +308,10 @@ private MatchResult portMatches(final Request request) {
     return port != null ? MatchResult.of(request.getPort() == port) : MatchResult.exactMatch();
   }
 
+  private MatchResult clientIpMatches(final Request request) {
```
Arc-E-Tect, line 311: I'm a fan of documenting the not-so-obvious. In this case, I would've added Javadoc explaining the result of the call as it is not that straight forward. That said, other methods don't have similar documentation either, but I would've added it.

Arc-E-Tect:
@lhcopetti this what I was looking for.

Looked at the code and it looks straight forward and easy enough to understand. Had one comment, but that's because sometimes I'm a bit of a puritan when it comes to documentation.

RafeArnold (approved):

src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
RafeArnold: agreed. additional documentation never hurts.

lhcopetti: Thanks both of you for the feedback. I'll keep this in mind for my next contributions.

If I were to add a comment though, I think I would've added in the public method `MappingBuilder::withClientIp` since it is a user-facing method but I didn't find any comments around so I thought it would be better to abide by the convention.

RafeArnold (commented):
could the relevant openapi (swagger) doc also be updated to document this new feature. and the wiremock.org docs in the wiremock.org repo

lhcopetti:
> could the relevant openapi (swagger) doc also be updated to document this new feature. and the wiremock.org docs in the wiremock.org repo

I added an entry for the client IP matcher in the docs and it is visible in the admin swagger-ui. Please, let me know if that works.

image

leeturner:
Hi @lhcopetti could you please pull in the latest main branch so we can resolve the current test failures

lhcopetti:
Sure, @leeturner . I rebased my changes.

leeturner:
Awesome, thank you for your contribution

PR #3026, opened 2025-04-15
Extend LogNormal to add an optional maxValue

paul-poulain-vodafone:
Extend LogNormal to add an optional maxValue to force long tails to be truncated and the value resampled if greater than the max. The reason this might be necessary is that with some log normal configurations, a curve with quite a long tail can be experienced that can cause unexpected or unwanted timeouts in upstream systems. Setting a maximum delay value can prevent this from happening. If the value generated is greater than the max, then it is resampled a number of times (10) and then finally, the max of the two values is taken (to prevent resampling indefinitely)

Documentation changes will be raised in a subsequent PR but look as follows:

image

## References

- This changes replaces [this original PR](https://github.com/wiremock/wiremock/pull/1694) which had `CappedLogNormal` as a new subclass of `LogNormal` 

## Submitter checklist

### DIFF
--- a/src/main/resources/swagger/schemas/equal-to-json-pattern.yaml
+++ b/src/main/resources/swagger/schemas/equal-to-json-pattern.yaml
@@ -2,9 +2,15 @@ title: JSON equals
 type: object
 properties:
   equalToJson:
-    type: string
-    example: |
-      { "message": "hello" }
+    oneOf:
+      - type: object
+        description: The JSON object to match.
+        example:
+          message: hello
+      - type: string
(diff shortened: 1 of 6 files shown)


paul-poulain-vodafone:
@tomakehurst and all, this is a replacement PR for #1694 that enhances `LogNormal` rather than adding a new subclass, as Tom proposed, thanks.

paul-poulain-vodafone:
Documentation PR is here: https://github.com/wiremock/wiremock.org/pull/353

leeturner:
Thank you for your contribution

PR #3028, opened 2025-04-23
Separate jetty

RafeArnold:


Mahoney:
A few questions

1. Should we separate out the servlet API only classes into a `wiremock-servlet` module? They could potentially be used by a server implementation other than Jetty.
2. Can we consolidate all the classes in the `wiremock-jetty` module into a single package? Either `com.github.tomakehurst.wiremock.jetty` or `com.github.tomakehurst.wiremock.jetty12`. It seems odd that they are scattered around as they currently are. We can probably then reduce the visibility to `package` nearly everywhere other than the service implementation, reducing our exposure to backwards incompatible changes in future.
3. Should the module be `wiremock-jetty` or `wiremock-jetty12`? My feeling is that it should be the latter. We have often ended up supporting two versions of Jetty and may well do again when Jetty 13 rears its head.
4. Can we lose some of the interfaces and abstract classes that were there to support multiple jetty versions? Like `JettyHttpServer` and `JettyUtils`.

On a side note - as and when Jetty 13 does happen I think that rather than have a common `wiremock-jetty` module for common code between Jetty 12 and Jetty 13 we should just accept the duplication rather than have a bad abstraction between the two that will probably cease to be correct with Jetty 14.

RafeArnold:
@Mahoney i agree with all your points and think they should be completed. they will be completed in subsequent prs with the intent of not expanding this already large pr

Mahoney (approved):

PR #3040, opened 2025-05-12
allow StubMappingTransformer to return errors.

RafeArnold:
## Submitter checklist


wiremock-common/src/main/java/com/github/tomakehurst/wiremock/recording/SnapshotOutputFormatter.java
```diff
@@ -20,21 +20,24 @@
...
-    public SnapshotRecordResult format(List<StubMapping> stubMappings) {
-      return SnapshotRecordResult.idsFromMappings(stubMappings);
+    public SnapshotRecordResult format(
+        List<StubMapping> stubMappings, List<? extends RecordError> errors) {
+      return SnapshotRecordResult.idsFromMappings(stubMappings, errors);
     }
   };
 
-  public abstract SnapshotRecordResult format(List<StubMapping> stubMappings);
+  public abstract SnapshotRecordResult format(
```
Mahoney, line 39: Can this method (and its overrides) also be package private?

Mahoney (approved):

wiremock-common/src/main/java/com/github/tomakehurst/wiremock/recording/SnapshotOutputFormatter.java
Mahoney: I think the methods can be package private, but the class needs to revert to being public because it is exposed as public properties of `RecordSpec` and `RecordSpecBuilder`

PR #3049, opened 2025-05-16
When using a dynamic port always bind to loopback

Mahoney:
By default, a wiremock instance (including the mock host) binds to address 0.0.0.0. When binding to a random port, that port is only unique for bind address 0.0.0.0; another application may be listening on the same TCP/IP port for a more specific bind address (like localhost). If so, and you request that port as localhost, you will talk to the other application, and your test will fail.

By binding explicitly to the IP address of the loopback adapter this problem is avoided.

Assumption here is that you would always want to do this... whether this is a correct assumption...

Of course you can still explicitly set the `bindAddress` to whatever you like.

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/BindAddressIssueTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/BindAddressIssueTest.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (C) 2025-2026 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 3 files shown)


Mahoney:
This probably ought to have some tests... going to be hard to make them deterministic though.

Still, can at least test that calling `dynamicPort()` without (or after) calling `bindAddress` sets it to the loopback, and that calling `bindAddress` _after_ `dynamicPort()` overwrites it.

Could also add a test as so, which would now always pass and in the past have been flaky:

* Start a WireMock on bindAddress loopback with dynamicPort
* Start a WireMock on default bindAddress with dynamicPort
* Make a request apparently to the second, and check it actually goes to the second. If they had the same port it would go to the first. This will require configuring the stubs and retrieving the log without using a network request...

PR #3071, opened 2025-06-18
jetty 12 improvements

RafeArnold:
implements most of the suggestions provided by @Mahoney in https://github.com/wiremock/wiremock/pull/3028

still todo: move servlet code to separate module. this is currently being prevented by `WireMockHttpServletRequestAdapter`'s dependency on jetty to extract form parameters from the request.


tomakehurst:
I think this is obsolete now. Please reopen if not @RafeArnold

PR #3084, opened 2025-06-25
__admin/requests query string filters for include and exclude

ericdriggs:
## References

https://github.com/wiremock/wiremock/issues/3083
https://github.com/wiremock/wiremock.org/pull/368

## Description

Implement query string filters include and exclude for __admin/requests endpoint 

simple matching on include and exclude query string

example: __admin/requests?include=foo&exclude=bar

AND: __admin/requests?include=foo&include=bar
OR: __admin/requests?include=foo,bar

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/AdminApiTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/AdminApiTest.java
@@ -231,7 +231,6 @@ class AdminApiTest extends AcceptanceTestBase {
 
   @Test
   void getFilteredLoggedRequests() {
-    dsl.stubFor(get(urlPathEqualTo("/received-request")).willReturn(aResponse()));
 
     for (int i = 1; i <= 10; i++) {
       final String id = "foo_" + i;


wiremock-common/src/main/resources/swagger/wiremock-admin-api.json
```diff
@@ -503,6 +503,24 @@
             "schema": {
               "type": "string"
             }
+          },
+          {
+            "description": "Exclude requests matching parameter. Multiple instances AND (include=foo&include=bar). Separate values with comma for OR (include=foo,bar).",
```
leeturner: Should the examples in this description read `exclude` instead of `include`?

wiremock-common/src/main/resources/swagger/wiremock-admin-api.yaml
```diff
@@ -256,6 +256,18 @@ paths:
           example: 2016-10-05T12:33:01Z
           schema:
             type: string
+        - description: Exclude requests matching parameter. Multiple instances AND (include=foo&include=bar). Separate values with comma for OR (include=foo,bar).
```
leeturner: Should the examples in this description read `exclude` instead of `include`?

wiremock-common/src/main/resources/swagger/wiremock-admin-api.json
ericdriggs: Thank you. Fixed.

wiremock-common/src/main/resources/swagger/wiremock-admin-api.yaml
ericdriggs: Thank you. Fixed.

leeturner:
This is looking really nice.  Could you run the spotlessapply gradle task to sort out the formatting issues.  Also, I was wondering, for completeness, whether it would be nice to have a test in `AdminApiTest` as we do for the `limit` and `since` query params ?

ericdriggs:
Thanks.
Formatting updated. 
AdminApiTest added.

softarn:
Since I'm looking into some filtering in this endpoint I got interested.

It is not obvious that `exclude` and `include` will filter only on the URL. Should we name it just `url` instead? That would remove the exclude option though since it would only be `?url=foo` to include URLs with foo in it, not sure if we should do exact matching there or not either.

Another solution to this problem would perhaps be to add filtering so that these events are never stored in the journal in the first place.

wiremock-common/src/main/java/com/github/tomakehurst/wiremock/admin/Conversions.java
```diff
@@ -39,4 +43,47 @@ public static Date toDate(QueryParameter parameter) {
...
+          }
+        } else {
+          if (actual.contains(item)) {
+            return true;
+          }
+        }
+      }
+      return false;
+    };
+  }
```
softarn, line 88: Looking at this logic it is a bit difficult to read and I'm not sure that it is working. Looks like we only require one include to be matched not respecting the AND behaviour documented. We need to loop through the entire list of items and not just return when we find one that fulfilled the requirement.

src/test/java/com/github/tomakehurst/wiremock/admin/ConversionsTest.java
```diff
@@ -51,4 +68,131 @@ void throwsExceptionWhenFirstParameterValueIsInvalidDate() {
...
+  @Test
+  void includesAndMatchTrue() {
+    // given
+    QueryParameter exclude = new QueryParameter("exclude", List.of());
+    QueryParameter include = new QueryParameter("include", List.of("key1=val1", "key1=val1b"));
+    // when
+    Predicate<ServeEvent> predicate = toPredicate(exclude, include);
+    // then
+    assertTrue(predicate.test(serveEvent));
+  }
```
softarn, line 108: To test the AND functionality you'd have to add another test to check that `new QueryParameter("include", List.of("key1=val1", "non-existing"));` returns false

src/test/java/com/github/tomakehurst/wiremock/admin/ConversionsTest.java
```diff
@@ -51,4 +68,131 @@ void throwsExceptionWhenFirstParameterValueIsInvalidDate() {
     // when + then
     assertThrows(InvalidInputException.class, () -> Conversions.toDate(queryParameter));
   }
+
+  final String url = "endpoint?key1=val1&key1=val1b&key2=val2";
```
softarn, line 72: The subtle difference between `val1` and `val1b` makes reading of these tests harder than they need to be

ericdriggs:
> It is not obvious that exclude and include will filter only on the URL. Should we name it just url instead? That would remove the exclude option though since it would only be ?url=foo to include URLs with foo in it, not sure if we should do exact matching there or not either.

Agreed on naming clarity. The naming could be more clear like excludeURL and includeURL. The exclude is actually a lot more important than include for my cases, where infrastructure health check calls are over 99.99% of the calls.

> Another solution to this problem would perhaps be to add filtering so that these events are never stored in the journal in the first place.

I looked at journal storage filtering initially but the scope for that was orders of magnitude larger. It would touch a lot more of the code base, has non-obvious implications for lifecycle, and to do so consistently and completely would also require implementing at the command line. Also, it would likely be far less flexible since you'd need to specify a single strategy before startup due to lifecycle.

I'll do the following:

- rename parameters for url clarity
- refactor test variables too look less ambiguous
- expand tests to ensure coverage for multiple values
- see if I can simplify comprehensibility of the predicate logic. 
- in relevant methods, clarify the cardinality of `QueryParameter ` -- it's non-obvious in that it can have multiple values.

PR #3087, opened 2025-06-26
fix: deletion of persisted stubs.

RafeArnold:
* delete persisted stubs when they are removed via request matching.
* run stub listeners _before_ deleting persisted stubs.
* do not attempt to remove stub if no matching stub exists.

## Submitter checklist


wiremock-common/src/main/java/com/github/tomakehurst/wiremock/core/MappingsSaver.java
```diff
@@ -17,13 +17,14 @@
...
 import java.util.List;
+import java.util.UUID;
 
 public interface MappingsSaver {
   void save(List<StubMapping> stubMappings);
 
   void save(StubMapping stubMapping);
 
-  void remove(StubMapping stubMapping);
+  void remove(UUID stubMappingId);
```
tomakehurst, line 27: This is a breaking change so we should consider creating an overload rather than removing the `StubMapping` version of the method signature.

RafeArnold: i was planning to add a migration step for v3.x that would add the overloaded method and deprecate the existing one, to allow for smooth transition to v4.x

tomakehurst: Why deprecate/remove it at all? If we leave it in there, nobody has to change their extensions when they upgrade to 4.x.

RafeArnold: the reason for taking it out/the reason for the pr is that allowing implementations to have access to the full stub meant different implementations used different methods for removing stubs, creating inconsistencies that meant some stubs were removed from the stub store but not the persisted stubs. limiting the implementations to only having access to the ids means this can't happen.

if the implementations should only ever remove a stub by its id then a method that provides the full stub isnt necessary and leads to potential bugs.

tomakehurst: OK, that makes sense. It's probably worth the breaking change in that case.

PR #3092, opened 2025-06-30
remove stubs in bulk where possible.

RafeArnold:
to allow implementations of MappingsSaver to be more efficient when persisting stubs.

this change necessitated adding the ability to delete stubs in bulk via the admin api, so RemoveMatchingStubMappingTask was updated to accept multiple stubs as well as a single stub mapping.

## References

https://github.com/wiremock/wiremock/pull/3075

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/RemoveStubMappingsAcceptanceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/RemoveStubMappingsAcceptanceTest.java
@@ -1,0 +1,233 @@
+/*
+ * Copyright (C) 2016-2025 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock;
+
+import static com.github.tomakehurst.wiremock.client.WireMock.aResponse;
+import static com.github.tomakehurst.wiremock.client.WireMock.get;
+import static com.github.tomakehurst.wiremock.client.WireMock.listAllStubMappings;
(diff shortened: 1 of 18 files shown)


src/test/java/com/github/tomakehurst/wiremock/RemoveStubMappingsAcceptanceTest.java
```diff
@@ -0,0 +1,233 @@
...
+import org.junit.jupiter.api.Test;
+
+class RemoveStubMappingsAcceptanceTest extends AcceptanceTestBase {
+
+  File rootDir;
+
+  @BeforeEach
+  public void init() throws InterruptedException {
+    rootDir = setupServerWithTempFileRoot();
+    super.init();
```
tomakehurst: Are we confident this isn't already done by the test framework due to the `@BeforeEach` annotation on the super method?

tomakehurst (commented):
One minor comment, otherwise LGTM.

src/test/java/com/github/tomakehurst/wiremock/RemoveStubMappingsAcceptanceTest.java
RafeArnold: yeah, the temp file created by that method is not available to tests in subclasses. also, AcceptanceTestBase sets up the server in the `@BeforeAll` method, not the `@BeforeEach` method (which only resets the wm server)

tomakehurst: I mean specifically the call to `super.init()` may not be necessary because I think JUnit will call all methods up the hierarchy with the annotation. But I might be wrong about that.

tomakehurst: Not a big deal here since what we're doing is idempotent, just a little less efficient.

RafeArnold: ah, i see. it's an explicit override of the super method, so would the overridden code even be available to junit to run?

RafeArnold: ive renamed the method so should remove any ambiguity.

tomakehurst: Hmmm...that seems to have broken the tests unfortunately

RafeArnold: oh dear. okay, ive set it back and not called the super method.

PR #3100, opened 2025-07-09
Use DateTimeFormatter instead of SimpleDateFormat in RenderableDate

pivovarit:
Replaced `SimpleDateFormat` with `DateTimeFormatter` in `RenderableDate` for the sake leveraging more user-friendly formatting.

For example, if you use template like ```{{now format="yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'"}}``` to format `"2025-07-07T15:23:11.123000Z"`, this is what you get:
- `DateTimeFormatter`: ```2025-07-07T15:23:11.123000Z``` (as expected)
- `SimpleDateFormat`:  ```2025-07-07T15:23:11.000123Z``` (with milliseconds converted into microseconds)

The new test fails when run against the SimpleDateFormat-based implementation:

```
RenderableDateTest > handlesCorrectlyFormatsExcessivePrecision() FAILED
    java.lang.AssertionError: 
    Expected: is "\"2024-10-18T15:48:24.010000Z\""
        but: was "\"2024-10-18T15:48:24.000010Z\""
```

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RenderableDateTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RenderableDateTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2024 Thomas Akehurst
+ * Copyright (C) 2024-2025 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,6 +19,7 @@ import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 
 import com.github.tomakehurst.wiremock.common.Json;
+import java.time.ZoneId;
 import java.util.Date;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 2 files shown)


src/test/java/com/github/tomakehurst/wiremock/extension/responsetemplating/helpers/RenderableDateTest.java
```diff
@@ -19,14 +19,23 @@
...
+import java.time.ZoneId;
 import java.util.Date;
 import org.junit.jupiter.api.Test;
 
 public class RenderableDateTest {
 
   @Test
   void writesToJsonInStringFormat() {
-    RenderableDate renderableDate = new RenderableDate(new Date(1729266504000L), null, null);
+    RenderableDate renderableDate = new RenderableDate(new Date(1729266504010L), null, null);
```
tomakehurst, line 30: Seems a bit dodgy that it's necessary to change this for the test to continue passing. What's happening here?

pivovarit: It's not necessary - added it to make the test more relevant. Default formatter truncates milliseconds, but the assertion won't show that because milliseconds part is just zeroes

pivovarit: Another change I'd like to submit it so actually stop cutting milliseconds off by the default formatter

tomakehurst:
I'm concerned this is not a completely backwards-compatible change. To your knowledge, are there date/time patterns that will work differently post this change?

pivovarit:
DateTimeFormatter supports all the letter from the official reference: https://docs.wiremock.io/response-templating/dates-and-times#format-string-reference

When it comes to parsing behaviour, generally, DateTimeFormatter is more strict by default

For example:
- format: `"dd/MM/yyyy"`
- input: `"1/1/2020"`

SimpleDateFormat parses it correctly despite format expecting minimum two digits per day/month, the other throws `DateTimeParseException`. That's probably the major difference.

To maximize compatibility, we could fall back to SimpleDateFormat if default DateTimeFormatter fails

tomakehurst:
@leeturner WDYT?

Seems like 4.x beta is a good place to make what probably amounts to a fairly minor set of breaking changes.

leeturner:
> @leeturner WDYT?
> 
> Seems like 4.x beta is a good place to make what probably amounts to a fairly minor set of breaking changes.

I would say definitely 4.x beta

pivovarit:
A backward-compatible change that would mitigate the original issue would be to expand the default format to include milliseconds - no one would be forced to recreate ISO-8601 with milliseconds by hand (just like I did)

Issue #3090, opened 2025-06-30
Original request URL no longer available in StubMapping for use by custom MappingsSource implementations


bjornarl:
### Proposal

Before release 3.13.0 and #3008 a custom implementation of `MappingsSource` (which can be passed to `WireMockConfiguration.mappingSource()`) could always use the full URL of the original request when determining how to store the mapping files.
With the changes in #3008, the JSON request mapping that used to look like:
```
"url" : "/v1.0/groups?%24select=id%2CdisplayName&%24filter=resourceProvisioningOptions%2FAny%28x%3Ax+eq+%27Team%27%29",
```
now looks like:
```
"urlPath" : "/v1.0/groups",
"queryParameters" : {
  "$select" : {
	"hasExactly" : [ {
	  "equalTo" : "id,displayName"
	} ]
  },
  "$filter" : {
	"hasExactly" : [ {
	  "equalTo" : "resourceProvisioningOptions/Any(x:x eq 'Team')"
	} ]
  }
}
```

This is due to the change in `RequestPatternTransformer`, but it does not appear possible to inject a custom `RequestPatternTransformer` at runtime as an extension in order to force the old behavior.

When `MappingsSource.save(List<StubMapping> stubMappings)` is called, the `stubMapping.getRequest().getUrl()` is no longer present if there were query arguments. It is sort of possible to manually reconstruct the URL using code but doing so is tedious, requires conditional code for the various matcher types, and there is no way to know which parameters went through decoding and thus whether to reverse the decoding.

In addition to the use case of a custom `MappingsSource` wanting to know the original full URL from the `StubMapping`, a second use case is allowing the stored JSON mapping files to be read/grepped by outside tools to match on specific original URLs.

Is one of the following an acceptable update to support these use cases?
1. Add a copy of the full original URL to the `StubMapping` and persisted mapping JSON "request", e.g., in an `"originalUrl"` field.
2. Make it possible to use a custom `RequestPatternTransformer` (via extensions) that can have the old behavior.

### Reproduction steps

This test case currently fails in 3.13.1 but would pass in 3.12.1.

```java
package com.github.tomakehurst.wiremock.extension.mappingssource;

import java.io.File;
import java.util.ArrayList;
import java.util.List;
import java.util.UUID;

import org.junit.jupiter.api.AfterEach;
import org.junit.jupiter.api.BeforeEach;
import org.junit.jupiter.api.Test;

import com.github.tomakehurst.wiremock.AcceptanceTestBase;
[...]

PR #3104, opened 2025-07-11
fix: preserve raw url in stub mapping request (#3090)

bjornarl:
Preserves the original raw URL (including query parameters) among stub mapping request parameters, for use by custom mappings sources and/or tools that analyze the JSON mappings files directly.

Please consider suitability for backporting to 3.x.

## References

Implements #3090 which was opened due to the effects of #3008.

## Submitter checklist


tomakehurst:
I don't think adding a new top-level model element is a good idea here, but we could add this under the `metadata` element.

We've been discussing this internally for a while i.e. having a "originating request" element documenting not just the URL but the complete request.

Something like this:

```json
{
  "request": { ... },
  "response": { ... },

  "metadata": {
    "wiremock": {
      "originatingRequest": {
        "absoluteUrl": "https://api.example.com/v1/leads",
        "method": "POST",
        "headers": {
          "Accept": [
            "application/json"
          ],
          "Accept-Encoding": [
            "gzip, deflate, br"
          ]
        },
        "body": "{ \"name\": \"John Doe\", \"email\": \"john.doe@example.com\" }"
      }
    }
  }
}
```

bjornarl:
Thanks for the follow up. Being able to instead get the original URL from `metadata/wiremock/originatingRequest/absoluteUrl` in the mapping should allow us to move forward with the two scenarios that are currently broken for us when we try to use 3.13.0+.

I'm guessing that is unlikely to become available in 3.x but I see 4.x is nearing non-beta so we will wait and hope it becomes available there so that we can get back to staying on a recent version.

Issue #3119, opened 2025-07-24
Using ISO 8601 date for since query parameter in admin api generates error "2025-07-03 is not a valid ISO8601 date"


softarn:
### Proposal

The `since` query parameter should be able to take an ISO 8601 date (YYYY-MM-DD) as a value. Currently, it requires a full date and time with offset like this `2023-10-07T00:00:00Z`. The expected case is to support both these formats.

Also, the error message says
```
"2011-12-03 is not a valid ISO8601 date"
```
Expected would be something like
```
"Invalid format for since, supported formats are `2025-07-24` and `2025-07-24T10:20:03Z`"
```

Since we are using `ZonedDateTime.parse()` method we are also supporting the format `2007-12-03T10:15:30+01:00[Europe/Paris]`. We should consider dropping support for including `[Europe/Paris]` since it is very unexpected, and I don't believe anyone has managed to find it and use it.

### Reproduction steps

Start latest version of Wiremock standalone

Go to this URL `/__admin/requests?since=2011-12-03`
This will be the response 
```
{
  "errors" : [ {
    "code" : 10,
    "source" : {
      "pointer" : "since"
    },
    "title" : "2011-12-03 is not a valid ISO8601 date"
  } ]
}
```

### References

When ZonedDateTime was introduced
https://github.com/wiremock/wiremock/commit/bbef0c26171e9e2685d8f85dfdcdc72454247e9c

PR #3120, opened 2025-07-24
Fix since query parameter format (#3119)

softarn:
Changed two things

- Remove the Dates class that was not used enough in production code to justify its existence.
- Fixed documentation and error message to communicate format needed for since query parameter. Removed support for zoneId in the param, e.g. `[Europe/Paris]` in the end of `2007-12-03T10:15:30+01:00[Europe/Paris]` as this is not needed, undocumented and not likely to be used by anyone.

I opted out to add support for only dates like `2025-07-23` as it is unclear if is in UTC or not and might cause unexpected behaviour for users of API. Better to force them to be specific and include time and offset/UTC.

## References

https://github.com/wiremock/wiremock/issues/3119

## Submitter checklist


leeturner:
Hi @softarn sorry for the delay in getting to this, we have had a lot of change in the codebase recently.  Would it be possible to rebase and we can take a fresh look at the PR?

Issue #3122, opened 2025-07-31
@WireMockTest annotation overwrites "javax.xml.transform.TransformerFactory" system property


akutschera:
### Proposal

Assume I am using Saxon to do some XSLT transformations. I am setting the System property `javax.xml.transform.TransformerFactory` to point to the Saxon Implementation.
When I run a test with `@WireMockTest`, it overwrites this (see below)

I would expect that while `@WireMockTest` uses the `newDefaultInstance()`, it lets the rest of the code use whatever TransformerFactory that I have configured.

### Reproduction steps

A test project to reproduce this is in the attached zip file

[wiremock-xml.zip](https://github.com/user-attachments/files/21532313/wiremock-xml.zip)

```
@WireMockTest
public class PropertyTest {
    
    @Test
    void test () {
        TransformerFactory actual = TransformerFactory.newInstance();
        assertEquals(net.sf.saxon.TransformerFactoryImpl.class, actual.getClass());
        assertEquals("net.sf.saxon.TransformerFactoryImpl", System.getProperty("javax.xml.transform.TransformerFactory"));
    }
}
```
The test will fail with
```
org.opentest4j.AssertionFailedError: 
Expected :class net.sf.saxon.TransformerFactoryImpl
Actual   :class com.sun.org.apache.xalan.internal.xsltc.trax.TransformerFactoryImpl
```

This test fails, when the WireMockTest annotation is present and it succeeds when it is not present.
The last versions of wiremock that worked, are 3.12.1 and 4.0.0-beta.4. Versions 3.13.0 and 4.0.0-beta.5 fail with the annotation.

### References

This was probably introduced with #2975 that uses the newDefaultInstance() instead of the newInstance() and then overwrites the System property [here](https://github.com/wiremock/wiremock/blob/c8a2a5e984c6ef3d2bb196cf89648b62be8a3ee5/wiremock-common/src/main/java/com/github/tomakehurst/wiremock/common/xml/Xml.java#L54).

akutschera:
When I look at this a bit more deeply, it seems that #2975 was actually introduced to fix a problem that exists when another TransformerFactoryImpl like Saxon exists (#2454).
We are not only setting the `javax.xml.transform.TransformerFactory` in Xml.java, we are also calling XmlUnit.setTransformerFactory, which sets the property also. Since this is addressing a performance issue (see #1048), it is probably a good idea to still set it, but only when it isn't already set.

tomakehurst:
[...]

Issue #1048, opened 2018-12-11
Xpath matcher poor performance


philippe-granet:
Wiremock version: 2.19.0

First of all, thank you for your work on wiremock, a very useful tool!

We are using wiremock for stubbing SOAP services in development environment.
We also tried to use Wiremock for performance tests, but there is some bottleneck:
* **javax.xml.transform.TransformerFactory.newInstance()**
This factory is instanciate everywhere in XMLUnit library and WireMock, but algorithm  to determine the implementation class to load is heavy:
https://docs.oracle.com/javase/7/docs/api/javax/xml/transform/TransformerFactory.html#newInstance()

Simply set system property `-Djavax.xml.transform.TransformerFactory=<implementation to load>` resolve this bottleneck

* **javax.xml.xpath.XPathFactory.newInstance()**
This factory is instanciate everywhere in XMLUnit library and WireMock, but algorithm  to determine the implementation class to load is heavy:
https://docs.oracle.com/javase/7/docs/api/javax/xml/xpath/XPathFactory.html#newInstance()
https://github.com/openjdk-mirror/jdk7u-jaxp/blob/master/src/javax/xml/xpath/XPathFactoryFinder.java#L156

Simply set system property `-Djavax.xml.xpath.XPathFactory:http://java.sun.com/jaxp/xpath/dom=<implementation to load>` resolve this bottleneck

* **XMLUnit.setXPathFactory** and **XMLUnit.setTransformerFactory**
Predefine theses factories in XMLUnit also improve performance
 
**Simple test case:**
```java
@Test
public void perfXmlTest() {
	perfRunDatas("without optimisations");

	// Optimisation 1
	String xPathFactoryImpl = XPathFactory.newInstance().getClass().getName();
	System.setProperty(XPathFactory.DEFAULT_PROPERTY_NAME + ":" 
		+ XPathFactory.DEFAULT_OBJECT_MODEL_URI, xPathFactoryImpl);
	perfRunDatas("with XPathFactory system property");

	// Optimisation 2
	String transformerFactoryImpl = TransformerFactory.newInstance().getClass().getName();
	System.setProperty(TransformerFactory.class.getName(), transformerFactoryImpl);
	perfRunDatas("with XPathFactory/TransformerFactory system properties");

	// Optimisation 3
	XMLUnit.setTransformerFactory(transformerFactoryImpl);
	perfRunDatas("with XPathFactory/TransformerFactory system properties and TransformerFactory predefined in XMLUnit");

[...]

PR #3125, opened 2025-08-07
Only set javax.xml.transform.TransformerFactory when it is not already set

akutschera:
This will only set the ` javax.xml.transform.TransformerFactory` System property if it is unset.

This allows users to define their own TransformerFactory, e.g. Saxon and use the DefaultFactory if they do not set the property.
We still set the property for XmlUnit so there should be no performance issues there (#1048)

I made two commits, one for the production code and one for the Test. The test helped me write the code, but since it uses a long TestTransformerFactory class, it might be a bit much for just one test. Feel free to discard the commit if it is too much.

## References

#3122 

## Submitter checklist


buijs-dev:
Is there an ETA for this change?

Issue #899, opened 2018-03-08
Should equalToJson treat null field and non-existent field the same?


yunj:
## Bug reports
The request mapping is configured with the following request body.
```json
{
   "email":"email@hotmail.com",
   "homeAddress":{
      "line1":"1 Road",
      "suburb":"St",
      "postCode":"1071"
   },
   "mobile":{
      "subscriberNumber":"21123456",
      "countryCode":"61"
   }
}
```
**But when the actual request with null field, it fails the matching.**
```json
{
   "email":"email@hotmail.com",
   "homeAddress":{
      "line1":"1 Road",
      "line2":null,
      "suburb":"St",
      "postCode":"1071"
   },
   "mobile":{
      "subscriberNumber":"21123456",
      "countryCode":"61"
   }
}
```
equalToJson(xxx, true, true) is not perfect because we do want it failing if line2 does have value rather than null.

## Feature requests
From the restful PUT point of view, these two jsons are the same, so I think it should treat null field and non-existent field the same at this case.

tomakehurst:
Please can you post the full stub mapping and requests that replicates this

yunj:
Hi @tomakehurst,

Thanks very much for your reply. The stub mapping is as follows:
```json
{
   "id":"051f1c61-51ba-474f-8eef-715fa00ce805",
   "request":{
      "url":"/api/v1/my/contact",
      "method":"PUT",
      "headers":{
         "token":{
            "equalTo":"34c91881-986c-4381-ac81-5d2ca85a987a"
         }
      },
      "bodyPatterns":[
         {
            "equalToJson":"{\r\n  \"email\": \"email@hotmail.com\"\r\n  ,\"homeAddress\": {\r\n  \"line1\": \"1 Road\"\r\n}\r\n\r\n}\r\n"
         }
      ]
   },
   "response":{
      "status":204
   },
   "uuid":"051f1c61-51ba-474f-8eef-715fa00ce805"
}
````

**Comment by tomakehurst** [comment]
And the request(s)?

**Comment by yunj** [comment]
Hi @tomakehurst ,

Not sure if this is what you need? When doing a restful PUT request with the following body.
```json
{
   "email":"email@hotmail.com",
   "homeAddress":{
      "line1":"1 Road",
      "line2":null
   }
}
```

yunj:
Hi @tomakehurst 

Does wiremock delegate the json comparison to the third-party library?

Thanks.

tomakehurst:
[...]

PR #3135, opened 2025-08-16
feat: add extra flag to ignore null elements in equalToJsonPattern

ohhyungsuh:
Changed one thing

Added an extra flag ignoreNullElements to equalToJson.

This PR achieves the following:

Added a new constructor that accepts the ignoreNullElements flag.
Overloaded the equalToJson method to use JSON Unit's TREATING_NULL_AS_ABSENT option, allowing null fields in JSON objects to be ignored during comparison.
Users can now specify ignoreNullElements in addition to ignoreArrayOrder and ignoreExtraElements.
Verified behavior with seven dedicated tests covering different scenarios involving null fields.
This makes JSON comparisons more flexible, especially when dealing with optional fields that might be set to null.
## References

- #899 

## Submitter checklist


udhayakumarth:
Hi @tomakehurst, Looks like this is a good feat. Why its not taken into consideration?

PR #3147, opened 2025-08-31
Qodo tests stress

laysaOlivveira:


tomakehurst:
Thanks for making this contribution. I'm not going to merge it, but since by the look of it this is a university assignment, I'll provide some feedback as to why:
* The PR has no description. We have a PR template partly for this reason, so that maintainers can understand the context of a change before considering whether to merge it, reject it or request alterations.
* This is a large PR that changes multiple concerns at the same time. Large PRs are harder to review, and combining concerns mean that it's not possible to accept some and reject others - it's all or nothing.
* This deletes a number of existing performance test code without explaining why.
* This comments out (which should usually never be committed) a number of functional tests and disables another without explaining why.

PR #3154, opened 2025-09-14
Soft parsing date to prevent DateTimeParseException

r0goyal:
In https://github.com/wiremock/wiremock/blob/master/wiremock-common/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java, for some reason (need some context here), isISOOffsetDateTime(String encoded) is called for each URL fragment.

In real world, most of the fragments would not be dates. However right now, for each non date fragment, a DateTimeParseException is thrown which is subsequently caught and then false is returned. This causes too many instances of DateTimeParseException being thrown leading to higher memory utilisation.

Potentially, a better way to do this is to do a fast check on whether the fragment indeed conforms to an ISO date. If it does, we can proceed with current parsing mechanism otherwise we can return early without incurring the cost of creating an expensive exception stack trace.

## References

(https://github.com/wiremock/wiremock/issues/3153)

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/common/UrlsTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/common/UrlsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2014-2024 Thomas Akehurst
+ * Copyright (C) 2014-2025 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -22,11 +22,9 @@ import com.github.tomakehurst.wiremock.http.QueryParameter;
 import java.net.URI;
 import java.net.URLDecoder;
 import java.nio.charset.StandardCharsets;
-import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Map;
-
 import org.junit.jupiter.api.Assertions;
 import org.junit.jupiter.api.Test;
 
@@ -189,16 +187,15 @@ public class UrlsTest {
     var dateAsString = "2023-02-30T10:00:00+01:00";
     var trickyDate = URI.create("/date?date=" + dateAsString);
     params = Urls.splitQuery(trickyDate);
-    Assertions.assertEquals(URLDecoder.decode(dateAsString, StandardCharsets.UTF_8), params.get("date").firstValue());
+    Assertions.assertEquals(
+        URLDecoder.decode(dateAsString, StandardCharsets.UTF_8), params.get("date").firstValue());
   }
 
-
   @Test
-    public void doesNotDecodeValidIsoOffsetDateTimeLikeString() {
+  public void doesNotDecodeValidIsoOffsetDateTimeLikeString() {
     var dateAsString = "2023-02-28T10:00:00+01:00";
     var trickyDate = URI.create("/date?date=" + dateAsString);
     params = Urls.splitQuery(trickyDate);
     Assertions.assertEquals(dateAsString, params.get("date").firstValue());
   }
-
 }


r0goyal:
Hi @tomakehurst. Requesting your review for the above mentioned possible performance optimisation.

buildSrc/.kotlin/sessions/kotlin-compiler-4605141833734471586.salive
tomakehurst, line 1: Looks like this probably shouldn't have been committed

wiremock-common/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
```diff
@@ -107,12 +109,22 @@ private static String decode(String encoded) {
...
+    exception handling in the case of a non-date string. If the soft parse succeeds, we
+    then do a full parse to ensure the string is a valid date.
+     */
+    ParsePosition pos = new ParsePosition(0);
+    TemporalAccessor temporalAccessor = ISO_OFFSET_DATE_TIME.parseUnresolved(encoded, pos);
+    if (temporalAccessor == null || pos.getIndex() != encoded.length()) {
+      return false;
+    }
     try {
       ISO_OFFSET_DATE_TIME.parse(encoded);
```
tomakehurst, line 123: Is there a way to avoid repeating the first phase of parsing here in the success case (because we've already done it)?

tomakehurst: If not we should have an extra test case demonstrating at least one case where making this call is necessary for a correct outcome.

buildSrc/.kotlin/sessions/kotlin-compiler-4605141833734471586.salive
r0goyal: Yes. My bad. Have removed this file.

wiremock-common/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
r0goyal: @tomakehurst  Avoiding repeat parsing is not possible. `parseUnresolved()` checks for syntax, not semantic correctness (Example - `2023-02-30T10:00:00+01:00` (Non leap year having 30 days in February)). 

As for test cases, currently this method is a private method 

`private static boolean isISOOffsetDateTime(String encoded)`

In order to test private method, I would have to either use some library like `Powermock` or make this method `public`. Would need your guidance on which route to take for this ? Another aspect is what exactly are we testing given that output of the method would remain same.

leeturner: Hi @r0goyal I don't think in this case we are thinking of testing the private method directly.  More that we want to make sure all the public methods that rely on the `decode` and then the `isISOOffsetDateTime` work in the same way now.  Also like Tom mentioned, it would be nice to have an extra test showing the call to `ISO_OFFSET_DATE_TIME.parse(encoded);` is required for the correct outcome.  Would a test parsing the date you gave above be a good one - (`2023-02-30T10:00:00+01:00`)?  It would show that the call to `ISO_OFFSET_DATE_TIME.parseUnresolved(encoded, pos)` would pass but then the call to `ISO_OFFSET_DATE_TIME.parse(encoded);` would fail and return a false from the method?

r0goyal: @leeturner Thanks for the clarification. Yes the case I mentioned above demonstrates the scenario where `ISO_OFFSET_DATE_TIME.parseUnresolved(encoded, pos)` would pass but `ISO_OFFSET_DATE_TIME.parse(encoded)` would fail. 

I have added a test case for the same. Requesting your review.

leeturner: Great stuff, thank you for adding the tests and thank you for your contribution

leeturner:
Could you run `./gradlew :spotlessApply` and commit any formatting changes it makes

r0goyal:
@leeturner Ran spotless plugin. Please check now.

leeturner:
Awesome, merged.  Thank you for your contribution

r0goyal:
Thanks @leeturner

PR #3163, opened 2025-09-18
Made changes to Notifier interface to improve performance

r0goyal:
## References

- TODO

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/extension/ExtensionLifeCycleAcceptanceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/extension/ExtensionLifeCycleAcceptanceTest.java
@@ -24,6 +24,8 @@ import com.github.tomakehurst.wiremock.AcceptanceTestBase;
 import com.github.tomakehurst.wiremock.common.Notifier;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.function.Supplier;
+
 import org.apache.hc.core5.http.io.entity.StringEntity;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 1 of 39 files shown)


leeturner:
Hi, many thanks for the PR.  We were were thinking that it might be better if the new methods were added to the existing methods instead of replacing them which would make this less of a breaking change to all the people in the community that are relying on them in their code.  Also, the uses of with a fixed string probably don' need to be moved over to the supplier based methods. Would you be able to make those changes ?

r0goyal:
Hi @leeturner. New methods can be added. However this would still be a breaking change from interface definition point of view. For a user of the library, they will then have to implement two set of methods (one containing `Supplier` and other without it) and might confuse users more (as to when what method is called). Would you like me to go ahead with that ?

leeturner:
Hi @r0goyal I see your point.  I guess the number of changes in this PR has been driven by the fact that all notifier call sites have needed to be updated which, like we discussed above, would mean all code in the community would need to be updated.  Would a good compromise here be keeping the old methods as defaults in the interface:

```java
public interface Notifier {
    String KEY = "Notifier";
    
    // New lazy‐friendly methods ― abstract so implementors override
    void info(Supplier<String> message);
    void error(Supplier<String> message);
    void error(Supplier<String> message, Throwable t);
    
    // Back‐compat default methods which delegate to the new ones
    default void info(String message) {
        info(() -> message);
    }
    default void error(String message) {
        error(() -> message);
    }
    default void error(String message, Throwable t) {
        error(() -> message, t);
    }
}
```

This would focus the changes to notifier implementors and would allow us to keep this PR smaller due to only needing to update the call sites where there is an expensive operation required to generate the string.  All the static string call sites could be left to the old methods.

What are your thoughts ?

r0goyal:
@leeturner Got your point. I was looking at it from library user's point of view instead of library developer. I have made the changes as per your suggestion. During the change I realised only `info()` method needs additional `Supplier` based method. `error` methods do not need anything extra so I have reverted those changes.

src/test/java/com/github/tomakehurst/wiremock/extension/ExtensionLifeCycleAcceptanceTest.java
```diff
@@ -65,8 +66,8 @@ public void reset() {
     }
 
     @Override
-    public void info(String message) {
```
tomakehurst: Let's keep this method for backwards compatibility, with a default impl delegating to the new supplier-based one.

PR #3172, opened 2025-09-26
Fix: case sensitivity in proxy request header removal

grachevko:
Ensure that header removal during proxying is case-insensitive by converting removeProxyRequestHeaders to lowercase before comparison.

This fixes issues where headers with different casing would not be properly removed from the proxied request.


leeturner:
Hi, many thanks for the PR.  Would it be possibe to add tests for the change you have made ?

grachevko:
Hi @leeturner, is this enought?

tomakehurst:
Thanks!

PR #3179, opened 2025-10-03
Draft: feat: adding various string number matchers

dirkbolte:
# Conflicts:
#	wiremock-common/src/main/java/com/github/tomakehurst/wiremock/client/WireMock.java

Adding matchers for numeric comparison:

- `equalToNumber`
- `largerThan` + `largerThanEqual`
- `lessThan` + `lessThanEqual`

## References

- TODO

## Submitter checklist

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (C) 2016-2023 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.matching;
+
+import com.github.tomakehurst.wiremock.client.WireMock;
+import com.github.tomakehurst.wiremock.common.Json;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.skyscreamer.jsonassert.JSONAssert.assertEquals;
+
+public class EqualToNumberPatternTest {
+
+    @Test
+    public void failsForNoMatchOnEqualsInt() {
+        StringValuePattern pattern = WireMock.equalToNumber("5");
+        assertFalse(pattern.match("1").isExactMatch());
+        assertThat(pattern.match("1").getDistance(), is(0.01));
+    }
+
+    @Test
+    public void failsForNoMatchOnEqualsIntContains() {
+        StringValuePattern pattern = WireMock.equalToNumber("1");
+        assertFalse(pattern.match("1111").isExactMatch());
+        assertThat(pattern.match("1111").getDistance(), is(0.04));
+    }
+
+    @Test
+    public void failsForNoMatchOnEqualsFloat() {
+        StringValuePattern pattern = WireMock.equalToNumber("5.5");
+        assertFalse(pattern.match("1.1").isExactMatch());
+        assertThat(pattern.match("1.1").getDistance(), is(0.01));
+    }
+
+    @Test
+    public void succeedsForExactMatchOnEqualsInt() {
+        StringValuePattern pattern = WireMock.equalToNumber("1");
+        assertTrue(pattern.match("1").isExactMatch());
+        assertThat(pattern.match("1").getDistance(), is(0.0));
+    }
+
+    @Test
+    public void succeedsForExactMatchOnEqualsFloat() {
+        StringValuePattern pattern = WireMock.equalToNumber("1.1111");
+        assertTrue(pattern.match("1.1111").isExactMatch());
+        assertThat(pattern.match("1.1111").getDistance(), is(0.0));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToNumberFromJson() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
+
+        assertThat(stringValuePattern, instanceOf(EqualToNumberPattern.class));
+        assertThat(stringValuePattern.getValue(), is("1"));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToNumberFromJsonWithIgnoreCase() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
+
+        assertThat(stringValuePattern, instanceOf(EqualToNumberPattern.class));
+        assertThat(stringValuePattern.getValue(), is("1"));
+    }
+
+    @Test
+    public void correctlySerialisesToJson() throws Exception {
+        assertEquals(
+            "{ \"equalToNumber\": \"1\" }",
+            Json.write(new EqualToNumberPattern("1")),
+            false);
+    }
+
+    @Test
+    public void noMatchOnNullValue() {
+        assertThat(WireMock.equalToNumber("1").match(null).isExactMatch(), is(false));
+    }
+
+    @Test
+    public void noMatchOnStringValue() {
+        assertThat(WireMock.equalToNumber("1").match("a string").isExactMatch(), is(false));
+    }
+
+    @Test
+    public void noMatchOnExpectedString() {
+        assertThat(WireMock.equalToNumber("a string").match("1").isExactMatch(), is(false));
+    }
+
+    @Test
+    public void objectsShouldBeEqualOnSameExpectedValue() {
+        var a = new EqualToNumberPattern("1");
+        var b = new EqualToNumberPattern("1");
+        var c = new EqualToNumberPattern("2");
+
+        Assertions.assertEquals(a, b);
+        Assertions.assertEquals(a.hashCode(), b.hashCode());
+        Assertions.assertEquals(b, a);
+        Assertions.assertEquals(b.hashCode(), a.hashCode());
+        assertNotEquals(a, c);
+        assertNotEquals(a.hashCode(), c.hashCode());
+        assertNotEquals(b, c);
+        assertNotEquals(b.hashCode(), c.hashCode());
+    }
+}
--- a/src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanEqualPatternTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanEqualPatternTest.java
@@ -1,0 +1,126 @@
+/*
+ * Copyright (C) 2016-2023 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.matching;
+
+import com.github.tomakehurst.wiremock.client.WireMock;
+import com.github.tomakehurst.wiremock.common.Json;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.instanceOf;
+import static org.hamcrest.Matchers.is;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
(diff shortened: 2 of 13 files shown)


dirkbolte:
This PR is still draft: just want to allow someone to have a look wether the implementation is going into the right direction.

src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
```diff
@@ -0,0 +1,122 @@
...
+        assertThat(stringValuePattern.getValue(), is("1"));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToNumberFromJsonWithIgnoreCase() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
+
+        assertThat(stringValuePattern, instanceOf(EqualToNumberPattern.class));
+        assertThat(stringValuePattern.getValue(), is("1"));
+    }
```
RafeArnold: what's the difference between these tests?

src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
```diff
@@ -0,0 +1,122 @@
...
+
+    @Test
+    public void noMatchOnStringValue() {
+        assertThat(WireMock.equalToNumber("1").match("a string").isExactMatch(), is(false));
+    }
+
+    @Test
+    public void noMatchOnExpectedString() {
+        assertThat(WireMock.equalToNumber("a string").match("1").isExactMatch(), is(false));
+    }
```
RafeArnold: should it even be possible to create a numeric matcher with a non-numeric value?

src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanEqualPatternTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void succeedsForExactMatchOnLargerThanFloat() {
+        StringValuePattern pattern = WireMock.largerThanEqual("1.1111");
+        assertTrue(pattern.match("1.1112").isExactMatch());
+        assertThat(pattern.match("1.1112").getDistance(), is(0.0));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToFromJson() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
```
RafeArnold:
```suggestion
        StringValuePattern stringValuePattern = Json.read("{ \"largerThanEqual\": \"1\" }", StringValuePattern.class);
```

src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanEqualPatternTest.java
```diff
@@ -0,0 +1,126 @@
...
+        assertThat(stringValuePattern.getValue(), is("1"));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToFromJsonWithIgnoreCase() {
+        StringValuePattern stringValuePattern = Json.read("{ \"largerThanEqual\": \"1\" }", StringValuePattern.class);
+
+        assertThat(stringValuePattern, instanceOf(LargerThanEqualPattern.class));
+        assertThat(stringValuePattern.getValue(), is("1"));
+    }
```
RafeArnold: again, what's the difference between these two tests?

src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanPatternTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void succeedsForExactMatchOnLargerThanFloat() {
+        StringValuePattern pattern = WireMock.largerThan("1.1111");
+        assertTrue(pattern.match("1.1112").isExactMatch());
+        assertThat(pattern.match("1.1112").getDistance(), is(0.0));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToFromJson() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
```
RafeArnold:
```suggestion
        StringValuePattern stringValuePattern = Json.read("{ \"largerThan\": \"1\" }", StringValuePattern.class);
```

src/test/java/com/github/tomakehurst/wiremock/matching/LessThanEqualPatternTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void succeedsForExactMatchOnLessThanEqualFloat() {
+        StringValuePattern pattern = WireMock.lessThanEqual("1.1111");
+        assertTrue(pattern.match("1.111").isExactMatch());
+        assertThat(pattern.match("1.111").getDistance(), is(0.0));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToFromJson() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
```
RafeArnold:
```suggestion
        StringValuePattern stringValuePattern = Json.read("{ \"lessThanEqual\": \"1\" }", StringValuePattern.class);
```

src/test/java/com/github/tomakehurst/wiremock/matching/LessThanPatternTest.java
```diff
@@ -0,0 +1,126 @@
...
+    @Test
+    public void succeedsForExactMatchOnLessThanFloat() {
+        StringValuePattern pattern = WireMock.lessThan("1.1111");
+        assertTrue(pattern.match("1.111").isExactMatch());
+        assertThat(pattern.match("1.111").getDistance(), is(0.0));
+    }
+
+    @Test
+    public void correctlyDeserialisesEqualToFromJson() {
+        StringValuePattern stringValuePattern = Json.read("{ \"equalToNumber\": \"1\" }", StringValuePattern.class);
```
RafeArnold:
```suggestion
        StringValuePattern stringValuePattern = Json.read("{ \"lessThan\": \"1\" }", StringValuePattern.class);
```

RafeArnold:
any objection to changing `largerThan` to `greaterThan`? "greater than" seems more like the standard naming convention.

RafeArnold:
out of curiosity, is there a situation where the `equalToNumber` matcher would provide matching functionality that's not available with the standard `equalTo` matcher?

dirkbolte:
> any objection to changing `largerThan` to `greaterThan`? "greater than" seems more like the standard naming convention.

No objection, you're right.

dirkbolte:
> out of curiosity, is there a situation where the `equalToNumber` matcher would provide matching functionality that's not available with the standard `equalTo` matcher?

I also wondered whether this can be merged in any way. Examples can be comparisons like: "1 == 1.0" . Pure string comparison would not work here.

RafeArnold:
> > out of curiosity, is there a situation where the `equalToNumber` matcher would provide matching functionality that's not available with the standard `equalTo` matcher?
> 
> I also wondered whether this can be merged in any way. Examples can be comparisons like: "1 == 1.0" . Pure string comparison would not work here.

good example. can a test be added for that example?

src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
dirkbolte: was taking the `equalTo` tests as reference, so: c&p error. Removed.

src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanEqualPatternTest.java
dirkbolte: was taking the equalTo tests as reference, so: c&p error. Removed.

src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
dirkbolte: I tried to align the behavior with what I saw in e.g. the date matchers: format is evaluated on actual matching and wrong input results in a `false`/non-match. Should I change it to an `IllegalStateException`?

src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanEqualPatternTest.java
dirkbolte: done.

src/test/java/com/github/tomakehurst/wiremock/matching/LargerThanPatternTest.java
dirkbolte: done.

src/test/java/com/github/tomakehurst/wiremock/matching/LessThanPatternTest.java
dirkbolte: done.

src/test/java/com/github/tomakehurst/wiremock/matching/LessThanEqualPatternTest.java
dirkbolte: done.

src/test/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPatternTest.java
tomakehurst: My inclination is to make the expected value a number rather than a string, since a string that's not a number can never be valid in this situation.

Then it's a compile error (or JSON parsing) rather than having to throw an exception.

Date/time matching is a bit of an odd case because of the complexity of resolution. In general the ideal is to provide feedback as early as possible to the user that they've created an invalid matcher, so if we can do it with the compiler in this case, that's a good thing.

dirkbolte: Just to double check - want to prevent running into the wrong direction: currently these matchers inherit from the `StringValueMatcher`. To make this work properly it feels l have to create a `NumberValueMatcher` hierarchy. Do I got it right?

tomakehurst: @dirkbolte `StringValuePattern` is intended to match parts of a request that are strings (which is nearly all of it usually), but the expected value doesn't have to be a string - there's no constraint on what the constructor can require.

For instance, `EqualToJsonPattern` internally constructs a `JsonNode` and stores some additional matching rules, then matches the incoming request element against these.

It would be similarly possible I think to create a matcher taking a number in its constructor as the expected value, then coerces the actual value from a string to a number before evaluating the match.

dirkbolte: While doing so, I ran into a small name clash: there were two methods `lessThan` accepting a number value. Thus I changed all new matchers to contain `Number`.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPattern.java
```diff
@@ -0,0 +1,39 @@
...
+
+public class EqualToNumberPattern extends AbstractNumberPattern {
+
+    public EqualToNumberPattern(@JsonProperty("equalToNumber") Number testValue) {
+        super(testValue);
+    }
+
+    public String getEqualToNumber() {
+        return expectedValue;
+    }
```
RafeArnold: can these methods return a number instead of the string representation, now that the number representation is available via `AbstractNumberPattern.number`?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/AbstractNumberMatchResult.java
```diff
@@ -0,0 +1,97 @@
...
+    protected static final float NO_MATCH = 1.0f;
+    private static final double MAX_LOG_DIFF = 176.0;
+    private final String expectedValue;
+    private final String value;
+
+
+    public AbstractNumberMatchResult(String expectedValue, String value) {
+        this.expectedValue = expectedValue;
+        this.value = value;
+    }
```
RafeArnold: can `expectedValue` be a number, now that the patterns have access to the number representation when they construct the `AbstractNumberMatchResult`? saves having to parse the string to a double unnecessarily.

src/test/java/com/github/tomakehurst/wiremock/SubServeEventsAcceptanceTest.java
```diff
@@ -37,7 +37,7 @@ void nonMatchDiffsAreSavedAsSubEvents() {
     ServeEvent serveEvent = wm.getAllServeEvents().get(0);
     SubEvent subEvent = serveEvent.getSubEvents().stream().findFirst().get();
     assertThat(subEvent.getType(), is("REQUEST_NOT_MATCHED"));
-    assertThat(subEvent.getTimeOffsetNanos(), greaterThan(0L));
+    assertThat(subEvent.getTimeOffsetNanos(), org.hamcrest.Matchers.greaterThan(0L));
```
RafeArnold: are these qualifiers necessary now the WireMock methods are suffixed with "Number"?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPattern.java
dirkbolte: That wouldn't work. It would result in the value being converted into a number in JSON, so it cannot be parsed by the same matcher anymore:

```json
{
  "lessThanEqualNumber" : 1.0
}
```

It might be sensible to have another set of matchers for numbers though.

RafeArnold: is it not possible to have the json deserialiser for these matchers expect a number instead of a string?

src/test/java/com/github/tomakehurst/wiremock/SubServeEventsAcceptanceTest.java
dirkbolte: done.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/AbstractNumberMatchResult.java
dirkbolte: done.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/EqualToNumberPattern.java
dirkbolte: I just checked: in the current settings of Jackson in wiremock, Jackson converts numbers to String if `asText` is required. So the constructor still works as is. So we just have to decide whether deserialization should be to number or to string. I would assume Number is more natural, so I will adapt it.

PR #3231, opened 2025-11-17
Split out JUnit 4 and 5 modules

tomakehurst:
Create separate JAR artifacts for the JUnit 4 and 5 integration code.

### DIFF
--- a/wiremock-junit4/build.gradle.kts
+++ b/wiremock-junit4/build.gradle.kts
@@ -5,9 +5,7 @@ plugins {
 dependencies {
     api(project(":wiremock-common"))
 
-    api(libs.junit4) {
-        exclude(group = "org.hamcrest", module = "hamcrest-core")
-    }
+    api(libs.junit4)
 
     testImplementation(testFixtures(project(":")))
     testImplementation(libs.apache.http5.client)


wiremock-common/build.gradle.kts
```diff
@@ -36,17 +33,12 @@ dependencies {
     implementation(libs.json.path) {
         exclude(group = "org.ow2.asm", module = "asm")
     }
-    implementation(libs.junit.platform.commons)
     implementation(libs.slf4j.api)
     implementation(libs.xmlunit.legacy) {
         exclude(group = "junit", module = "junit")
     }
```
Mahoney, line 39: Slightly worried about this one now... though hopefully tests would reveal any resulting NoSuchMethodError

wiremock-junit4/build.gradle.kts
```diff
@@ -0,0 +1,36 @@
+plugins {
+    id("wiremock.common-conventions")
+}
+
+dependencies {
+    api(project(":wiremock-common"))
+
+    api(libs.junit4) {
+        exclude(group = "org.hamcrest", module = "hamcrest-core")
```
Mahoney: I wonder if we should remove this exclusion now you are consciously depending on JUnit 4 rather than accidentally bringing it in with WireMock?

Or replace it with this:
```
dependencies {
    modules {
        module("org.hamcrest:hamcrest-core") {
            replacedBy("org.hamcrest:hamcrest")
        }
        module("org.hamcrest:hamcrest-library") {
            replacedBy("org.hamcrest:hamcrest")
        }
    }
}
```
since they have now merged `library` and `core` into the one module?

Mahoney (commented):
Couple of queries

wiremock-common/build.gradle.kts
tomakehurst: Good catch. I'll see if I can kick the tyres on this with some external projects e.g. mock-host.

wiremock-junit4/build.gradle.kts
tomakehurst: Did you mean to paste something extra after "Or replace it with this:"?

Mahoney: No, duplicated the line with slightly different wording by mistake

tomakehurst: I guess it's up to the user to whether they want to use the depended-on hamcrest so we could just remove the exclusion and leave it at that.

Mahoney: Agreed

PR #3241, opened 2025-11-19
Decouple from Apache HTTP Client 5

Mahoney:
Decouples the `HttpAdminClient` from the `ApacheBackedHttpClient` and `ApacheHttpClientFactory`, using our own abstraction (`com.github.tomakehurst.wiremock.http.client.HttpClient`) instead.

This allows us to move the Apache implementaiton into a separate module, loaded as an extension, decoupling `wiremock-core` from Apache HTTP 5.

Well... nearly. We still have `ApacheHttpHostNameMatcher` in core, `org.apache.hc.client5.http.ssl.DefaultHostnameVerifier.verify(String, X509Certificate)` is used by the MITM TLS `CertificateGeneratingX509ExtendedKeyManager`. Unfortunately it's not easy to inline.

## Submitter checklist

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StaticExtensionLoader.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StaticExtensionLoader.java
@@ -71,7 +71,7 @@ public final class StaticExtensionLoader<T extends Extension> {
   private T pickMostAppropriateFrom(Stream<T> candidates) {
     return candidates
         .min(defaultFactoryLast())
-        .orElseThrow(this::couldNotFindSuitableClientException);
+        .orElseThrow(this::couldNotFindSuitableImplementationException);
   }
 
   private Comparator<Object> defaultFactoryLast() {
@@ -88,7 +88,7 @@ public final class StaticExtensionLoader<T extends Extension> {
     };
   }
 
-  private FatalStartupException couldNotFindSuitableClientException() {
+  private FatalStartupException couldNotFindSuitableImplementationException() {
     return new FatalStartupException(
         "No suitable "
             + type.getSimpleName()


wiremock-httpclient-apache5/build.gradle.kts
```diff
@@ -0,0 +1,55 @@
...
+}
+
+publishing {
+    publications {
+        create<MavenPublication>("mavenJava") {
+            artifactId = tasks.jar.get().archiveBaseName.get()
+            from(components["java"])
+
+            pom {
+                name = "WireMock Core"
```
tomakehurst: This needs updating I think

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StaticExtensionLoader.java
```diff
@@ -0,0 +1,107 @@
...
+        return 0;
+      } else if (o1IsDefault) {
+        return 1;
+      } else {
+        return -1;
+      }
+    };
+  }
+
+  private FatalStartupException couldNotFindSuitableClientException() {
```
tomakehurst: This method name seems overly specific

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StaticExtensionLoader.java
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    if (!extensions.isEmpty()) {
+      return pickMostAppropriateFrom(extensions.stream());
+    }
+
+    return pickMostAppropriateFrom(serviceLoader.get());
+  }
+
+  public StaticExtensionLoader<T> setDefaultInstance(T defaultInstance) {
```
tomakehurst: Calling this defaultInstance is slightly confusing in this context as it actually has the highest precedence, whereas default tends to suggest lowest.

wiremock-httpclient-apache5/build.gradle.kts
Mahoney: Fixed

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StaticExtensionLoader.java
Mahoney: Renamed to `specificInstance`

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StaticExtensionLoader.java
Mahoney: Renamed to `couldNotFindSuitableImplementationException`

PR #3248, opened 2025-11-21
Make primary domain classes immutable

tomakehurst:
WireMock currently has a messy, mixed approach to immutability in its core classes, which makes reasoning about when e.g. a stub mapping can me modified very difficult.

Additionally it makes it far harder to safely implement Store bindings for off-heap persistence, as it's currently possible to rely on mutation in-place with the default (in-memory) implementations.

This change takes steps towards full immutability of the stub definition classes.

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
@@ -17,6 +17,7 @@ package com.github.tomakehurst.wiremock.recording;
 
 import static com.github.tomakehurst.wiremock.common.ParameterUtils.getFirstNonNull;
 import static java.util.stream.Collectors.toList;
+import static java.util.stream.Collectors.toMap;
 
 import com.github.tomakehurst.wiremock.common.Urls;
 import com.github.tomakehurst.wiremock.matching.RequestPattern;
@@ -24,6 +25,7 @@ import com.github.tomakehurst.wiremock.stubbing.Scenario;
 import com.github.tomakehurst.wiremock.stubbing.StubMapping;
 import java.net.URI;
 import java.util.*;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 
@@ -42,7 +44,7 @@ class ScenarioProcessor {
         stubsGroupedByRequest.entrySet().stream()
             .filter(entry -> entry.getValue().size() > 1)
             .collect(
-                Collectors.toMap(
+                toMap(
                     Map.Entry::getKey,
                     Map.Entry::getValue,
                     (entry1, entry2) -> entry1,
@@ -59,14 +61,12 @@ class ScenarioProcessor {
       stubsInScenario.addAll(putStubsInScenario(scenarioIndex, batch));
     }
 
+    Map<UUID, StubMapping> stubsInScenarioById =
+        stubsInScenario.stream().collect(toMap(StubMapping::getId, Function.identity()));
+
     return stubMappings.stream()
-        .map(
-            originalStub ->
-                stubsInScenario.stream()
-                    .filter(stubMapping -> stubMapping.getId().equals(originalStub.getId()))
-                    .findFirst()
-                    .orElse(originalStub))
-        .collect(toList());
+        .map(originalStub -> stubsInScenarioById.getOrDefault(originalStub.getId(), originalStub))
+        .toList();
   }
 
   private List<StubMapping> putStubsInScenario(int scenarioIndex, List<StubMapping> stubMappings) {


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/SnapshotStubMappingPostProcessor.java
```diff
@@ -83,24 +84,27 @@ Pair<List<RecordError>, List<StubMapping>> process(
...
-        bodyExtractor.extractInPlace(stubMapping);
-      }
-    }
+    return stubMappings.stream()
+        .map(
+            sm ->
+                bodyExtractMatcher.match(sm.getResponse()).isExactMatch()
+                    ? bodyExtractor.extractInPlace(sm)
+                    : sm)
+        .collect(Collectors.toList());
```
Mahoney: Could just be `.toList();`

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/standalone/JsonFileMappingsSource.java
```diff
@@ -47,7 +47,7 @@ public JsonFileMappingsSource(FileSource mappingsFileSource, FilenameMaker filen
   @Override
   public void save(List<StubMapping> stubMappings) {
     for (StubMapping mapping : stubMappings) {
-      if (mapping != null && mapping.isDirty()) {
+      if (mapping != null) {
```
Mahoney, line 50: I suppose this was really just an efficiency saving, so not an issue that we don't check if it's changed?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
```diff
@@ -65,23 +78,25 @@ private void putStubsInScenario(int scenarioIndex, List<StubMapping> stubMapping
...
+                          if (i == 1) {
+                            stub.setRequiredScenarioState(Scenario.STARTED);
+                          } else {
+                            stub.setRequiredScenarioState(scenarioName + "-" + i);
+                          }
+                          if (i < stubMappings.size()) {
+                            stub.setNewScenarioState(scenarioName + "-" + (i + 1));
+                          }
+                        }))
+        .collect(toList());
```
Mahoney, line 101: If you're OK with using Guava's `com.google.common.collect.Streams` you could use this:
```suggestion
    return Streams.mapWithIndex(
            stubMappings.stream(),
            (it, i) ->
                it.transform(
                    stub -> {
                      stub.setScenarioName(scenarioName);
                      if (i == 0) {
                        stub.setRequiredScenarioState(Scenario.STARTED);
                      } else {
                        stub.setRequiredScenarioState(scenarioName + "-" + (i + 1));
                      }
                      if (i < (stubMappings.size() - 1)) {
                        stub.setNewScenarioState(scenarioName + "-" + (i + 2));
                      }
                    }))
        .toList();
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
```diff
@@ -46,17 +48,28 @@ void putRepeatedRequestsInScenarios(List<StubMapping> stubMappings) {
...
     }
+
+    return stubMappings.stream()
+        .map(
+            originalStub ->
+                stubsInScenario.stream()
+                    .filter(stubMapping -> stubMapping.getId().equals(originalStub.getId()))
+                    .findFirst()
+                    .orElse(originalStub))
+        .collect(toList());
```
Mahoney: What do you think of this? It slightly alters the semantics (the returned stubs will be in a different order, stubs not in a scenario first, then stubs in a scenario). Not sure if that matters.
```java
  List<StubMapping> putRepeatedRequestsInScenarios(List<StubMapping> stubMappings) {
    Map<RequestPattern, List<StubMapping>> stubsGroupedByRequest =
        stubMappings.stream()
            .collect(
                Collectors.groupingBy(
                    StubMapping::request,
                    LinkedHashMap::new,
                    Collectors.toCollection(LinkedList::new)));

    Pair<Map<RequestPattern, List<StubMapping>>, Map<RequestPattern, List<StubMapping>>> partitioned = partition(stubsGroupedByRequest, entry -> entry.getValue().size() > 1);
    Map<RequestPattern, List<StubMapping>> groupsWithMoreThanOneStub = partitioned.a;
    List<StubMapping> singleStubs = partitioned.b.values().stream().map(s -> s.get(0)).toList();

    final List<StubMapping> stubsInScenario = new LinkedList<>();
    int scenarioIndex = 0;
    for (Map.Entry<RequestPattern, List<StubMapping>> entry :
        groupsWithMoreThanOneStub.entrySet()) {
      scenarioIndex++;
      final List<StubMapping> batch = new LinkedList<>(entry.getValue());
      Collections.reverse(batch);

      stubsInScenario.addAll(putStubsInScenario(scenarioIndex, batch));
    }

    return Stream.concat(singleStubs.stream(), stubsInScenario.stream()).toList();
  }

  private static <K, V> Pair<Map<K, V>, Map<K, V>> partition(
      Map<K, V> input, Predicate<Map.Entry<K, V>> predicate) {
    Map<Boolean, Map<K, V>> partitioned =
        input.entrySet().stream()
            .collect(
                Collectors.partitioningBy(
                    predicate, Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue)));

    return new Pair<>(partitioned.get(true), partitioned.get(false));
  }
```

Mahoney: Or if you don't like the big suggestion, how about this?
```suggestion
    Map<UUID, StubMapping> stubsInScenarioById =
        stubsInScenario.stream().collect(toMap(StubMapping::getId, Function.identity()));

    return stubMappings.stream()
        .map(originalStub -> stubsInScenarioById.getOrDefault(originalStub.getId(), originalStub))
        .toList();
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMapping.java
```diff
@@ -201,102 +177,227 @@ public int comparePriorityWith(StubMapping otherMapping) {
...
 
-  public void setMetadata(Metadata metadata) {
-    this.metadata = metadata;
+  @Override
+  public boolean isMulti() {
+    return false;
   }
 
   @Override
   public boolean equals(Object o) {
```
Mahoney, line 198: Do we still need equals and hashCode now it's a record class? Would avoid them going stale when new fields are added.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/standalone/JsonFileMappingsSource.java
tomakehurst: I spent a bit of time looking at it and yes, it's an efficiency measure, and only one that has any effect when you hit the save API endpoint (which saves everything currently persistable). Slightly worrying that I didn't have to fix any tests after removing it, but still...

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
tomakehurst: Would be nice to remove Guava as a long-term goal, given its size.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMapping.java
tomakehurst: Yes - I tried using the defaults, but these include `insertionIndex`, which causes a lot of problems.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/ScenarioProcessor.java
Mahoney: Fair, just a shame that the JVM doesn't provide `Stream.mapWithIndex` out of the box.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMapping.java
tomakehurst: I'm sure it could be argued that `insertionIndex` is dodgy as hell and should go away...

src/test/java/com/github/tomakehurst/wiremock/matching/RequestPatternBuilderTest.java
```diff
@@ -90,20 +92,29 @@ void likeRequestPatternWithCustomMatcher() {
...
-    assertThat(newRequestPattern, is(requestPattern));
+    assertThat(
+        newRequestPattern
+            .match(
+                ImmutableRequest.create()
+                    .withMethod(GET)
+                    .withAbsoluteUrl("https://localhost/match-me")
+                    .build())
+            .isExactMatch(),
+        is(true));
```
RafeArnold, line 110: this would technically pass if `like` _didn't_ actually copy the initial matcher and just created an ANY request pattern, so could a negative match also be added?

PR #3276, opened 2025-12-19
improvements to immutable data classes.

RafeArnold:
* make and mark all collection/collection-like properties non-nullable.
* fix HttpHeaders builder not catering for multiple values for the same
header name.
* add additional methods to HttpHeaders builder and clarify names of
existing methods.
* fix inline custom matcher methods of RequestPattern: changed
getMatcher to getInlineCustomMatcher and have it actually return the
custom matcher.
* have old builders (RequestPatternBuilder and
ResponseDefinitionBuilder) delegate to new builders.
* add a couple methods to ProxyResponseDefinitionBuilder.

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMapping.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMapping.java
@@ -38,6 +38,7 @@ import org.jspecify.annotations.NonNull;
 @JsonIgnoreProperties({
   "$schema", "uuid"
 }) // $schema allows this to be added as a hint to IDEs like VS Code
+@JsonInclude(Include.NON_NULL)
 @JsonDeserialize() // stops infinite recursion when deserializing as StubMappingOrMappings
 public final class StubMapping implements StubMappingOrMappings {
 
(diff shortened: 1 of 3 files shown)


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StubMapping.java
```diff
@@ -31,11 +32,13 @@
 import java.util.Objects;
 import java.util.UUID;
 import java.util.function.Consumer;
+import org.jspecify.annotations.NonNull;
 
 @JsonPropertyOrder({"id", "name", "request", "newRequest", "response"})
 @JsonIgnoreProperties({
   "$schema", "uuid"
 }) // $schema allows this to be added as a hint to IDEs like VS Code
+@JsonInclude(Include.NON_NULL)
```
tomakehurst, line 41: Are all these Include.NON_NULLs actually necessary since we enable this globally on the object mapper in `Json`?

RafeArnold: they're not necessary for WireMock's `Json`, but useful when serialising the classes using other object mappers

Issue #3031, opened 2025-04-24
Support WebSockets


leonard84:
### Proposal

Please add support for WebSockets. 
Initially, I thought it would not be possible with the WireMock architecture, but then I saw https://wiremock.org/docs/grpc/, which gave me hope that it would be.

At the minimum, a pass-through for the proxy would be greatly appreciated, as in my use case, the server I'm proxying is used for both regular HTTP requests and WebSockets and I need to stub some HTTP requests, while proxying the WebSockets to the server.

leeturner:
Websocket support was implemented in https://github.com/wiremock/wiremock/pull/3286

Have you tried this for your use case ?  I realise that proxying websocket requests is yet to be implemented but was interested in whether you have used any of the websocket mocking features ?

PR #3286, opened 2026-01-04
Websockets mocking support

tomakehurst:
Adds support for mocking websocket exchanges.

Currently this includes:
* Establishing a websocket connection on any URL path (currently can't be accepted/rejected selectively)
* A new message stub type, mapping a trigger to one or more actions
* A stub action responding to a message on the same channel (websocket session) 
* Responding by targeting other channels using matching against the originating HTTP request that opened the channel.
* Using stubs or HTTP requests as triggers for message actions e.g. send a message on all websocket sessions when a specific HTTP URL is POSTed to.
* Message action templating.
* A message action transformer extension point.
* Message journal, similar to the request journal.
* Text or binary message bodies, defined inline or as files.

Still to come:
* Other extension points analogous ones used for HTTP.
* Other protocols - SSE, webhooks
* Extension points allowing other protocols to use the message mocking core (will enable the gRPC extension to support streaming properly).

## References

#3031

### DIFF
--- a/schemas/wiremock-stub-mapping-or-mappings.json
+++ b/schemas/wiremock-stub-mapping-or-mappings.json
@@ -891,7 +891,7 @@
       "type" : "object",
       "description" : "Pattern for matching messages",
       "properties" : {
-        "channelPattern" : {
+        "channel" : {
           "$ref" : "#/definitions/request-pattern",
           "description" : "Pattern to match the channel's initiating request"
         },


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/MessageJournal.java
```diff
@@ -0,0 +1,48 @@
...
+
+  void reset();
+
+  void messageReceived(MessageServeEvent event);
+
+  void removeEvent(UUID eventId);
+
+  List<MessageServeEvent> removeEventsMatching(MessagePattern pattern);
+
+  List<MessageServeEvent> removeEventsForStubsMatchingMetadata(StringValuePattern metadataPattern);
```
RafeArnold, line 43: looks like this isn't tested anywhere, and the implementation is incomplete.

src/test/java/com/github/tomakehurst/wiremock/verification/MessageSerializationTest.java
```diff
@@ -0,0 +1,965 @@
...
+        }
+        """;
+
+    MessageStubMapping stub = Json.read(json, MessageStubMapping.class);
+
+    assertThat(stub.getId().toString(), is("d8e8fca2-dc0f-11db-8314-0800200c9a66"));
+    assertThat(stub.getName(), is("Deserialized stub"));
+    assertThat(stub.getPriority(), is(5));
+    assertThat(stub.getChannelPattern().getUrl(), is("/my-channel"));
+    assertThat(stub.getMessagePattern(), notNullValue());
```
RafeArnold: should we assert more for the value of `stub.getMessagePattern()`?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/SendMessageAction.java
```diff
@@ -0,0 +1,142 @@
...
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+
+public class SendMessageAction implements MessageAction {
+
+  private final MessageDefinition message;
+  private final RequestPattern targetChannelPattern;
+  private final boolean sendToOriginatingChannel;
```
RafeArnold: rather than this class holding fields for each possible target type (currently just `targetChannelPattern` and `sendToOriginatingChannel`), could the target of the send action be a single polymorphic field? then you couldnt get invalid states where `sendToOriginatingChannel` is true and `targetChannelPattern` is non-null. and would probably make adding new target types simpler in the future.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/Message.java
```diff
@@ -0,0 +1,106 @@
...
+      return Objects.equals(getBodyAsString(), o);
+    }
+    if (getClass() != o.getClass()) return false;
+    Message message = (Message) o;
+    return Objects.equals(body, message.body);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hashCode(getBodyAsString());
```
leeturner: Do we need to keep the implementations of `equals` and `hashcode` consistent? The `equals` method handles the `String` using `getBodyAsString()` as does `hashcode` but then performs full `Entity` equality on the `Message` instance which `hashcode` doesn't handle

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/websocket/WireMockWebSocketEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+      payload.get(data);
+      Message message =
+          MessageStubRequestHandler.resolveToMessage(MessageDefinition.fromBytes(data), null);
+      messageStubRequestHandler.processMessage(messageChannel, message);
+    }
+    callback.succeed();
+  }
+
+  @Override
+  public void onWebSocketClose(int statusCode, String reason) {
```
leeturner: Looks like this is deprecated in favour of - `onWebSocketClose(int, String, Callback)`

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/websocket/WireMockWebSocketEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+    if (messageChannel != null) {
+      messageStubRequestHandler.getMessageChannels().remove(messageChannel.getId());
+    }
+  }
+
+  @Override
+  public void onWebSocketError(Throwable cause) {}
```
leeturner: Do we want to do something here to notify of an error or is this handled elsewhere?  If we don't need it here we can remove it as it is the same as the default implementation

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/StoreBackedMessageJournal.java
```diff
@@ -0,0 +1,175 @@
...
+  private void removeOldEntries() {
+    if (maxEntries != null) {
+      while (store.getAllKeys().count() > maxEntries) {
+        store.removeLast();
+      }
+    }
+  }
+
+  private static MessagePattern withStubMetadataMatching(final StringValuePattern metadataPattern) {
+    return new MessagePattern(null, null) {
```
leeturner, line 178: This will always return false which doesn't seem correct

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/entity/Entity.java
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null || getClass() != o.getClass()) return false;
+    Entity entity = (Entity) o;
+    return Objects.equals(encoding, entity.encoding)
+        && Objects.equals(format, entity.format)
+        && Objects.equals(compression, entity.compression)
+        && Arrays.equals(getData(), entity.getData());
```
leeturner: It has been a while since I have looked at things like this but doesn't the `equals` method performing I/O where exceptions can be thrown break the `equals` contract ?

src/test/java/com/github/tomakehurst/wiremock/MessageActionTransformerAcceptanceTest.java
```diff
@@ -0,0 +1,233 @@
...
+  @Test
+  void multipleTransformersAreAppliedInOrder() {
+    wm =
+        new WireMockServer(
+                wireMockConfig()
+                    .dynamicPort()
+                    .extensions(
+                        new PrefixingMessageActionTransformer(),
+                        new SuffixingMessageActionTransformer()))
+            .startServer();
```
RafeArnold, line 75: this test doesnt actually test that ordering matters. the same result is output regardless of the order that the extensions are registered.

src/test/java/com/github/tomakehurst/wiremock/MessageTemplatingAcceptanceTest.java
```diff
@@ -0,0 +1,153 @@
...
+            .withName("Date template stub")
+            .withBody(equalTo("time"))
+            .willTriggerActions(
+                sendMessage("Current year: {{now format='yyyy'}}").onOriginatingChannel()));
+
+    WebsocketTestClient testClient = new WebsocketTestClient();
+    String url = websocketUrl("/date-test");
+
+    String response = testClient.sendMessageAndWaitForResponse(url, "time");
+    assertThat(response, is("Current year: 2026"));
```
RafeArnold: could this be made less brittle by dynamically retrieving the current system datetime?

src/test/java/com/github/tomakehurst/wiremock/MessageTemplatingAcceptanceTest.java
```diff
@@ -0,0 +1,153 @@
...
+  }
+
+  @Test
+  void templatesMessageBodyWithMultipleHelpers() {
+    wm = new WireMockServer(wireMockConfig().dynamicPort()).startServer();
+
+    wm.messageStubFor(
+        message()
+            .withName("Multi-helper template stub")
+            .withBody(matching(".*"))
```
RafeArnold: is it possible to not specify a body pattern at all and have the same effect as `.withBody(matching(".*"))`?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/admin/model/SendChannelMessageResult.java
```diff
@@ -0,0 +1,32 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.github.tomakehurst.wiremock.admin.model;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/** Result model for sending a message to channels. */
+public class SendChannelMessageResult {
```
RafeArnold, line 25: should this contain the ids of the channels that were messaged, not just the count?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannel.java
```diff
@@ -0,0 +1,34 @@
...
+import com.github.tomakehurst.wiremock.http.Request;
+import java.util.UUID;
+
+public interface MessageChannel {
+
+  ChannelType getType();
+
+  UUID getId();
+
+  Request getInitiatingRequest();
```
RafeArnold: this field feels specific to the channel type. if other channel types are introduced that don't originate from an http request, then this field won't be relevant.

src/testFixtures/java/com/github/tomakehurst/wiremock/testsupport/WebsocketTestClient.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class WebsocketTestClient {
+
+  private final WebSocketContainer websocketClient = ContainerProvider.getWebSocketContainer();
+
+  private final NotificationCapturingEndpoint endpoint = new NotificationCapturingEndpoint();
+
+  private Session persistentSession;
```
RafeArnold, line 41: a single session per client seems unfortunate.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/Message.java
tomakehurst: Good catch. Think this can now just be done the normal way.

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/websocket/WireMockWebSocketEndpoint.java
tomakehurst: Choosing to ignore this for now, but perhaps at some point we'll need to do something with this.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/StoreBackedMessageJournal.java
tomakehurst: Yes, unfinished by the look of it, good catch.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageAction.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package com.github.tomakehurst.wiremock.message;
+
+import com.fasterxml.jackson.annotation.JsonSubTypes;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+import com.github.tomakehurst.wiremock.extension.Extension;
+
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonSubTypes({@JsonSubTypes.Type(value = SendMessageAction.class, name = "send")})
+public interface MessageAction {
```
RafeArnold, line 24: could this be a sealed interface?

RafeArnold: and other interfaces like MessageTrigger and MessageChannel?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
```diff
@@ -0,0 +1,142 @@
...
+  }
+
+  public void add(MessageChannel channel) {
+    store.add(channel);
+  }
+
+  public void remove(UUID id) {
+    Optional<MessageChannel> channel = store.get(id);
+    store.remove(id);
+    channel.ifPresent(MessageChannel::close);
```
RafeArnold: this looks like a race condition. can `remove` just return an optional so there's no need to call `get` first?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
```diff
@@ -0,0 +1,142 @@
...
+    Message message = MessageStubRequestHandler.resolveToMessage(messageDefinition, null);
+    for (MessageChannel channel : matchingChannels) {
+      channel.sendMessage(message);
+    }
+    return matchingChannels.size();
+  }
+
+  public void clear() {
+    store.getAll().forEach(MessageChannel::close);
+    store.clear();
```
RafeArnold: this also looks like a race condition but harder to avoid unless clear also returns all the cleared channels.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
```diff
@@ -0,0 +1,267 @@
...
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+import java.util.function.Consumer;
+
+@JsonInclude(NON_EMPTY)
+public class MessageStubMapping {
```
RafeArnold: is it worth annotating these new data classes with jspecify annotations?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
```diff
@@ -0,0 +1,267 @@
...
+      @JsonProperty("id") UUID id,
+      @JsonProperty("name") String name,
+      @JsonProperty("priority") Integer priority,
+      @JsonProperty("trigger") MessageTrigger trigger,
+      @JsonProperty("actions") List<MessageAction> actions) {
+    this.id = id != null ? id : UUID.randomUUID();
+    this.name = name;
+    this.priority = priority;
+    this.trigger = trigger != null ? trigger : IncomingMessageTrigger.ANYTHING;
+    this.actions = actions != null ? actions : Collections.emptyList();
```
RafeArnold, line 72:
```suggestion
    this.actions = actions != null ? List.of(actions) : Collections.emptyList();
```

makes the class actually immutable. i'd advocate for adding a test for this.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/store/MessageJournalStore.java
```diff
@@ -0,0 +1,34 @@
...
+@Beta(justification = "Externalized State API: https://github.com/wiremock/wiremock/issues/2144")
+public interface MessageJournalStore extends Store<UUID, MessageServeEvent> {
+
+  Stream<MessageServeEvent> getAll();
+
+  void add(MessageServeEvent event);
+
+  void removeLast();
+
+  void registerEventListener(Consumer<? super StoreEvent<UUID, MessageServeEvent>> listener);
```
RafeArnold, line 33: feels like there should be a way to unregister listeners. could easily end up with leaks. returning a closeable is quite a nice way of doing it.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/StoreBackedMessageJournal.java
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    return store.getAll().filter(pattern::matches).findFirst();
+  }
+
+  @Override
+  public List<MessageServeEvent> waitForEvents(
+      MessagePattern pattern, int count, Duration maxWait) {
+    long deadline = System.currentTimeMillis() + maxWait.toMillis();
+    List<MessageServeEvent> collected = new ArrayList<>();
```
RafeArnold: what is the use of this list?

src/test/java/com/github/tomakehurst/wiremock/client/HttpAdminClientTest.java
```diff
@@ -194,6 +202,98 @@ void shouldParseErrorsLeniently() {
...
+    server.start();
+    try {
+      var client = WireMock.create().port(server.port()).buildAdminClient();
+      ListMessageChannelsResult result = client.listAllMessageChannels();
+
+      // No channels connected yet, so should be empty
+      assertThat(result.getChannels()).isEmpty();
+    } finally {
+      server.stop();
+    }
```
RafeArnold: feels like this isn't testing a whole lot

src/test/java/com/github/tomakehurst/wiremock/client/HttpAdminClientTest.java
```diff
@@ -194,6 +202,98 @@ void shouldParseErrorsLeniently() {
...
+    try {
+      var client = WireMock.create().port(server.port()).buildAdminClient();
+      List<MessageServeEvent> result =
+          client.waitForMessageEvents(MessagePattern.ANYTHING, 2, Duration.ofMillis(100));
+
+      assertThat(result).isEmpty();
+    } finally {
+      server.stop();
+    }
+  }
```
RafeArnold: also need the non-timeout versions of these?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
```diff
@@ -0,0 +1,142 @@
...
+ */
+package com.github.tomakehurst.wiremock.message;
+
+import com.github.tomakehurst.wiremock.matching.RequestMatcherExtension;
+import com.github.tomakehurst.wiremock.matching.RequestPattern;
+import com.github.tomakehurst.wiremock.store.MessageChannelStore;
+import java.util.*;
+import java.util.stream.Collectors;
+
+public class MessageChannels {
```
leeturner, line 24: There are a reasonable number of unused methods in this class and a number of places where we are calling `getAll()` and then filtering by open channels.  There is already a method called `getAllOpen()` in this class that could be used for that.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/entity/Entity.java
tomakehurst: I agree this is a bit dodgy (even if it's not contract violating). Not sure what to do about it at the moment...

tomakehurst: Dropping that part from `equals` doesn't break and tests so I'll do that for now and we can worry about it later if there's a reason to.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/MessageJournal.java
tomakehurst: Tested by this I think: https://github.com/wiremock/wiremock/pull/3286/files#diff-ada01dfba46fccd2b46021d25f852154bbf88de4c027644ea0654ad2ab463f31R371

src/test/java/com/github/tomakehurst/wiremock/verification/MessageSerializationTest.java
tomakehurst: Agreed

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/SendMessageAction.java
tomakehurst: Agree - this is on my own list of improvements to make

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageAction.java
tomakehurst: At least some of these will eventually be provided by extensions, and sealing will make this tricky I think.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannel.java
tomakehurst: Agree

src/test/java/com/github/tomakehurst/wiremock/MessageTemplatingAcceptanceTest.java
tomakehurst: It is implicitly anything if `withBody` is not called

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/admin/model/SendChannelMessageResult.java
tomakehurst: Probably a good idea

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
tomakehurst: Yes. Also on my list, but if you'd like to propose the changes that'd be very helpful.

src/test/java/com/github/tomakehurst/wiremock/MessageTemplatingAcceptanceTest.java
RafeArnold: yeah, so is `matching(".*")` achieving anything?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageAction.java
RafeArnold: at the moment it looks like we're just switching on the types, so having it sealed for now seems to make sense for catching missing branches at compile time. we could unseal them later if we take a different approach to handling different types.

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/websocket/WireMockWebSocketEndpoint.java
```diff
@@ -0,0 +1,77 @@
...
+      payload.get(data);
+      Message message =
+          MessageStubRequestHandler.resolveToMessage(MessageDefinition.fromBytes(data), null);
+      messageStubRequestHandler.processMessage(messageChannel, message);
+    }
+    callback.succeed();
+  }
+
+  @Override
+  public void onWebSocketClose(int statusCode, String reason, Callback callback) {
```
leeturner, line 68: Just wondering now we have implemented the non depreated method signature whether the `succeed` or `fail` needs to be called on the Callback?

```java
try {
      if (messageChannel != null) {
        messageStubRequestHandler.getMessageChannels().remove(messageChannel.getId());
      }
      callback.succeed();
    } catch (Exception e) {
      callback.fail(e);
    }
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/SendMessageAction.java
```diff
@@ -32,29 +32,26 @@
...
-      @JsonProperty("targetChannelPattern") RequestPattern targetChannelPattern,
-      @JsonProperty("sendToOriginatingChannel") Boolean sendToOriginatingChannel,
+      @JsonProperty("channelTarget") ChannelTarget channelTarget,
       @JsonProperty("transformers") List<String> transformers,
       @JsonProperty("transformerParameters") Parameters transformerParameters) {
     this.message = message;
-    this.targetChannelPattern = targetChannelPattern;
-    this.sendToOriginatingChannel =
-        sendToOriginatingChannel != null ? sendToOriginatingChannel : false;
+    this.channelTarget = channelTarget != null ? channelTarget : OriginatingChannelTarget.INSTANCE;
```
RafeArnold: are we happy with defaulting the channel target if it's missing, or should we require an explicit target?

src/testFixtures/java/com/github/tomakehurst/wiremock/testsupport/WebsocketTestClient.java
tomakehurst: Can live with this for now. Easily changed if it gets too messy.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannel.java
tomakehurst: Created `RequestInitiatedMessageChannel` to fix this

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/SendMessageAction.java
```diff
@@ -0,0 +1,130 @@
...
+  private final Parameters transformerParameters;
+
+  @JsonCreator
+  public SendMessageAction(
+      @JsonProperty("message") MessageDefinition message,
+      @JsonProperty("channelTarget") ChannelTarget channelTarget,
+      @JsonProperty("transformers") List<String> transformers,
+      @JsonProperty("transformerParameters") Parameters transformerParameters) {
+    this.message = message;
+    this.channelTarget = channelTarget != null ? channelTarget : OriginatingChannelTarget.INSTANCE;
```
RafeArnold, line 50: are we happy with defaulting the channel target if it's missing, or should we require an explicit target?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
tomakehurst: Done

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
```diff
@@ -0,0 +1,317 @@
...
+  }
+
+  public boolean matches(MessageChannel channel, Message message) {
+    if (trigger instanceof IncomingMessageTrigger) {
+      return ((IncomingMessageTrigger) trigger).matches(channel, message);
+    }
+    return false;
+  }
+
+  public int comparePriorityWith(MessageStubMapping other) {
```
leeturner: Was this intended to be used?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/store/MessageJournalStore.java
tomakehurst: Good catch, fixed

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/StoreBackedMessageJournal.java
tomakehurst: Removed

src/test/java/com/github/tomakehurst/wiremock/MessageTemplatingAcceptanceTest.java
tomakehurst: Not in this case :-)

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
```diff
@@ -0,0 +1,317 @@
...
+
+  public MessageTrigger getTrigger() {
+    return trigger;
+  }
+
+  @JsonIgnore
+  public RequestPattern getChannelPattern() {
+    if (trigger instanceof IncomingMessageTrigger) {
+      return ((IncomingMessageTrigger) trigger).getChannelPattern();
+    }
```
leeturner: Just style really so feel free to ignore 
```suggestion
    if (trigger instanceof IncomingMessageTrigger incoming) {
      return incoming.getChannelPattern();
    }
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
```diff
@@ -0,0 +1,317 @@
...
+    return null;
+  }
+
+  @JsonIgnore
+  public MessagePattern getMessagePattern() {
+    if (trigger instanceof IncomingMessageTrigger) {
+      return ((IncomingMessageTrigger) trigger).getMessagePattern();
+    }
```
leeturner: Just style really so feel free to ignore 
```suggestion
    if (trigger instanceof IncomingMessageTrigger incoming) {
      return incoming.getMessagePattern();
    }
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
```diff
@@ -0,0 +1,317 @@
...
+
+  @NonNull
+  public Metadata getMetadata() {
+    return metadata;
+  }
+
+  public boolean matches(MessageChannel channel, Message message) {
+    if (trigger instanceof IncomingMessageTrigger) {
+      return ((IncomingMessageTrigger) trigger).matches(channel, message);
+    }
```
leeturner: Just style really so feel free to ignore 
```suggestion
    if (trigger instanceof IncomingMessageTrigger incoming) {
      return incoming.matches(channel, message);
    }
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubRequestHandler.java
```diff
@@ -0,0 +1,254 @@
...
+      }
+    }
+    return result;
+  }
+
+  private void executeAction(
+      MessageAction action, MessageChannel originatingChannel, Message incomingMessage) {
+    if (action instanceof SendMessageAction) {
+      executeSendMessageAction((SendMessageAction) action, originatingChannel);
+    }
```
leeturner: Just style really so feel free to ignore 
```suggestion
  if (action instanceof SendMessageAction sendAction) {
    executeSendMessageAction(sendAction, originatingChannel);
  }

```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubRequestHandler.java
```diff
@@ -0,0 +1,254 @@
...
+
+  private void executeSendMessageAction(
+      SendMessageAction action, MessageChannel originatingChannel) {
+    Message message = resolveToMessage(action.getMessage());
+    ChannelTarget target = action.getChannelTarget();
+
+    if (target instanceof OriginatingChannelTarget) {
+      originatingChannel.sendMessage(message);
+    } else if (target instanceof RequestInitiatedChannelTarget) {
+      RequestInitiatedChannelTarget requestTarget = (RequestInitiatedChannelTarget) target;
```
leeturner: Just style really so feel free to ignore 
```suggestion
} else if (target instanceof RequestInitiatedChannelTarget requestTarget) {
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubRequestHandler.java
```diff
@@ -0,0 +1,254 @@
...
+      byte[] bytes =
+          resolvedData != null ? resolvedData.getBytes(StandardCharsets.UTF_8) : new byte[0];
+      InputStreamSource streamSource = () -> new ByteArrayInputStream(bytes);
+      return new Entity(
+          EncodingType.TEXT, textDef.getFormat(), textDef.getCompression(), streamSource);
+    }
+
+    throw new UnsupportedOperationException(
+        "Resolution of " + definition.getClass().getSimpleName() + " is not yet supported");
+  }
```
leeturner, line 167: Just style really so feel free to ignore 
```suggestion
private static Entity resolveEntity(EntityDefinition definition, Stores stores) {
    if (definition instanceof StringEntityDefinition stringDef) {
      String value = stringDef.getValue();
      byte[] bytes = value != null ? value.getBytes(StandardCharsets.UTF_8) : new byte[0];
      InputStreamSource streamSource = () -> new ByteArrayInputStream(bytes);
      return new Entity(EncodingType.TEXT, FormatType.TEXT, CompressionType.NONE, streamSource);
    }

    if (definition instanceof BinaryEntityDefinition binaryDef) {
      byte[] bytes = resolveBinaryEntityData(binaryDef, stores);
      InputStreamSource streamSource = () -> new ByteArrayInputStream(bytes);
      return new Entity(
          EncodingType.BINARY, FormatType.BASE64, binaryDef.getCompression(), streamSource);
    }

    if (definition instanceof TextEntityDefinition textDef) {
      String resolvedData = resolveTextEntityData(textDef, stores);
      byte[] bytes =
          resolvedData != null ? resolvedData.getBytes(StandardCharsets.UTF_8) : new byte[0];
      InputStreamSource streamSource = () -> new ByteArrayInputStream(bytes);
      return new Entity(
          EncodingType.TEXT, textDef.getFormat(), textDef.getCompression(), streamSource);
    }

    throw new UnsupportedOperationException(
        "Resolution of " + definition.getClass().getSimpleName() + " is not yet supported");
  }
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
tomakehurst: This isn't actually called anywhere at the moment, so one option would be to remove it.

I don't think it's a big issue though - the only impl of `MessageChannel.close()` checks for openness before closing making it idempotent. If we stick to that for others (making idempotency part of the contract)I think it'll be OK.

src/test/java/com/github/tomakehurst/wiremock/client/HttpAdminClientTest.java
tomakehurst: Removed this test

src/test/java/com/github/tomakehurst/wiremock/client/HttpAdminClientTest.java
tomakehurst: Removed this test

src/test/java/com/github/tomakehurst/wiremock/client/HttpAdminClientTest.java
RafeArnold: do we test that endpoint anywhere else?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
RafeArnold: the race condition is if another channel gets added between getting and closing all the channels and clearing the store. if a new channel is added between these, it will be cleared from the store without being closed.

RafeArnold: i guess removing the function if it's not actually used would make sense.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/SendMessageAction.java
tomakehurst: I think "send it back on the same channel" is a reasonable default so I'm tempted to leave this alone.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageStubMapping.java
tomakehurst: Yes, and I've now moved message stubs to use the same data structure as HTTP stubs.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageChannels.java
tomakehurst: Removed

PR #3288, opened 2026-01-08
Add wiremock-url library

Mahoney:
A type-safe, immutable URL parsing and manipulation library for Java, designed for lenient parsing
and RFC 3986 compliance.

WireMock URL provides a robust set of tiny types for working with URLs and their components. The
library balances strict RFC 3986 compliance with practical leniency, accepting all valid RFC 3986
URLs while being permissive with path, query, and fragment components.

## References

[RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986) is the strict standard for URLs (a subset of URIs)

[WHATWG URL Spec](https://url.spec.whatwg.org/) diverges from RFC 3986 and is what modern browsers mostly adhere to

## Submitter checklist

### DIFF
--- a/wiremock-url/src/main/java/org/wiremock/url/Fragment.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/Fragment.java
@@ -31,6 +31,8 @@ import org.wiremock.stringparser.ParsedString;
  */
 public interface Fragment extends PercentEncoded<Fragment>, ParsedString {
 
+  Fragment EMPTY = new FragmentValue("", true);
+
   /**
    * Returns a normalised form of this fragment.
    *
--- a/wiremock-url/src/main/java/org/wiremock/url/FragmentParser.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/FragmentParser.java
@@ -28,14 +28,22 @@ public final class FragmentParser implements PercentEncodedStringParser<Fragment
 
   @Override
   public Fragment parse(String stringForm) {
-    return new FragmentValue(stringForm);
+    if (stringForm.isEmpty()) {
+      return Fragment.EMPTY;
+    } else {
+      return new FragmentValue(stringForm);
+    }
   }
 
   static final boolean[] fragmentCharSet = combine(pcharCharSet, include('/', '?'));
 
   @Override
   public Fragment encode(String unencoded) {
-    var result = Constants.encode(unencoded, fragmentCharSet);
-    return new FragmentValue(result, true);
+    if (unencoded.isEmpty()) {
+      return Fragment.EMPTY;
+    } else {
+      var result = Constants.encode(unencoded, fragmentCharSet);
+      return new FragmentValue(result, true);
+    }
   }
 }
--- a/wiremock-url/src/main/java/org/wiremock/url/HostParser.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/HostParser.java
@@ -67,10 +67,10 @@ public final class HostParser implements PercentEncodedStringParser<Host> {
 
   @Override
   public Host encode(String unencoded) {
-    var result = Constants.encode(unencoded, hostCharSet);
     if (unencoded.isEmpty()) {
       return Host.EMPTY;
     } else {
+      var result = Constants.encode(unencoded, hostCharSet);
       return new HostValue(result, true);
     }
   }
--- a/wiremock-url/src/main/java/org/wiremock/url/HostValue.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/HostValue.java
@@ -20,19 +20,22 @@ import static org.wiremock.url.Constants.pctEncodedPattern;
 import static org.wiremock.url.Strings.transform;
 
 import java.util.Objects;
+import org.jspecify.annotations.Nullable;
 
 final class HostValue implements Host {
 
   private final String host;
-  private final boolean isNormalForm;
+  private final MemoisedNormalisable<Host> normalisable;
 
   HostValue(String host) {
-    this(host, false);
+    this(host, null);
   }
 
-  HostValue(String host, boolean isNormalForm) {
+  @SuppressWarnings("DataFlowIssue")
+  HostValue(String host, @Nullable Boolean isNormalForm) {
     this.host = host;
-    this.isNormalForm = isNormalForm;
+    this.normalisable =
+        new MemoisedNormalisable<>(this, isNormalForm, this::doIsNormalForm, this::doNormalise);
   }
 
   @Override
@@ -42,29 +45,26 @@ final class HostValue implements Host {
 
   @Override
   public Host normalise() {
+    return normalisable.normalise();
+  }
 
-    if (isNormalForm) {
-      return this;
-    }
-
+  private @Nullable Host doNormalise() {
     String normalised =
         transform(
             host,
             pctEncodedPattern,
             matched -> matched.toUpperCase(ROOT),
             unmatched -> unmatched.toLowerCase(ROOT));
-    if (normalised.equals(host)) {
-      return this;
-    } else if (normalised.isEmpty()) {
-      return Host.EMPTY;
-    } else {
-      return new HostValue(normalised, true);
-    }
+    return normalised.equals(host) ? null : new HostValue(normalised, true);
   }
 
   @Override
   public boolean isNormalForm() {
-    return isNormalForm || normalise().equals(this);
+    return normalisable.isNormalForm();
+  }
+
+  private boolean doIsNormalForm() {
+    return normalise().equals(this);
   }
 
   @Override
--- a/wiremock-url/src/main/java/org/wiremock/url/PathParser.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/PathParser.java
@@ -36,7 +36,11 @@ public final class PathParser implements PercentEncodedStringParser<Path> {
 
   @Override
   public Path parse(String stringForm) {
-    if (pathPattern.matcher(stringForm).matches()) {
+    if (stringForm.isEmpty()) {
+      return Path.EMPTY;
+    } else if (stringForm.equals("/")) {
+      return Path.ROOT;
+    } else if (pathPattern.matcher(stringForm).matches()) {
       return new PathValue(stringForm);
     } else {
       throw new IllegalPath(stringForm);
@@ -49,7 +53,13 @@ public final class PathParser implements PercentEncodedStringParser<Path> {
 
   @Override
   public Path encode(String unencoded) {
-    return new PathValue(Constants.encode(unencoded, pathCharSet), true);
+    if (unencoded.isEmpty()) {
+      return Path.EMPTY;
+    } else if (unencoded.equals("/")) {
+      return Path.ROOT;
+    } else {
+      return new PathValue(Constants.encode(unencoded, pathCharSet), true);
+    }
   }
 
   String normalisePercentEncoded(String unencoded) {
--- a/wiremock-url/src/main/java/org/wiremock/url/Query.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/Query.java
@@ -35,7 +35,7 @@ import org.wiremock.stringparser.ParsedString;
  */
 public interface Query extends PercentEncoded<Query>, QueryParamReader, ParsedString {
 
-  Query EMPTY = QueryValue.EMPTY;
+  Query EMPTY = new QueryValue("", List.of(), true);
 
   /**
    * Parses a string into a query.
--- a/wiremock-url/src/main/java/org/wiremock/url/QueryParamKey.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/QueryParamKey.java
@@ -19,6 +19,8 @@ import org.wiremock.stringparser.ParsedString;
 
 public interface QueryParamKey extends PercentEncoded<QueryParamKey>, ParsedString {
 
+  QueryParamKey EMPTY = new QueryParamKeyValue("");
+
   static QueryParamKey parse(String key) throws IllegalQueryParamKey {
     return QueryParamKeyParser.INSTANCE.parse(key);
   }
--- a/wiremock-url/src/main/java/org/wiremock/url/QueryParamKeyParser.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/QueryParamKeyParser.java
@@ -36,7 +36,9 @@ public final class QueryParamKeyParser implements PercentEncodedStringParser<Que
 
   @Override
   public QueryParamKey parse(String stringForm) {
-    if (queryParamKeyPattern.matcher(stringForm).matches()) {
+    if (stringForm.isEmpty()) {
+      return QueryParamKey.EMPTY;
+    } else if (queryParamKeyPattern.matcher(stringForm).matches()) {
       return new QueryParamKeyValue(stringForm);
     } else {
       throw new IllegalSegment(stringForm);
@@ -47,6 +49,8 @@ public final class QueryParamKeyParser implements PercentEncodedStringParser<Que
 
   @Override
   public QueryParamKey encode(String unencoded) {
-    return new QueryParamKeyValue(Constants.encode(unencoded, queryParamKeyCharSet), true);
+    return unencoded.isEmpty()
+        ? QueryParamKey.EMPTY
+        : new QueryParamKeyValue(Constants.encode(unencoded, queryParamKeyCharSet), true);
   }
 }
--- a/wiremock-url/src/main/java/org/wiremock/url/QueryParser.java
+++ b/wiremock-url/src/main/java/org/wiremock/url/QueryParser.java
@@ -46,7 +46,11 @@ public final class QueryParser implements PercentEncodedStringParser<Query> {
 
   @Override
   public Query encode(String unencoded) {
-    var result = Constants.encode(unencoded, queryCharSet);
-    return new QueryValue(result, true);
+    if (unencoded.isEmpty()) {
+      return Query.EMPTY;
+    } else {
+      var result = Constants.encode(unencoded, queryCharSet);
+      return new QueryValue(result);
+    }
   }
 }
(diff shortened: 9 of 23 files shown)


wiremock-url/src/main/java/org/wiremock/url/AuthorityValue.java
```diff
@@ -0,0 +1,148 @@
...
+import java.util.Objects;
+import java.util.Optional;
+import org.jspecify.annotations.Nullable;
+
+@SuppressWarnings({"OptionalUsedAsFieldOrParameterType", "ClassCanBeRecord"})
+final class AuthorityValue implements Authority {
+
+  private final @Nullable UserInfo userInfo;
+  private final Host host;
+  private final @Nullable Optional<Port> maybePort;
```
leeturner, line 30: Would it make sense to just have this as a nullable field and use optionals for the getters ?

wiremock-url/src/main/java/org/wiremock/url/AbsoluteUrl.java
```diff
@@ -0,0 +1,64 @@
...
+public interface AbsoluteUrl extends Url {
+
+  /**
+   * {@implSpec} Implementations must ALWAYS return null
+   *
+   * @deprecated This always returns null so you have no reason to ever call it
+   * @return null
+   */
+  @Override
+  @Deprecated(forRemoval = true) // not actually for removal, just no point ever calling
```
leeturner: I wonder if this might be confusing for users ?

wiremock-url/src/main/java/org/wiremock/url/AuthorityValue.java
Mahoney: How would that work? There are 3 possible values. See Javadoc on `Authority.getMaybePort`.

wiremock-url/src/main/java/org/wiremock/url/AbsoluteUrl.java
Mahoney: Do you mean the `@Deprecated` generally, or specifically the `forRemovel = true`?

I'm trying to find a way to get the compiler to point out when there's no point calling a method.

wiremock-url/src/main/java/org/wiremock/url/AuthorityValue.java
leeturner: Ah, sorry, my bad.

wiremock-url/src/main/java/org/wiremock/url/AbsoluteUrl.java
leeturner: I was thinking more about the `forRemovel = true`

leeturner:
This is really nice !!!

wiremock-url/src/main/java/org/wiremock/url/AbsoluteUrl.java
Mahoney: I wouldn't mind taking that away, it does feel misleading - a warning, which is what you get from `@Deprecated` without the `forRemoval` is probably enough.

Mahoney:
Point for consideration - I added some JMH benchmarks:

| Benchmark                   | Mode  | Cnt |    Score | Error    | Units  |
|-----------------------------|-------|-----|---------:|----------|--------|
| URI.create(valid)           | thrpt | 10  | 3037.588 | ± 50.716 | ops/ms |
| UriReference.parse(valid)   | thrpt | 10  |  678.386 | ±  3.216 | ops/ms |
| URI.create(invalid)         | thrpt | 10  |  832.652 | ± 30.814 | ops/ms |
| UriReference.parse(invalid) | thrpt | 10  |  385.218 | ± 17.014 | ops/ms |

Essentially `URI.create` is about 4.5 times faster than `UriReference.parse` for a valid URI string, and just over 2 times faster for an invalid one. I have not yet done any profiling to see where performance gains could be made.

wiremock-url/src/main/java/org/wiremock/url/AbsoluteUrl.java
leeturner: It is a shame there isn't a better way to provide this information other than just javadoc comments

Mahoney: I've replaced this:
`@Deprecated(forRemoval = true) // not actually for removal, just no point ever calling`
with this:
`@Deprecated // no point ever calling on this subtype`
in all cases.

wiremock-url/src/test/java/org/wiremock/url/UrlTests.java
```diff
@@ -0,0 +1,305 @@
...
+@SuppressWarnings("HttpUrlsUsage")
+class UrlTests {
+
+  @Nested
+  class ParseMethod {
+
+    @ParameterizedTest
+    @MethodSource("validUrls")
+    void parses_valid_url(UrlReferenceParseTestCase urlTest) {
+      UriReference url = UriReference.parse(urlTest.stringForm);
```
tomakehurst: Seems like this test is in the wrong class given that it's testing UriReferece rather than Url

wiremock-url/src/test/java/org/wiremock/url/AbsoluteUrlTests.java
```diff
@@ -0,0 +1,305 @@
...
+      UriReference url = UriReference.parse(urlTest.stringForm);
+      assertThat(url.isUrl()).isTrue();
+      assertThat(url.getScheme()).isEqualTo(urlTest.expectation.scheme);
+      assertThat(url.getPath()).isEqualTo(urlTest.expectation.path);
+      assertThat(url.getQuery()).isEqualTo(urlTest.expectation.query);
+      assertThat(url.getFragment()).isEqualTo(urlTest.expectation.fragment);
+    }
+
+    @Test
+    void normalise() {
```
tomakehurst: This name doesn't make it clear what this is actually testing

tomakehurst:
I find the single package structure makes it harder than necessary to determine which classes are public/significant for me as a user. Perhaps pushing implementation classes (such as `UrlParser`) into one or more sub-packages (e.g. `internal`) would help, and possibly also pure model classes, so that the classes in the top-level package represent the entry point into the library for the user.

tomakehurst:
It would be good to see some test cases for `Url` parsing and manipulation that look a bit more like real-world, end user usage, since this is probably going to be 99% of how it gets used in practice.

tomakehurst:
I'm not keen on the fact that you can't parse relative URLs from `Url.parse()` i.e. the main entry point if you're not sure what you have is `UrlReference.parse`. Although this is most aligned with the RFC terminology, I don't think it's how most people mentally model URLs. I had to look up what a URI reference was (maybe this is just me flaunting my ignorance).

Mahoney:
Possible renaming:
`UriReference` -> `Uri`
`UrlReference` -> `Url`
`Uri` -> `AbsoluteUri`
`Url` -> `AbsoluteUrl`
`RelativeRef` -> `RelativeUrl`
`AbsoluteUrl` -> `ServersideAbsoluteUrl`
Screenshot 2026-01-09 at 14 12 11

Mahoney:
> Possible renaming: `UriReference` -> `Uri` `UrlReference` -> `Url` `Uri` -> `AbsoluteUri` `Url` -> `AbsoluteUrl` `RelativeRef` -> `RelativeUrl` `AbsoluteUrl` -> `ServersideAbsoluteUrl`

Done in https://github.com/wiremock/wiremock/commit/fcfd4c38e87b977c329ca31028f57c95cb23f7b0

Mahoney:
Benchmarks:

| Benchmark                                        | Mode  | Cnt |     Score |      Error | Units |
|--------------------------------------------------|-------|----:|----------:|-----------:|-------|
| `HandlebarsOptimizedTemplateBenchmark.transform` | thrpt |   5 |   192.548 | ±    7.882 | ops/s |
| `JsonPathAdvancedMatchingBenchmark.matched`      | thrpt |   5 | 45053.881 | ± 1244.326 | ops/s |
| `JsonPathSimpleMatchingBenchmark.matched`        | thrpt |   5 | 44320.814 | ±  943.878 | ops/s |
| `PathAndMethodMatchingBenchmark.matched`         | thrpt |   5 |  6078.785 | ±  154.080 | ops/s |

After:

| Benchmark                                        | Mode  | Cnt |     Score |      Error | Units |
|--------------------------------------------------|-------|----:|----------:|-----------:|-------|
| `HandlebarsOptimizedTemplateBenchmark.transform` | thrpt |   5 |   199.869 | ±    2.793 | ops/s |
| `JsonPathAdvancedMatchingBenchmark.matched`      | thrpt |   5 | 43954.533 | ± 1429.163 | ops/s |
| `JsonPathSimpleMatchingBenchmark.matched`        | thrpt |   5 | 44457.289 | ± 1114.565 | ops/s |
| `PathAndMethodMatchingBenchmark.matched`         | thrpt |   5 |  5675.349 | ±  304.105 | ops/s |

Difference:

| Benchmark                                        |  ops/s |
|--------------------------------------------------|-------:|
| `HandlebarsOptimizedTemplateBenchmark.transform` | +3.80% |
| `JsonPathAdvancedMatchingBenchmark.matched`      | -2.44% |
| `JsonPathSimpleMatchingBenchmark.matched`        | +0.31% |
| `PathAndMethodMatchingBenchmark.matched`         | -6.64% |

All are within the margin of error of each other - that is the best case for the (on average) slower is faster than the worst case for the (on average) faster.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestLine.java
```diff
@@ -88,7 +91,7 @@ public String getPath() {
     return getPathSegments().toString();
   }
 
-  public String getUrl() {
+  public PathAndQuery getUrl() {
```
tomakehurst: This may have some unintended consequences when using other template helpers that expect a string for this value. Not 100% sure this is a problem, but worth trying to think of a test for this.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestTemplateModel.java
```diff
@@ -73,7 +73,7 @@ public Object getPath() {
   }
 
   public String getUrl() {
-    return requestLine.getUrl();
+    return requestLine.getUrl().toString();
```
tomakehurst: I guess this somewhat mitigates my previous comment

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
```diff
@@ -43,10 +45,18 @@ interface Part {
     Body getBody();
   }
 
+  @Deprecated // use getPathAndQuery()
   String getUrl();
```
tomakehurst: Could maybe add a default impl delgating to the typed version .toString() ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
```diff
@@ -43,10 +45,18 @@ interface Part {
...
 
+  @JsonIgnore
+  PathAndQuery getPathAndQuery();
+
+  @Deprecated // use getTypedAbsoluteUrl()
   String getAbsoluteUrl();
 
+  @JsonIgnore
+  AbsoluteUrl getTypedAbsoluteUrl();
```
tomakehurst: Having to name it this way is a bit of an ugly workaround. Wondering whether we should just change the type of the original method.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
```diff
@@ -332,11 +331,11 @@ private MatchResult allPathParamsMatch(final Request request) {
     if (url.getClass().equals(UrlPathTemplatePattern.class) && !pathParams.isEmpty()) {
       final UrlPathTemplatePattern urlPathTemplatePattern = (UrlPathTemplatePattern) url;
       final PathTemplate pathTemplate = urlPathTemplatePattern.getPathTemplate();
-      if (!pathTemplate.matches(request.getUrl())) {
+      if (!pathTemplate.matches(request.getPathAndQuery().toString())) {
```
tomakehurst: Maybe better to change the signature of PathTemplate to take Path or PathAndQuery directly?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/RequestPatternTransformer.java
```diff
@@ -44,24 +43,30 @@ class RequestPatternTransformer implements Function<Request, RequestPatternBuild
...
+    PathAndQuery pathAndQuery = request.getPathAndQuery();
+    var queryParameters = pathAndQuery.getQueryOrEmpty().asMap();
     // urlEqualTo is used when there are no query parameters to be as least disruptive to existing
     // behaviour as possible.
     // TODO: could be changed to always use urlPathEqualTo in next major release.
     var urlMatcher =
         queryParameters.isEmpty()
-            ? urlEqualTo(request.getUrl())
-            : urlPathEqualTo(getPath(request.getUrl()));
+            ? urlEqualTo(pathAndQuery.toString())
```
tomakehurst: Likewise, perhaps the URL matchers could be overloaded to take the new types

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
```diff
@@ -278,7 +308,7 @@ public Set<String> getAllHeaderKeys() {
 
   @Override
   public QueryParameter queryParameter(String key) {
-    return getFirstNonNull(queryParams.get(key), QueryParameter.absent(key));
+    return getQueryParameter(getPathAndQuery().getQueryOrEmpty(), key);
```
tomakehurst: Do we need to delegate to a Urls method for things like this any more?

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/WireMockHttpServletRequestAdapter.java
```diff
@@ -90,11 +90,32 @@ private String adaptUrl() {
     return withQueryStringIfPresent(url);
   }
 
+  private volatile PathAndQuery pathAndQuery = null;
```
tomakehurst: It would be more consistent to use Lazy here maybe?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/responsetemplating/RequestLine.java
Mahoney: Fair - `RequestTemplateModel` exposes `public RequestLine getRequestLine()`, though it's deprecated, so I should maintain the contract.

Mahoney: Fixed

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
Mahoney: Done

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/RequestPattern.java
Mahoney: Done

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/WireMockHttpServletRequestAdapter.java
Mahoney: Yup, done

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/recording/RequestPatternTransformer.java
Mahoney: Nice, done

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
Mahoney: It's the transformation of the URL types to the `QueryParameter` type that's still being done here.

`Request. queryParameter(String)` is only actually used in one place in the WireMock codebase (`RequestPattern.allQueryParamsMatch`), and I can inline the work there. So we could delete the method, and then we would not need the conversion helper. But not sure how bad that would be for existing clients? We use it in Mock Host a reasonable amount.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
Mahoney: I was worried about the amount of damage to existing clients, but I'm game if you are.

Raises the question of whether `getUrl()` should just return `PathAndQuery` as well.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
tomakehurst: OK, let's leave it in place then. We're probably breaking enough with this as it is.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/Request.java
tomakehurst: Yeah, I think we either do it with both or neither.

Agree this could potentially break a lot of calling code, especially since `LoggedRequest` can be used in ordinary test code. So maybe it needs to stay as you've done it.

Although is `getPathAndQuery()` necessary, since it can be got from `AbsoluteUrl`?

Mahoney: We agreed to leave it as is to save pain for clients

PR #3317, opened 2026-01-28
Use Entity, EntityDefinition and EntityResolver for HTTP stubs in addition to messages

tomakehurst:
Expand the use of the `Entity*` classes for HTTP stubbing, building towards a uniform way to represent entities/bodies and to map them directly or via stores/files.


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/entity/EntityDefinition.java
```diff
@@ -15,27 +15,479 @@
...
+    }
+    if (data != null && dataRef != null) {
+      throw new IllegalArgumentException(
+          "Cannot specify an entity with both data and data store reference");
+    }
+    if (filePath != null && dataStore != null) {
+      throw new IllegalArgumentException(
+          "Cannot specify an entity with both filePath and data store reference");
+    }
+  }
```
Mahoney: This method feels like it's telling us this shouldn't be a single class but multiple classes each of which can only contain valid combinations of parameters.

I also wonder if `filePath`, `dataStore` and `dataRef` could be something more specific than Strings?

Mahoney: My attempt at a truth table:

| data | filePath | dataStore | dataRef | Currently Valid? | Notes                                                                                                  |
|------|----------|-----------|---------|--------|--------------------------------------------------------------------------------------------------------|
| null | null     | null      | null    | Y      | Shouldn't this be illegal?                                                                             |
| null | null     | null      |         | Y      | Is it meaningful to have dataStore without dataRef?                                                    |
| null | null     |           | null    | Y      | Is it meaningful to have dataStore without dataRef?                                                    |
| null | null     |           |         | Y      | Data in a dataStore                                                                                    |
| null |          | null      | null    | Y      | Data in a file                                                                                         |
| null |          | null      |         | Y      | Is it meaningful to have dataStore without dataRef?                                                    |
| null |          |           | null    | N      |                                                                                                        |
| null |          |           |         | N      |                                                                                                        |
|      | null     | null      | null    | Y      | Simple inline data                                                                                     |
|      | null     | null      |         | N      |                                                                                                        |
|      | null     |           | null    | Y      | Shouldn't this be illegal? Why would you have both data and dataStore? Also dataStore without dataRef. |
|      | null     |           |         | N      |                                                                                                        |
|      |          | null      | null    | N      |                                                                                                        |
|      |          | null      |         | N      |                                                                                                        |
|      |          |           | null    | N      |                                                                                                        |
|      |          |           |         | N      |                                                                                                        |

tomakehurst: I agree with your comment that having assertions in the constructor seems like a smell, but I originally didn't go down the class hierarchy route because there was already a hierarchy for text/binary/json etc. and having a second "dimension" was likely to produce a combinatorial nightmare. I guess maybe it's worth revisiting, although I also find things like builders and transform methods are a pain as soon as you introduce a type hierarchy.

Per your comments about data sources, it's not reasonable to have a store name and not a key, so perhaps introducing a `DataRef(String store, String key)` type would make sense (used instead of the individual keys).

Mahoney (approved):

PR #3325, opened 2026-02-03
Entity resolver refactor

Mahoney:
Suggested refactorings


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/entity/EntityResolver.java
```diff
@@ -31,49 +26,6 @@ public EntityResolver(Stores stores) {
   }
 
   public Entity resolve(EntityDefinition definition) {
```
Mahoney: Probably not, we can just inject the stores wherever it was injected.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/entity/EntityDefinition.java
```diff
@@ -157,21 +89,35 @@ private static CompressionType tryToGuessCompressionTypeIfNotSpecified(
...
           "Cannot specify an entity with both data and data store reference");
     }
-    if (filePath != null && dataStore != null) {
+    if (filePath != null && dataStoreRef != null) {
       throw new IllegalArgumentException(
           "Cannot specify an entity with both filePath and data store reference");
     }
   }
 
+  @NonNull Entity resolve(@Nullable Stores stores) {
```
Mahoney: It uses polymorphism - the different subtypes all resolve an entity in different ways, so polymorphism seems the natural way to do it. Effectively we were doing a switch on the types otherwise.

tomakehurst: Ignore, I see it now

PR #3329, opened 2026-02-11
add fromProxy to LoggedResponse.

RafeArnold:


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/LoggedResponse.java
```diff
@@ -31,21 +31,25 @@ public class LoggedResponse {
...
 
   public LoggedResponse(
       @JsonProperty("status") int status,
       @JsonProperty("headers") HttpHeaders headers,
       @JsonProperty("bodyAsBase64") String bodyAsBase64,
       @JsonProperty("fault") Fault fault,
-      @JsonProperty("body") String ignoredBodyOnlyUsedForBinding) {
-    this(status, headers, Encoding.decodeBase64(bodyAsBase64), fault);
+      @JsonProperty("body") String ignoredBodyOnlyUsedForBinding,
+      @JsonProperty("fromProxy") boolean fromProxy) {
```
tomakehurst, line 42: Nullable `fromProxy` parameter might be better for backwards compatibility in JSON deserialisation (or does Jackson default primitive booleans when the JSON value is null/absent?)

RafeArnold: it defaults to false, if the field isn't present

PR #3330, opened 2026-02-13
Fix the jsonPath helper to write extracted arrays in JSON format

picimako:
Previously, when the `jsonPath` helper was used to extract an array, the resulting array was written as e.g.
```
[{name=One}, {name=Two}, {name=Three}]
```
intead of in JSON format:
```json
[{"name" : "One"}, {"name" : "Two"}, {"name" : "Three"}]
```

which made it inconsistent with object extraction since it produces the actual JSON object.

This behaviour of array extraction also made it necessary to apply the `toJson` helper too to receive the array in the actual JSON format.

## References

- TODO

## Submitter checklist


RafeArnold:
thanks @picimako !

PR #3344, opened 2026-02-26
Update com.networknt:json-schema-validator to 2.0.1

pedroboado:
The dependency `json-schema-validator` has seen a [major API rework after `1.5.6`](https://github.com/networknt/json-schema-validator/blob/master/doc/migration-2.0.0.md) , making keeping it up-to-date a tad more challenging. This PR addresses the required code changes to unblock the dependency update from the now discontinued branch `1.5.x` .  `json-schema-validator:2.0.1` retains Java 8 and Jackson 2 requirements.

Migrate from the removed 1.x API to the 2.x equivalents:
- Updated JsonParseTest to reflect that Error.getMessage() no longer prefixes the instance location (e.g. "$: ") to the message text.

## References
- https://github.com/wiremock/wiremock/issues/3342

## Submitter checklist

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonSchemaPattern.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonSchemaPattern.java
@@ -175,12 +175,10 @@ public class MatchesJsonSchemaPattern extends StringValuePattern {
   }
 
   private static Dialect dialectFor(SpecificationVersion version) {
-    return switch (version) {
-      case DRAFT_4 -> Dialects.getDraft4();
-      case DRAFT_6 -> Dialects.getDraft6();
-      case DRAFT_7 -> Dialects.getDraft7();
-      case DRAFT_2019_09 -> Dialects.getDraft201909();
-      case DRAFT_2020_12 -> Dialects.getDraft202012();
-    };
+    if (version == SpecificationVersion.DRAFT_4) return Dialects.getDraft4();
+    if (version == SpecificationVersion.DRAFT_6) return Dialects.getDraft6();
+    if (version == SpecificationVersion.DRAFT_7) return Dialects.getDraft7();
+    if (version == SpecificationVersion.DRAFT_2019_09) return Dialects.getDraft201909();
+    return Dialects.getDraft202012();
   }
 }
--- a/src/test/java/com/github/tomakehurst/wiremock/matching/MatchesJsonSchemaPatternTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/matching/MatchesJsonSchemaPatternTest.java
@@ -443,7 +443,7 @@ public class MatchesJsonSchemaPatternTest {
 
   @ParameterizedTest
   @EnumSource(WireMock.JsonSchemaVersion.class)
-  void nullValueMatchesWhenSchemaDeclaresNullable(WireMock.JsonSchemaVersion  jsonSchemaVersion) {
+  void nullValueMatchesWhenSchemaDeclaresNullable(WireMock.JsonSchemaVersion jsonSchemaVersion) {
     MatchesJsonSchemaPattern pattern =
         new MatchesJsonSchemaPattern(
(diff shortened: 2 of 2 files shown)


tomakehurst:
Seems this is the reason we didn't upgrade right away:
https://github.com/wiremock/wiremock/pull/3192#issuecomment-3581985765

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/matching/MatchesJsonSchemaPattern.java
```diff
@@ -48,9 +52,13 @@ public MatchesJsonSchemaPattern(
 
     SchemaRegistryConfig config = SchemaRegistryConfig.builder().typeLoose(false).build();
 
+    Dialect dialectWithNullable =
+        Dialect.builder(dialectFor(schemaVersion.toVersionFlag()))
+            .keyword(new NonValidationKeyword("nullable"))
```
RafeArnold, line 57: what is the `.keyword` method call doing exactly? the tests seem to pass without it.

pedroboado: Hi, I'm following the migration notes [here](https://raw.githubusercontent.com/networknt/json-schema-validator/master/doc/migration-2.0.0.md) 
```
`The `com.networknt.schema.SchemaValidatorsConfig` file has been replaced by either `com.networknt.schema.SchemaRegistryConfig` or `com.networknt.schema.walk.WalkConfig` or moved to `com.networknt.schema.ExecutionConfig` and can no longer be configured on a per schema basis.
```

Concretely, the migration action for this is 

```
| Name                                  | Migration                                                                               |
| ------------------------------------- | --------------------------------------------------------------------------------------- |
| `nullableKeywordEnabled`              | Removed. Dialect must contain a `nullable` keyword.                                     
```

`nullableKeywordEnabled` is the new definition of `config.setHandleNullableField(true);` [introduced with version 1.4.1](https://raw.githubusercontent.com/networknt/json-schema-validator/master/doc/upgrading.md) , never applied to wiremock. 

In version 2.0.x it seems `nullable` keyword is only supported in [OpenAPI 3.0](https://github.com/networknt/json-schema-validator/blob/2.x/src/main/java/com/networknt/schema/dialect/OpenApi30.java#L57) .  What we are doing is adding that keyword `nullable` to the _draft_ dialects in order to keep the same behaviour . 

This is tested in `MatchesJsonSchemaPatternTest.nullValueMatchesWhenSchemaDeclaresNullable()` : 
- The test *passes* on master branch , running with version `1.5.6`  of `json-schema-validator` .
- The test *fails* after migrating to 2.0.1 right after the first commit in this PR.
- Adding the `nullable` keyword to the draft dialects allows it to  *pass* again.

tomakehurst: I think there a legitimate question as to whether we actually want the nullable keyword in here, since it's technically more correct not to support it (we're doing pure JSON Schema here, not OpenAPI), and 4.x is still in beta and thus breaking changes are possible.

RafeArnold: @pedroboado sorry, you're right, the tests do fail without it. i was just looking at the wrong tests 🤦

tomakehurst: Having done a bit of data analysis I think this would be pretty nasty break, so I think we should keep this. Will merge.

tomakehurst:
@pedroboado please could you run `./gradlew spotlessApply` and push the results?

pedroboado:
> @pedroboado please could you run `./gradlew spotlessApply` and push the results?

Done, apologies I missed running it with the last commit.

tomakehurst:
There's an architecture test failing now. Could you take a quick look at that and push a fix?

tomakehurst:
Thanks @pedroboado !

PR #3351, opened 2026-03-04
add batch methods to StubLifecycleListener.

RafeArnold:
add batch methods to StubLifecycleListener so that multiple stubs can be processed at once if imported in a batch.

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
@@ -252,20 +252,20 @@ public abstract class AbstractStubMappings implements StubMappings {
       listener.beforeStubsAltered(toAlterStubs);
     }
 
-    for (int i = 0; i < toAlterStubs.size(); i++) {
-      StubMappingToAlter alter = toAlterStubs.get(i);
-      if (alter instanceof StubLifecycleListener.StubMappingToCreate create) {
+    List<StubMapping> result = new ArrayList<>(toInsert.size());
+    for (StubMappingToAlter alter : toAlterStubs) {
+      if (alter instanceof StubMappingToCreate create) {
         create.setStub(store.add(create.getStub()));
         scenarios.onStubMappingAdded(create.getStub());
-        toInsert.set(i, create.getStub());
-      } else if (alter instanceof StubLifecycleListener.StubMappingToEdit edit) {
+        result.add(create.getStub());
+      } else if (alter instanceof StubMappingToEdit edit) {
         edit.setNewStub(
             edit.getNewStub()
                 .transform(b -> b.setInsertionIndex(edit.getOldStub().getInsertionIndex())));
         store.replace(edit.getOldStub(), edit.getNewStub());
         scenarios.onStubMappingUpdated(edit.getOldStub(), edit.getNewStub());
-        toInsert.set(i, edit.getNewStub());
-      } else if (alter instanceof StubLifecycleListener.StubMappingToRemove remove) {
+        result.add(edit.getNewStub());
+      } else if (alter instanceof StubMappingToRemove remove) {
         store.remove(remove.getStub().getId());
         scenarios.onStubMappingRemoved(remove.getStub());
       }
@@ -276,7 +276,7 @@ public abstract class AbstractStubMappings implements StubMappings {
           toAlterStubs.stream().map(toAlter -> (AlteredStubMapping) toAlter).toList());
     }
 
-    return toInsert;
+    return result;
   }
 
   @Override


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
```diff
@@ -38,6 +43,8 @@
 import com.github.tomakehurst.wiremock.store.files.BlobStoreFileSource;
 import com.github.tomakehurst.wiremock.verification.LoggedRequest;
 import java.util.*;
+import java.util.stream.Stream;
+import org.jspecify.annotations.NullMarked;
 
 public abstract class AbstractStubMappings implements StubMappings {
```
leeturner, line 49: Should this class also be `@NullMarked` ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StubLifecycleListener.java
```diff
@@ -38,4 +41,58 @@ default void afterStubRemoved(StubMapping stub) {}
   default void beforeStubsReset() {}
 
   default void afterStubsReset() {}
+
+  default void beforeStubsAltered(List<StubMappingToAlter> stubs) {
```
leeturner, line 45: Do we expect multiple listeners to all see the original list here or listeners later in the chain to see the updates made by listeners earlier in the chain ?

If all listeners are given the same list and we are mutating internal state of `StubMappingToCreate` or `StubMappingToEdit`, won't listeners later in the chain see those updates instead of the original ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
```diff
@@ -226,6 +233,52 @@ public StubMapping editMapping(StubMapping stubMapping) {
     return stubMapping;
   }
 
+  @Override
+  public List<StubMapping> updateMappings(List<StubMapping> toInsert, List<StubMapping> toRemove) {
```
leeturner, line 237: This method updates `toInsert` in place and then aslo returns it.  Is the in place updating of the list relied on by callers of this method and the return value relied on in others?  Was wondering if it would be a good idea to standardise on one.  Maybe not update in place but build a separate list to return?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
```diff
@@ -226,6 +233,52 @@ public StubMapping editMapping(StubMapping stubMapping) {
...
+                              : new CreateStubMapping(stub);
+                        }),
+                toRemove.stream().<StubMappingToAlter>map(RemoveStubMapping::new))
+            .toList();
+
+    for (StubLifecycleListener listener : stubLifecycleListeners) {
+      listener.beforeStubsAltered(toAlterStubs);
+    }
+
+    for (int i = 0; i < toAlterStubs.size(); i++) {
```
leeturner: In this block, `i` directly relates to a position in the `toInsert` list which as mentioned above, is updated in place.  I guess this works because of the way `toAlterStubs` is created and the order is consistent.

Would it be better to decouple this link on indexes ?  Not editing in place and building a new list to return would help here maybe?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/extension/StubLifecycleListener.java
RafeArnold: it is meant to replicate how the individual stub methods work. if beforeStubEdited/beforeStubCreated returns a new stub, that stub will be passed to the next listener. so this batch one will do the same. i.e. mutating state of the stub to alter within one listener will be reflected in the next listeners

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
RafeArnold: yeah that would be best practice. the caller shouldnt rely on the passed in list, only the return value. the reason i kept them as the same list was purely for convenience, to save having to generate a new list unnecessarily. we control both ends of the call to this method, so there's no risk within the current code, but if you'd prefer cleaner practice i can copy the list.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
RafeArnold: yeah im fully relying on the order that toAlterStubs is created. i felt this was safe because we control what happens inside the function but i can change it if you think more defensive code is better.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
leeturner: I am sure in the years to come I will forget this conversation and have to go and figure out how it works and which list is being updated where.  If it isn't a major pain, might be nice to clarify that now

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
leeturner: Should be fine for now

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
RafeArnold: done

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
RafeArnold: fixed by resolving the other comment anyway

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
```diff
@@ -271,4 +324,112 @@ public List<StubMapping> findByMetadata(final StringValuePattern pattern) {
...
+    }
+
+    @Override
+    public StubMapping getNewStub() {
+      return newStub;
+    }
+
+    @Override
+    public void setNewStub(StubMapping newStub) {
+      this.newStub = newStub;
```
tomakehurst, line 385: Making these mutable seems a bit grubby. Is there no practical immutable alternative?

RafeArnold: as far as i can see this is the only reasonable way to replicate the single stub methods, `beforeStubCreated` and `beforeStubEdited`, that can return a new stub value to be passed to the next listener/saved. initially i tried to use the return value of `beforeStubsAltered` in the same way but because its a list of stubs it creates huge potential for misaligned behaviour. the implementor of the listener could completely remove a bunch of stubs from the returned list so they wouldnt be processed. this is not an ability that the listeners have had so far so didnt feel appropriate to add here. i kind of agree that on first sight, the mutability looks a bit nasty, but i actually think it creates quite a clean api for the listener implementors that replicates the api of the single stub methods. to be clear, this mutability only lasts for the duration of the call to this method, so doesnt create the potential for mutable references to hang around and edit stubs at will.

PR #3379, opened 2026-03-31
fix: Improve consistency of stub modification

Mahoney:
There's a classic two phase commit problem here. For all stub
modifications, WireMockApp first mutates StubMappings, then calls
MappingsSaver to persist the change. If the call to MappingsSaver
fails to persist the change then what is running, and what is persisted,
are out of sync until the persisted changes reload.

There is of course no 100% robust solution to committing transactions on
two independent data stores. However, catching any exception in the
calls to MappingsSaver and reverting the in-memory changes to
StubMappings seems superior to not doing so.

The fix moves persistence into AbstractStubMappings itself. Each
mutation method now follows the pattern: run before-listeners, update
the store, attempt to save via protected save()/remove() hooks,
revert the store silently on failure, and only then run after-listeners.
StoreBackedStubMappings overrides the hooks to delegate to a
MappingsSaver. This means after-listeners which trigger external
side effects  only fire once the save has succeeded, and rollback
is a direct store revert with no listener involvement.

WireMockApp.addStubMapping(), editStubMapping(), and
removeStubMapping() are now simple delegations to StubMappings.

It is of course possible that the changes were, in fact, persisted
despite the exception being thrown; in which case the running state and
the persisted state will still be out of sync. I believe there is no
solution to this - it could be further mitigated by rechecking, but not
ultimately solved.

Breaking change: third-party implementations of StubMappings that
extend AbstractStubMappings will inherit the new behaviour
automatically. **Direct implementors of the StubMappings interface will
lose the save behaviour that WireMockApp previously provided**, and will
need to implement it themselves if required.

## Submitter checklist

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
@@ -263,20 +263,20 @@ public class StoreBackedStubMappings implements StubMappings {
     mappingsSaver.save(stubMapping);
   }
 
-  private void save(List<StubMapping> stubMappings) {
-    mappingsSaver.save(stubMappings);
-  }
-
   private void remove(UUID stubMappingId) {
     mappingsSaver.remove(stubMappingId);
   }
 
-  private void remove(List<UUID> stubMappingIds) {
-    mappingsSaver.remove(stubMappingIds);
-  }
-
   private void removeAllPersisted() {
     mappingsSaver.removeAll();
+  }
+
+  private void mutate(List<StubMapping> toSave, List<StubMapping> toRemove) {
+    List<UUID> ids =
+        toRemove.stream().filter(StubMapping::shouldBePersisted).map(StubMapping::getId).toList();
+    if (!toSave.isEmpty() || !ids.isEmpty()) {
+      mappingsSaver.mutate(toSave, ids);
+    }
   }
 
   @Override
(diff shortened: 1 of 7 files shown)


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
```diff
@@ -182,6 +182,16 @@ public StubMapping addMapping(StubMapping mapping) {
     mapping = store.add(mapping);
     scenarios.onStubMappingAdded(mapping);
 
+    if (mapping.shouldBePersisted()) {
+      try {
+        save(mapping);
+      } catch (Exception e) {
+        store.remove(mapping.getId());
+        scenarios.onStubMappingRemoved(mapping);
```
RafeArnold: would it be simpler to execute the scenario operations after the catch blocks? then you dont have to call the inverse operation in the catch block?

Mahoney: I should hold off looking at this too hard yet, it's more a POC - it's the architectural breaking change that's most interesting to comment on

Mahoney:
Should have marked this as a draft to start with...

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/AbstractStubMappings.java
RafeArnold: makes sense. pushing persistence down into the stub mappings implementation is a great change imo

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
```diff
@@ -224,16 +244,31 @@ public WireMockApp(
             List.copyOf(extensions.ofType(MessageActionTransformer.class).values()));
     serveEventListeners = Map.of(httpStubListener.getName(), httpStubListener);
 
+    BlobStore filesBlobStore = stores.getFilesBlobStore();
+    StubMappingStore stubStore = stores.getStubStore();
+
+    nonSavingStubMappings =
+        new SavingStubMappings(
```
leeturner: Should this be `NonSavingStubMappings` ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
leeturner, line 378: From what I can tell, this has changed the way mappings are saved.  Before it was a single batched operation but now it calls `editMapping` for each stub.  I guess this means everything is then done individually rather than an atomic operation - before/after listener, stub persistence.

What happens if stub persistence fails part way through - say the 10th stub of 20?  The first 9 are persisted and all the listeners fired but the remaining 10 are not so we have a partial save and not full batch rollback ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
```diff
@@ -0,0 +1,277 @@
...
+    for (StubLifecycleListener listener : stubLifecycleListeners) {
+      listener.afterStubsAltered(
+          toAlterStubs.stream().map(toAlter -> (AlteredStubMapping) toAlter).toList());
+    }
+
+    return result;
+  }
+
+  @Override
+  public void resetMappings() {
```
leeturner, line 372: Is there a 2 phase commit issue here if one of these methods fails ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
Mahoney: I've reverted this, but probably

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
Mahoney: Yes, but they are in the right order - the first call is the one that might fail

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/core/WireMockApp.java
Mahoney: Good catch. I've added a single method to `MappingsSaver`, `mutate`, so that we can add, update and remove in an atomic operation if desired. I've also rejigged the whole `StoreBackedStubMappings.updateMappings` method to follow the same patterns as the other persist-then-update-in-memory-state methods.

PR #3383, opened 2026-04-10
Remove need for breaking change MappingsSaver.mutate

Mahoney:
It wasn't necessary.

## Submitter checklist

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
@@ -248,7 +248,7 @@ public class StoreBackedStubMappings implements StubMappings {
 
     afterStubsAltered(toAlterStubs);
 
-    return List.of();
+    return toRemove;
   }
 
   @Override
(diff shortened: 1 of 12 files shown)


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
```diff
@@ -228,6 +232,25 @@ public StubMapping removeMapping(StubMapping mapping) {
...
+    remove(toRemove.stream().filter(StubMapping::shouldBePersisted).toList());
+
+    for (StubMapping remove : toRemove) {
+      store.remove(remove.getId());
+      scenarios.onStubMappingRemoved(remove);
+    }
+
+    afterStubsAltered(toAlterStubs);
+
+    return List.of();
```
leeturner: Just wondering if this should return the removed mappings.  I guess the return value isn't used so not a blocker

leeturner (approved):
Like this much better.  One minor question but not a blocker by any means. Nice work

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
Mahoney: Fair - fixed

PR #3403, opened 2026-04-29
Recorder serve event transformer

tomakehurst:
Supports transformation of serve events before they are processed by the recorder, providing another way to customise recorded output.

Intended for situations like where the the request and response body are in a format like protobuf and must be converted to JSON for use within WireMock.

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
@@ -418,9 +418,125 @@ public class LoggedRequest implements Request {
       this.formParameters = original.formParameters;
     }
 
+    public UUID getId() {
+      return id;
+    }
+
+    public Builder withId(UUID id) {
+      this.id = id;
+      return this;
+    }
+
+    public String getScheme() {
+      return scheme;
+    }
+
+    public Builder withScheme(String scheme) {
+      this.scheme = scheme;
+      return this;
+    }
+
+    public String getHost() {
+      return host;
+    }
+
+    public Builder withHost(String host) {
+      this.host = host;
+      return this;
+    }
+
+    public Integer getPort() {
+      return port;
+    }
+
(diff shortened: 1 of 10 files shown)


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
```diff
@@ -368,4 +368,95 @@ public Part getPart(final String name) {
...
+      this.clientIp = original.clientIp;
+      this.headers = original.headers;
+      this.pathParams = original.pathParams;
+      this.cookies = original.cookies;
+      this.isBrowserProxyRequest = original.isBrowserProxyRequest;
+      this.loggedDate = original.loggedDate;
+      this.body = original.body;
+      this.multiparts = original.multiparts;
+      this.protocol = original.protocol;
+      this.formParameters = original.formParameters;
```
RafeArnold, line 419: we need to copy any elements of collections into new ones here. see StubMapping.Builder or RequestPattern.Builder for examples.

RafeArnold: we're also missing a bunch of methods on the builder for mutating these fields

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
RafeArnold, line 400: aren't a bunch of these fields derived from others field (e.g. scheme, host, post, pathAndQuery derived from absoluteUrl. cookies derived from headers. formParameters derived from body)? are we okay with allowing them to not correspond at all?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/verification/LoggedRequest.java
RafeArnold: there's still a bunch of collections we're not copying (e.g. headers)

PR #3457, opened 2026-06-05
Switch to tag-based releases with version derived from tag

tomakehurst:
Releases are now triggered by pushing a semver tag (e.g. 4.0.1) rather than via workflow_dispatch. The version is no longer committed to source — all projects default to 0.0.0-dev, and the release workflow passes -PreleaseVersion=<tag> to Gradle to stamp the build and all version files (version.properties, swagger specs, package.json) transiently via the new apply-release-version task.

The set-snapshot-version task now derives its base version from the latest git tag when the committed version is the 0.0.0-dev placeholder, keeping snapshot publishing working without manual version bumps.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -207,7 +207,6 @@ publishing {
 
 val checkReleasePreconditions by tasks.registering  {
   doLast {
-    if (System.getenv("CI") == "true") return@doLast
     val releaseBranches = listOf("master", "v4.x")
     val currentGitBranch = providers.exec {
       commandLine("git", "rev-parse", "--abbrev-ref", "HEAD")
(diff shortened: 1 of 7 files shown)


build.gradle.kts
```diff
@@ -207,6 +207,7 @@ publishing {
 
 val checkReleasePreconditions by tasks.registering  {
   doLast {
+    if (System.getenv("CI") == "true") return@doLast
```
Mahoney: Not sure I understand this? Why would we want to skip this check on CI?

If anything we should require CI _and_ a release branch to do a release, shouldn't we?

Mahoney (commented):
One query, otherwise looks good

oleg-nenashev (commented):
I agree with @Mahoney . Otherwise, this PR looks ready to go.
I will probably synchronize the WireMock modules to follow the same approach later

PR #3488, opened 2026-06-28
Use configured notifier in request handlers

### LINKED ISSUE #3285: Notifier doesn't take effect when using a custom HttpServerFactory
### Proposal

Make notifier setup part of `AbstractRequestHandler`, that is: instead calling `LocalNotifier.notifier()` to log:
https://github.com/wiremock/wiremock/blob/73557184738f7efa2d9630b0bd9dbc4ba168d1b0/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandler.java#L81-L90

inject a `Notifier` to the handlers' constructor. These handlers should come with logging wired correctly already:

https://github.com/wiremock/wiremock/blob/73557184738f7efa2d9630b0bd9dbc4ba168d1b0/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/http/HttpServerFactory.java#L29-L33

This would remove the need for the `HttpServer` implementations to set global thread-local variables.

### Reproduction steps

```kotlin
    @get:Rule(order = 0)
    val wiremockRule = WireMockRule(
        wireMockConfig()
            .stubRequestLoggingDisabled(false)
            .notifier(ConsoleNotifier(true))
    )
```
\+ a custom `com.github.tomakehurst.wiremock.http.HttpServerFactory`

Because we're not using the Jetty server, the logging simply doesn't work, it only outputs:
```
2025-12-29 12:50:33.232 Using HTTP server impl: MockWebServerHttpServer
```

### References

Version 4.0.0-beta.20

For Jetty, this line takes care of setting up the logging:
https://github.com/wiremock/wiremock/blob/73557184738f7efa2d9630b0bd9dbc4ba168d1b0/wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/WireMockHandlerDispatchingServlet.java#L136

DragonFSKY:
Fixes #3285.

This wires the configured `Notifier` into the admin and stub request handlers created by `WireMockApp`, so non-Jetty/custom `HttpServerFactory` implementations receive request logging without needing to set `LocalNotifier` themselves.

The existing handler constructors are preserved and continue to fall back to `LocalNotifier`, and the `HttpServerFactory` API is unchanged.

DirectCall HTTP server integration tests cover both stub and admin request logging through the custom-server path.

Tested with:

```
./gradlew :test --tests com.github.tomakehurst.wiremock.direct.DirectCallHttpServerIntegrationTest --tests com.github.tomakehurst.wiremock.stubbing.AdminRequestHandlerTest --tests com.github.tomakehurst.wiremock.http.AdminRequestHandlerTest
./gradlew spotlessCheck
```

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/direct/DirectCallHttpServerIntegrationTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/direct/DirectCallHttpServerIntegrationTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2021-2022 Thomas Akehurst
+ * Copyright (C) 2021-2026 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


tomakehurst:
Closing this as it doesn't align with our future architectural direction.

The intention is to remove dependence on notifiers entirely for request logging and instead work entirely with serve events and sub-events.

@DragonFSKY thanks for taking the time to contribute. I suggest next time opening an issue first and possibly dropping us a Slack message so we can discuss options a bit before writing code.

PR #3504, opened 2026-07-14
Add missing scenario state path to admin OpenAPI spec

### LINKED ISSUE #3319: OpenAPI spec for admin API is missing the set scenario state path
### Summary

The OpenAPI spec does not include the path /__admin/scenarios/{name}/state which supports PUT requests to set the state of a scenario. I would appreciate if it could be added. 
This concerns the documentation on the wiremock website as well as the one provided by the running application.

The following change should take care of this (based on master):
[WIremock_AdminAPI_OpenAPI_fix.patch](https://github.com/user-attachments/files/24919782/WIremock_AdminAPI_OpenAPI_fix.patch)

### References

_No response_

Chandu0501:
Closes #3319.

## Summary
- add the missing `/__admin/scenarios/{name}/state` PUT path to the admin OpenAPI spec
- keep the JSON and YAML spec resources aligned for the embedded docs endpoint
- extend `AdminApiTest` so the served swagger document covers this route

## Verification
- `./gradlew.bat :test --tests com.github.tomakehurst.wiremock.AdminApiTest`


tomakehurst:
Any reason this still needs to be in draft?

PR #3506, opened 2026-07-16
Fix scenario stub becoming unmatched under concurrent load

Mahoney:
## Problem

Under concurrent load on a single scenario, a small fraction of requests are served as **unmatched (404)** even though a stub exists for every scenario state.

`StoreBackedStubMappings.serveFor` filters candidate stubs by re-reading the scenario's current state **per candidate** (`mappingMatchesScenarioState`). There is no consistent snapshot across the scan, so if another request advances the scenario mid-scan to a state whose stub was already passed in the iteration — e.g. a loop wrapping `DELIVERED → Started` — no candidate matches and the request falls through to `NOT_CONFIGURED`.

Observed at ~0.45% of requests against a single wrapping 4-state scenario at ~1000 req/s.

## Fix

Take one immutable `ScenarioSnapshot` of scenario state per request and match against it, so state changes during a scan can no longer strand a request. `Scenarios.mappingMatchesScenarioState` moves onto the snapshot.

## Test

`ScenarioConcurrencyTest` hammers one wrapping scenario from 8 threads and asserts zero unmatched responses — fails before the change, passes after.

## Note

This addresses the *matching* race only. The state *transition* in `onStubServed` remains a non-atomic get-check-put (can drop transitions under concurrency), which is a separate issue and not a 404 cause.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
@@ -38,8 +38,5 @@ public interface Scenarios {
 
   void clear();
 
-  @SuppressWarnings("unused")
-  boolean mappingMatchesScenarioState(StubMapping mapping);
-
   ScenarioSnapshot snapshot();
 }
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2022-2025 Thomas Akehurst
+ * Copyright (C) 2022-2026 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,5 +38,8 @@ public interface Scenarios {
 
   void clear();
 
+  @SuppressWarnings("unused")
   boolean mappingMatchesScenarioState(StubMapping mapping);
+
+  ScenarioSnapshot snapshot();
 }
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
@@ -95,13 +95,17 @@ public class StoreBackedStubMappings implements StubMappings {
 
     final List<SubEvent> subEvents = new ArrayList<>();
 
+    // Match against one point-in-time snapshot of scenario state so that a concurrent request
+    // advancing a scenario mid-scan cannot cause every candidate to be skipped (which would
+    // otherwise leave the request unmatched even though a stub exists for every state).
+    final ScenarioSnapshot scenarioSnapshot = scenarios.snapshot();
     StubMapping matchingStub =
         store
             .findAllMatchingRequest(request, customMatchers, subEvents::add)
             .filter(
                 stubMapping ->
                     stubMapping.isIndependentOfScenarioState()
-                        || scenarios.mappingMatchesScenarioState(stubMapping))
+                        || scenarioSnapshot.mappingMatchesScenarioState(stubMapping))
             .findFirst()
             .orElse(StubMapping.NOT_CONFIGURED);
 
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
@@ -15,6 +15,7 @@
  */
 package com.github.tomakehurst.wiremock.stubbing;
 
+import static com.github.tomakehurst.wiremock.common.Lazy.lazy;
 import static com.github.tomakehurst.wiremock.common.LocalNotifier.notifier;
 import static com.github.tomakehurst.wiremock.common.Pair.pair;
 import static com.github.tomakehurst.wiremock.extension.ServeEventListener.RequestPhase.AFTER_MATCH;
@@ -28,6 +29,7 @@ import com.github.tomakehurst.wiremock.common.Errors;
 import com.github.tomakehurst.wiremock.common.FileSource;
 import com.github.tomakehurst.wiremock.common.InvalidInputException;
 import com.github.tomakehurst.wiremock.common.Json;
+import com.github.tomakehurst.wiremock.common.Lazy;
 import com.github.tomakehurst.wiremock.common.Pair;
 import com.github.tomakehurst.wiremock.core.MappingsSaver;
 import com.github.tomakehurst.wiremock.extension.*;
@@ -97,15 +99,17 @@ public class StoreBackedStubMappings implements StubMappings {
 
     // Match against one point-in-time snapshot of scenario state so that a concurrent request
     // advancing a scenario mid-scan cannot cause every candidate to be skipped (which would
-    // otherwise leave the request unmatched even though a stub exists for every state).
-    final ScenarioSnapshot scenarioSnapshot = scenarios.snapshot();
+    // otherwise leave the request unmatched even though a stub exists for every state). Taken
+    // lazily, on the first candidate that is constrained by scenario state, so that requests
+    // matching no scenario never pay for it.
+    final Lazy<ScenarioSnapshot> scenarioSnapshot = lazy(scenarios::snapshot);
     StubMapping matchingStub =
         store
             .findAllMatchingRequest(request, customMatchers, subEvents::add)
             .filter(
                 stubMapping ->
                     stubMapping.isIndependentOfScenarioState()
-                        || scenarioSnapshot.mappingMatchesScenarioState(stubMapping))
+                        || scenarioSnapshot.get().mappingMatchesScenarioState(stubMapping))
             .findFirst()
             .orElse(StubMapping.NOT_CONFIGURED);
 
--- a/src/test/java/com/github/tomakehurst/wiremock/common/LazyTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/common/LazyTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (C) 2023 Thomas Akehurst
+ * Copyright (C) 2023-2026 Thomas Akehurst
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -18,6 +18,12 @@ package com.github.tomakehurst.wiremock.common;
 import static com.github.tomakehurst.wiremock.common.Lazy.lazy;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.Future;
 import java.util.concurrent.atomic.AtomicInteger;
 import org.junit.jupiter.api.Test;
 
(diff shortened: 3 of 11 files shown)


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
```diff
@@ -95,13 +95,17 @@ public ServeEvent serveFor(ServeEvent initialServeEvent) {
 
     final List<SubEvent> subEvents = new ArrayList<>();
 
+    // Match against one point-in-time snapshot of scenario state so that a concurrent request
+    // advancing a scenario mid-scan cannot cause every candidate to be skipped (which would
+    // otherwise leave the request unmatched even though a stub exists for every state).
+    final ScenarioSnapshot scenarioSnapshot = scenarios.snapshot();
```
leeturner: If I understand this correctly, snapshot() copies every scenario into a new HashMap on every request eagerly - before findAllMatchingRequest, so it does this regardless of whether scenarios are relevant to the stub.

The old code did a get only for scenario candidates, and no work when none existed. Are we confident this isn't going to be a performance issue for people who use loads and loads of scenarios ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
```diff
@@ -38,5 +38,8 @@ public interface Scenarios {
 
   void clear();
 
+  @SuppressWarnings("unused")
```
leeturner: Are we keeping this because it is part of the public API ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
```diff
@@ -38,5 +38,8 @@ public interface Scenarios {
...
   boolean mappingMatchesScenarioState(StubMapping mapping);
+
+  ScenarioSnapshot snapshot();
```
leeturner, line 41: Is it worth making this a default method so we don't break anyone implementing this ?

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
Mahoney: Removed in latest commit as Scenarios is not part of the public API.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/Scenarios.java
Mahoney: Scenarios is not part of the public API, so no need I think.

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/stubbing/StoreBackedStubMappings.java
Mahoney: I would anticipate it isn't an issue as everything is in memory. The work is just:
```java
ConcurrentHashMap<String, Scenario> scenarioMap = new ConcurrentHashMap<>();
Stream<Scenario> all = scenarioMap.values().stream();
all.collect(toMap(Scenario::getId, scenario -> scenario))
```
which I would anticipate being very cheap even if there were, say, 500 scenarios.

In the old code, an API with 500 scenarios would have to retrieve individual scenario state multiple times as it iterated over every stub, as presumably many many stubs would match based on scenario.

So I would expect this would improve performance in all but the most unusual cases.

I have proven none of this.

Mahoney:
## Performance impact of the scenario-state snapshot

This PR replaces a lazy, per-candidate scenario state lookup (`scenarios.mappingMatchesScenarioState(mapping)`, one O(1) map get per candidate) with a single immutable snapshot of **all** scenario state, taken unconditionally at the top of `StoreBackedStubMappings.serveFor`. That fixes requests being served as unmatched (404) when a concurrent request advances a scenario mid-scan, but it changes the shape of the per-request cost:

- **Before:** O(k) map lookups, where k is the number of *scenario-constrained* candidate stubs actually examined before a match. Stubs outside any scenario short-circuit on `isIndependentOfScenarioState()`, so k is 0 for a server with no scenarios and small whenever the match is found early.
- **After:** O(S) on every request, where S is the total number of scenarios — paid whether or not the request has anything to do with a scenario.

Here are the numbers to judge that trade.

### Method

`ScenarioMatchingBenchmark` (added in this PR, `wiremock-core/src/jmh/`) drives `StubMappings.serveFor` on an `InMemoryStubMappings`. It is written against that one entry point only, so the identical source compiles and runs on `master`, making a like-for-like A/B possible. 3 forks × 8 × 1s measurement iterations, at 1 and 4 threads.

| Fixture | Scenarios | Stubs |
|---------|-----------|-------|
| `NONE`  | 0         | 601   |
| `FEW`   | 2         | 11    |
| `MANY`  | 200       | 601   |

`NONE` deliberately carries the same stub count as `MANY`, so comparing the two isolates the cost of scenario count from the cost of stub count.

```
./gradlew :wiremock-core:jmhRun --args="ScenarioMatchingBenchmark -t 4"
```

### Results (µs/op, average time, lower is better)

| Benchmark | Fixture | master 1t | this PR 1t | master 4t | this PR 4t |
|--------------------------------|---------|-----------|-----------|-----------|-----------|
| `matchScenarioIndependentStub` | `NONE`  | 3.64 | 3.66 (+0.6%) | 4.82 | 5.03 (+4%) |
| `matchScenarioIndependentStub` | `FEW`   | 3.59 | 3.73 (+4%)   | 4.75 | 4.84 (+2%) |
| `matchScenarioIndependentStub` | `MANY`  | 3.61 | **8.51 (+136%)** | 4.96 | **10.01 (+102%)** |
| `matchScenarioStub`            | `FEW`   | 7.08 | 7.33 (+4%)   | 10.56 | 8.86 (−16%) |
| `matchScenarioStub`            | `MANY`  | 6.08 | **11.15 (+83%)** | 19.38 | **13.04 (−33%)** |

### What this means

**1. With no scenarios, and with few scenarios, the change is free.** The 0-scenario delta is +0.02 µs single-threaded, and the +0.21 µs at 4 threads is inside the ±0.21 error bar. At 2 scenarios it is ~0.1–0.25 µs. That covers the large majority of users.

**2. The cost is linear in scenario count, and every request pays it** — roughly **25 ns per scenario**, so ~5 µs at 200 scenarios. Note this is paid even by requests matching a stub that has nothing to do with any scenario: that is exactly what the `matchScenarioIndependentStub` row measures (its stub is matched on the first candidate), and `snapshot()` runs before matching begins. Framing matters here — ~5 µs roughly doubles in-memory matching cost, but is small against a real HTTP request measured in hundreds of µs.

**3. The 4-thread rows where this PR appears *faster* are not a design win — they are the bug being fixed.** On `master`, contention leaves a few percent of requests unmatched, and an unmatched request is scanned against every stub. Running this PR's `ScenarioConcurrencyTest` against `master` gives **5223 / 40000 unmatched** (13%, 8 threads), and a third benchmark (`failToMatchAnyStub`) prices that full-scan path at **~300 µs** with 600 stubs, versus ~6 µs when matched. So `master`'s 19.38 µs is an average inflated by a catastrophic tail. Read positively: this PR removes a ~300 µs tail latency in exchange for a flat ~25 ns per scenario.

### Suggested read

Acceptable as-is for typical deployments. If hundreds of scenarios is a supported target, the ~5 µs is avoidable without giving up the fix — either build the snapshot lazily (only once a scenario-constrained stub actually becomes a candidate), or maintain an immutable state map updated on scenario transition rather than rebuilding it per request. Happy to do either if reviewers would prefer.

**Caveats:** in-memory stores only, no HTTP layer. `serveFor` carries ~3.6 µs of fixed overhead (request ID decoration and similar), so the percentages understate the change as a fraction of matching work and overstate it as a fraction of a served request.

Mahoney:
## Follow-up: taking the snapshot lazily removes the regression

Acting on the numbers in the comment above, `serveFor` now defers `scenarios.snapshot()` to the first candidate stub that is actually constrained by scenario state, via `Lazy`:

```java
final Lazy<ScenarioSnapshot> scenarioSnapshot = lazy(scenarios::snapshot);
...
.filter(stubMapping ->
    stubMapping.isIndependentOfScenarioState()
        || scenarioSnapshot.get().mappingMatchesScenarioState(stubMapping))
```

The property the fix relies on is unaffected — the snapshot is still taken at most once per request, so every state comparison within one scan still reads the same immutable state. `ScenarioConcurrencyTest` still passes (0 unmatched out of 40 000).

Same benchmark and same machine as before, re-run in one session. **master** is the baseline, **eager** is the snapshot taken up front, **lazy** is the current state of the PR.

### 1 thread (µs/op)

| Benchmark | Fixture | master | eager | lazy |
|--------------------------------|---------|---------------|---------------|---------------|
| `matchScenarioIndependentStub` | `NONE`  | 3.76 ±0.10 | 3.66 ±0.03 | 3.64 ±0.04 |
| `matchScenarioIndependentStub` | `FEW`   | 3.67 ±0.03 | 3.73 ±0.03 | 3.64 ±0.09 |
| `matchScenarioIndependentStub` | `MANY`  | 3.74 ±0.04 | **8.51 ±0.08** | **3.83 ±0.32** |
| `matchScenarioStub`            | `FEW`   | 7.31 ±0.08 | 7.33 ±0.09 | 7.31 ±0.08 |
| `matchScenarioStub`            | `MANY`  | 6.04 ±0.07 | 11.15 ±0.08 | 11.37 ±0.95 |
| `failToMatchAnyStub`           | `NONE`  | 312.9 ±6.3 | 296.7 ±3.5 | 301.1 ±7.2 |
| `failToMatchAnyStub`           | `FEW`   | 7.97 ±0.10 | 8.05 ±0.34 | 7.76 ±0.18 |
| `failToMatchAnyStub`           | `MANY`  | 315.5 ±3.8 | 312.5 ±3.8 | 315.0 ±8.1 |

### 4 threads (µs/op)

| Benchmark | Fixture | master | eager | lazy |
|--------------------------------|---------|---------------|---------------|---------------|
| `matchScenarioIndependentStub` | `NONE`  | 4.47 ±0.03 | 5.03 ±0.19 | 4.63 ±0.24 |
| `matchScenarioIndependentStub` | `FEW`   | 4.47 ±0.07 | 4.84 ±0.11 | 4.44 ±0.15 |
| `matchScenarioIndependentStub` | `MANY`  | 4.82 ±0.29 | **10.01 ±0.16** | **4.60 ±0.16** |
| `matchScenarioStub`            | `FEW`   | 9.30 ±0.16 | 8.86 ±0.11 | 8.95 ±0.15 |
| `matchScenarioStub`            | `MANY`  | 18.94 ±0.10 | 13.04 ±0.26 | 12.72 ±0.19 |
| `failToMatchAnyStub`           | `NONE`  | 309.7 ±4.4 | 317.2 ±7.6 | 303.2 ±1.9 |
| `failToMatchAnyStub`           | `FEW`   | 9.39 ±0.11 | 9.77 ±0.18 | 9.28 ±0.05 |
| `failToMatchAnyStub`           | `MANY`  | 312.0 ±6.0 | 352.1 ±69.7 | 305.9 ±2.6 |

### What this leaves

**The +136% row is gone.** A request matching a stub unrelated to any scenario, on a server with 200 scenarios, is back to master's cost: 3.83 vs 3.74 µs at 1 thread, 4.60 vs 4.82 at 4 threads — parity within the error bars. Same for `failToMatchAnyStub`, where no candidate ever reaches the scenario filter, so the snapshot is never built.

**One case still pays: `matchScenarioStub` with `MANY`** — 11.37 vs master's 6.04 µs at 1 thread. That is expected and unavoidable with this design: the request genuinely matches a scenario-constrained stub, so the snapshot has to be built. The cost is now confined to requests that actually use scenarios on a server with hundreds of them, rather than being charged to every request. The per-request `Lazy` allocation does not show up anywhere.

At 4 threads that same row reads 12.72 vs master's 18.94, but as noted in the previous comment that is master's bug rather than a win here — `master` strands a few percent of requests on the ~300 µs full-scan path.

If reviewers want the last ~5 µs back, the remaining option is to stop rebuilding the map per request and instead maintain an immutable scenario-state map updated on transition. Happy to do that as a follow-up, though it is a larger change to `AbstractScenarios` and I would rather not fold it into this fix.

RafeArnold (approved):

PR #3522, opened 2026-07-27
fix: handle boolean XPath results in matchesXPath (#3510) [FaaFyfxR9WAQrL7FcAgEHJvztd8cVMxvjHRS55rw1nwH]

### LINKED ISSUE #3510: matchesXPath with a boolean-valued expression matches even when the boolean is FALSE, so any XML body matches count(//x)=99, 1 = 2, or contains('abc','z')
## Proposal

**Environment:** `wiremock-standalone` 4.0.0-beta.38 and 3.13.1, both reproduced; source `master` @ `25294e8de`. JDK 17+.

A one-argument `matchesXPath` whose expression evaluates to a BOOLEAN (or NUMBER/STRING) is treated as a non-empty node set and always matches, regardless of the boolean's value. WireMock evaluates the XPath, wraps any non-node result in a single `XmlPrimitiveNode`, then tests `!nodeList.isEmpty()`, which is `true` because the list holds that one primitive node. A condition that is false still fires the stub.

Differential against an independent `javax.xml.xpath` BOOLEAN oracle, using
`XML = <solar-system><planet name='Earth' position='3'/><planet name='Mars' position='4'/></solar-system>`:

| xpath | wiremock match | correct (BOOLEAN) |
|---|---|---|
| `count(//planet) = 2` | true | true ✓ |
| `count(//planet) = 99` | **true** | false ✗ |
| `//planet[@name='Earth']/@position = '999'` | **true** | false ✗ |
| `1 = 2` | **true** | false ✗ |
| `2 > 5` | **true** | false ✗ |
| `contains('abc','z')` | **true** | false ✗ |
| `boolean(//nonexistent)` | **true** | false ✗ |

Metamorphic laws also fail, so the finding does not depend on any chosen oracle: the constant-false predicate `1 = 2` / `false()` references nothing in the document yet matches; the node-set encoding `//planet[count(//planet)=99]` correctly yields no match while the boolean encoding `count(//planet)=99` matches (self-contradiction); the two-argument form `matchingXPath("count(//planet)", equalTo("99"))` correctly returns no match while the one-argument boolean form matches. Node-set sanity holds: `//planet[@name='Jupiter']` (absent) correctly does not match, so only the primitive/boolean path is broken.

**Expected:** `matchesXPath("<false boolean expression>")` does not match. **Actual:** every well-formed XML body matches any boolean expression and fires the wrong stub. Boolean idioms (`count(//x)=N`, `//a/@b='v'`, `contains(...)`) are the natural way users write value and count conditions.

## Reproduction steps

```java
import static com.github.tomakehurst.wiremock.client.WireMock.matchingXPath;

public class Repro {
  static final String XML =
    "<solar-system><planet name='Earth' position='3'/><planet name='Mars' position='4'/></solar-system>";
  static boolean m(String xp) { return matchingXPath(xp).match(XML).isExactMatch(); }
  public static void main(String[] a) {
    System.out.println("count=2  (true)  -> " + m("count(//planet) = 2"));   // true  ✓
    System.out.println("count=99 (false) -> " + m("count(//planet) = 99"));  // true  ✗ BUG
    System.out.println("1 = 2   (false)  -> " + m("1 = 2"));                  // true  ✗ BUG
    System.out.println("contains z (false) -> " + m("contains('abc','z')")); // true  ✗ BUG
  }
}
```

```
javac -cp wiremock-standalone-4.0.0-beta.38.jar Repro.java
java  -cp .:wiremock-standalone-4.0.0-beta.38.jar Repro
```

Reproduces on `wiremock-standalone-3.13.1.jar`, so it is not a beta-only regression.

## References

- **Root cause:** `wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/xml/XmlNode.java:81`: in `toListOrSingle`, the `default ->` branch wraps any non-node result (`BOOLEAN`/`NUMBER`/`STRING`) in `new XmlPrimitiveNode<>(evaluationResult.value())`, adding exactly one node regardless of a boolean's value. `wiremock-core/.../matching/MatchesXPathPattern.java:75` then returns `MatchResult.of(nodeList != null && !nodeList.isEmpty(), ...)`, so the size-1 primitive list always matches. `XmlPrimitiveNode` was introduced for the Handlebars `xPath` helper (#2696) and reused by the matcher without guarding boolean semantics.
- **Suggested fix:** in `XmlNode.toListOrSingle`, for a `BOOLEAN` result add a node only when the value is `true` (a false boolean gives an empty list and no match); keep NUMBER/STRING wrapping for the helper and two-argument paths. Alternatively special-case a single `XmlPrimitiveNode<Boolean>` in `MatchesXPathPattern.isSimpleMatch` and match on its value rather than list non-emptiness.

---
*Found via property-based & differential bug-hunting, part of an effort to scale PBT (DepTyCheck-based) testing across the OSS ecosystem.*

*If this is intended / by-design: I'm really sorry, please just close it — no need to flag or ban me. I'm trying to scale property-based testing across the whole ecosystem and my publishing agents may have gotten this one wrong. I read every issue and follow up on each.*

waterWang:
## What

Fixes #3510

`matchesXPath` with a boolean-valued expression (e.g. `count(//planet) = 99`, `1 = 2`, `contains('abc','z')`) always matches, even when the boolean is false. The root cause is in `XmlNode.toListOrSingle()`: the `default` branch wraps any non-node XPath evaluation result in a single `XmlPrimitiveNode`, making the node list always non-empty, which `isSimpleMatch` interprets as a match.

## Fix

Add an explicit `BOOLEAN` case in `toListOrSingle()` that only adds a node when the value is `true`. A false boolean produces an empty list, which correctly causes `isSimpleMatch` to return no match. `NUMBER` and `STRING` results still use the default wrapping for the Handlebars xPath helper and two-argument paths.

## Tests

- `returnsNoExactMatchWhenBooleanXPathExpressionIsFalse` — `count(//planet) = 99` should not match
- `returnsExactMatchWhenBooleanXPathExpressionIsTrue` — `count(//planet) = 2` should match
- `returnsNoExactMatchWhenContainsExpressionIsFalse` — `contains('abc','z')` should not match
- `returnsNoExactMatchWhenComparisonExpressionIsFalse` — `1 = 2` should not match

All existing tests continue to pass.

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/matching/MatchesXPathPatternTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/matching/MatchesXPathPatternTest.java
@@ -35,6 +35,62 @@ import org.skyscreamer.jsonassert.JSONAssert;
 public class MatchesXPathPatternTest {
 
   @Test
+  public void returnsNoExactMatchWhenBooleanXPathExpressionIsFalse() {
+    String mySolarSystemXML =
+        "<solar-system>"
+            + "<planet name='Earth' position='3' supportsLife='yes'/>"
+            + "<planet name='Venus' position='4'/></solar-system>";
+
+    StringValuePattern pattern = WireMock.matchingXPath("count(//planet) = 99");
+
+    MatchResult match = pattern.match(mySolarSystemXML);
+    assertFalse(match.isExactMatch(), "Expected XPath boolean-false to not match");
+    assertThat(match.getDistance(), is(1.0));
+  }
+
+  @Test
+  public void returnsExactMatchWhenBooleanXPathExpressionIsTrue() {
+    String mySolarSystemXML =
+        "<solar-system>"
+            + "<planet name='Earth' position='3' supportsLife='yes'/>"
+            + "<planet name='Venus' position='4'/></solar-system>";
+
+    StringValuePattern pattern = WireMock.matchingXPath("count(//planet) = 2");
+
+    MatchResult match = pattern.match(mySolarSystemXML);
+    assertTrue(match.isExactMatch(), "Expected XPath boolean-true to match");
+    assertThat(match.getDistance(), is(0.0));
+  }
+
+  @Test
+  public void returnsNoExactMatchWhenContainsExpressionIsFalse() {
+    String mySolarSystemXML =
+        "<solar-system>"
+            + "<planet name='Earth' position='3' supportsLife='yes'/>"
+            + "<planet name='Venus' position='4'/></solar-system>";
+
+    StringValuePattern pattern = WireMock.matchingXPath("contains('abc','z')");
+
+    MatchResult match = pattern.match(mySolarSystemXML);
+    assertFalse(match.isExactMatch(), "Expected contains false to not match");
+    assertThat(match.getDistance(), is(1.0));
+  }
+
+  @Test
+  public void returnsNoExactMatchWhenComparisonExpressionIsFalse() {
+    String mySolarSystemXML =
+        "<solar-system>"
+            + "<planet name='Earth' position='3' supportsLife='yes'/>"
+            + "<planet name='Venus' position='4'/></solar-system>";
+
+    StringValuePattern pattern = WireMock.matchingXPath("1 = 2");
+
+    MatchResult match = pattern.match(mySolarSystemXML);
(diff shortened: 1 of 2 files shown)


tomakehurst:
Please can you run `./gradlew spotlessApply` and push the result, then we can merge this.

PR #3539, opened 2026-08-09
fix: inject Notifier into AbstractRequestHandler to support custom HttpServerFactory logging

### LINKED ISSUE #3285: Notifier doesn't take effect when using a custom HttpServerFactory

waterWang:
## Summary

When using a custom `HttpServerFactory`, the `Notifier` (set via `.notifier(ConsoleNotifier(...))`) silently doesn't take effect — only the default Jetty log line appears.

## Root Cause

`AbstractRequestHandler.handle()` and `AdminRequestHandler.handleRequest()` call `LocalNotifier.notifier()`, which reads a thread-local set by `WireMockHandlerDispatchingServlet.service()`. Custom `HttpServerFactory` implementations never call `LocalNotifier.set()`, so the thread-local defaults to `NullNotifier`.

## Fix

Inject a `Notifier` instance into `AbstractRequestHandler` via constructor, replacing the `LocalNotifier.notifier()` static call. The change flows through:

1. **`AbstractRequestHandler`** — Added `Notifier notifier` field, constructor parameter, and replaced `notifier()` calls with `notifier`.
2. **`StubRequestHandler`** — Added `Notifier` constructor parameter, passed to super.
3. **`AdminRequestHandler`** — Added `Notifier` constructor parameter, passed to super, replaced `notifier()` calls.
4. **`WireMockApp`** — Passes `options.notifier()` to both handler constructors.

The Jetty servlet's `LocalNotifier.set()` call is retained for backward compatibility with other code that still uses the thread-local pattern.

## Testing

- All existing tests pass (AdminRequestHandlerTest, JettyHttpServerTest)
- `./gradlew :wiremock-core:compileJava` — BUILD SUCCESSFUL

Fixes #3285

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/jetty/JettyHttpServerTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/jetty/JettyHttpServerTest.java
@@ -25,6 +25,7 @@ import com.github.tomakehurst.wiremock.admin.AdminRoutes;
 import com.github.tomakehurst.wiremock.common.DataTruncationSettings;
 import com.github.tomakehurst.wiremock.common.FatalStartupException;
 import com.github.tomakehurst.wiremock.common.Limit;
+import com.github.tomakehurst.wiremock.common.Notifier;
 import com.github.tomakehurst.wiremock.core.Admin;
 import com.github.tomakehurst.wiremock.core.StubServer;
 import com.github.tomakehurst.wiremock.core.WireMockConfiguration;
@@ -77,7 +78,17 @@ public class JettyHttpServerTest {
             false,
             Collections.emptyList(),
             Collections.emptyList(),
-            NO_TRUNCATION);
+            NO_TRUNCATION,
+            new Notifier() {
+              @Override
+              public void info(String message) {}
+
+              @Override
+              public void error(String message) {}
+
+              @Override
+              public void error(String message, Throwable t) {}
+            });
     stubRequestHandler =
         new StubRequestHandler(
             Mockito.mock(StubServer.class),
@@ -90,7 +101,17 @@ public class JettyHttpServerTest {
             Collections.emptyList(),
             false,
             NO_TRUNCATION,
-            new PlainTextStubNotMatchedRenderer(Extensions.NONE));
+            new PlainTextStubNotMatchedRenderer(Extensions.NONE),
+            new Notifier() {
+              @Override
+              public void info(String message) {}
+
+              @Override
+              public void error(String message) {}
+
+              @Override
+              public void error(String message, Throwable t) {}
+            });
     messageStubRequestHandler =
         new MessageStubRequestHandler(
             messageStubMappings,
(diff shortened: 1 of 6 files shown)


src/test/java/com/github/tomakehurst/wiremock/jetty/JettyHttpServerTest.java
```diff
@@ -77,7 +78,17 @@ public void init() {
             false,
             Collections.emptyList(),
             Collections.emptyList(),
-            NO_TRUNCATION);
+            NO_TRUNCATION,
+            new Notifier() {
```
tomakehurst, line 82: Could we use `TestNotifier` here in order to be less verbose?

src/test/java/com/github/tomakehurst/wiremock/jetty/JettyHttpServerTest.java
```diff
@@ -90,7 +101,17 @@ public void init() {
             Collections.emptyList(),
             false,
             NO_TRUNCATION,
-            new PlainTextStubNotMatchedRenderer(Extensions.NONE));
+            new PlainTextStubNotMatchedRenderer(Extensions.NONE),
+            new Notifier() {
```
tomakehurst, line 105: Could we use `TestNotifier` here in order to be less verbose?

src/test/java/com/github/tomakehurst/wiremock/stubbing/AdminRequestHandlerTest.java
```diff
@@ -64,7 +65,17 @@ public void init() {
             false,
             Collections.emptyList(),
             Collections.emptyList(),
-            new DataTruncationSettings(Limit.UNLIMITED));
+            new DataTruncationSettings(Limit.UNLIMITED),
+            new Notifier() {
```
tomakehurst, line 69: Could we use `TestNotifier` here in order to be less verbose?

tomakehurst (changes requested):
I suggest also adding an explicit log message + assertion for it in `ProgrammaticHttpServerAcceptanceTest`

PR #3546, opened 2026-08-10
fix: handle Bazel sandbox symlinks in assertFilePathIsUnderRoot [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

### LINKED ISSUE #2964: assertFilePathIsUnderRoot reports false positive under Bazel
### Proposal

When using Bazel's sandbox feature AbstractFileSource.assertFilePathIsUnderRoot() reports a false positive error.

E.g. I set up a mock with `"bodyFileName": "responses/features.xml",` then I get an error:

```
com.github.tomakehurst.wiremock.security.NotAuthorisedException: Access to file responses/features.xml is not permitted. An absolute path from the filesystem root might be specified. Your file must be in /home/user/.cache/bazel/_bazel_user/7f83a0b1f0940f88aec1f4530f927169/sandbox/processwrapper-sandbox/447/execroot/my-project/bazel-out/k8-fastbuild/bin/my-package/src/test/some_test.runfiles/my-project/my-package/src/test/resources/wiremock/__files
```

### Reproduction steps

After looking at AbstractFileSource.assertFilePathIsUnderRoot()'s source I think there is an issue with canonical paths and Bazel's sandboxing.

Created the following repro (converting the method's code to Scala and adding some test values):

```scala
import java.io.File
import java.nio.file.Paths

val path = "./__files/responses/features.xml"
val rootDirectory = new File("/home/user/.cache/bazel/_bazel_user/7f83a0b1f0940f88aec1f4530f927169/sandbox/processwrapper-sandbox/448/execroot/my-project/bazel-out/k8-fastbuild/bin/my-package/src/test/some_test.runfiles/my-project/my-package/src/test/resources/wiremock")

val rootPath = rootDirectory.getCanonicalPath

val file = new File(path)
val filePath =
  if (file.isAbsolute)
    new File(path).getCanonicalPath
  else
    new File(rootDirectory, path).getCanonicalPath
    
println(filePath)
println(rootPath)
println(Paths.get(filePath).normalize())

if (!Paths.get(filePath).normalize().startsWith(rootPath)) {
  println("Unauthorized!")
}
```

And this prints:

```
/home/user/Work/my-project/my-package/src/test/resources/wiremock/__files/responses/features.xml
/home/user/.cache/bazel/_bazel_user/7f83a0b1f0940f88aec1f4530f927169/sandbox/processwrapper-sandbox/448/execroot/my-project/bazel-out/k8-fastbuild/bin/my-package/src/test/some_test.runfiles/my-project/my-package/src/test/resources/wiremock
/home/gege/Work/my-project/my-package/src/test/resources/wiremock/__files/responses/features.xml
Unauthorized!
```

To be able to reproduce this the Bazel sandbox with id 448 needs to be "active", meaning, it needs to have a tree of symlinks inside it (while my test is executing).
I think it's the symlinks that cause this, as they link to another root.

### References

_No response_

waterWang:
### What

Fix `AbstractFileSource.assertFilePathIsUnderRoot()` to handle filesystems where the root directory contains symlinks to files outside of it, such as Bazel's sandboxed runfiles.

### Why (context)

Under Bazel's sandbox, the root directory (e.g., `.../sandbox/NNN/execroot/.../wiremock`) is a real directory, but the files inside it are symlinks to the source tree. When `getCanonicalPath()` is called on such a file, it resolves the symlink to the real source path (e.g., `/home/user/Work/.../features.xml`), which is outside the root directory. This causes the security check to incorrectly throw `NotAuthorisedException`.

See issue #2964 for full reproduction details.

### How

The fix adds a fallback check using the **absolute path** (which does not resolve symlinks). The method now checks both:

1. The **canonical path** (with symlinks resolved) — catches path traversal attacks like `../../../etc/passwd`
2. The **absolute path** (without resolving symlinks) — allows access through symlinks that are legitimately inside the root directory, such as Bazel's sandboxed runfiles

A file is only rejected when **both** checks fail, ensuring backwards compatibility and security.

### Testing

- All existing tests continue to pass
- Added a new test `allowsReadingFileViaSymlinkInsideRootWhenCanonicalPathIsOutsideRoot` that creates a symlink inside the root pointing to a file outside of it, and verifies that reading the file via the symlink does not throw `NotAuthorisedException`

Closes #2964

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/common/SingleRootFileSourceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/common/SingleRootFileSourceTest.java
@@ -18,13 +18,16 @@ package com.github.tomakehurst.wiremock.common;
 import static com.github.tomakehurst.wiremock.testsupport.TestFiles.filePath;
 import static com.github.tomakehurst.wiremock.testsupport.WireMatchers.fileNamed;
 import static com.github.tomakehurst.wiremock.testsupport.WireMatchers.hasExactlyIgnoringOrder;
+import static java.nio.charset.StandardCharsets.UTF_8;
 import static org.hamcrest.MatcherAssert.assertThat;
 import static org.hamcrest.Matchers.is;
 import static org.junit.jupiter.api.Assertions.*;
+import static org.junit.jupiter.api.Assumptions.assumeTrue;
 
 import com.github.tomakehurst.wiremock.security.NotAuthorisedException;
 import java.io.File;
 import java.io.FileNotFoundException;
+import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.nio.file.Paths;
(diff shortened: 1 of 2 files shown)


src/test/java/com/github/tomakehurst/wiremock/common/SingleRootFileSourceTest.java
```diff
@@ -231,4 +234,28 @@ void throwsExceptionWhenAttemptingToGetStreamForNonExistentFile() {
...
+    } catch (IOException | UnsupportedOperationException e) {
+      // Symbolic links may not be supported on this platform (e.g. Windows without privileges)
+      assumeTrue(false, "Symbolic links not supported on this platform");
+      return;
+    }
+
+    SingleRootFileSource fileSource = new SingleRootFileSource(rootDir.toFile());
+
+    // Must NOT throw NotAuthorisedException
+    fileSource.getBinaryFileNamed("aliased.xml");
```
tomakehurst, line 259: An explicit `assertDoesNotThrow(...)` here would be good

PR #3554, opened 2026-08-14
Fix NPE when scenario name generated from a request URL matcher that is not URL or URL path

tomakehurst:
ScenarioProcessor and SnapshotStubMappingBodyExtractor both derived a request's path via getUrl()/getUrlPath(), which only return a value for urlEqualTo/urlPathEqualTo matchers. Any other UrlPattern (e.g. one set by a recording StubMappingTransformer via urlPathTemplate, urlMatching, etc.) made both return null, causing an NPE during snapshot recording.

Urls.urlPatternToPathSegments() normalises any UrlPattern generically by parsing its own toString() description instead of switching on concrete matcher classes, so it keeps working if new UrlPattern subtypes are added.

## References

- TODO

## Submitter checklist

### DIFF
--- a/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
+++ b/wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
@@ -16,6 +16,8 @@
 package com.github.tomakehurst.wiremock.common;
 
 import com.github.tomakehurst.wiremock.http.QueryParameter;
+import com.github.tomakehurst.wiremock.matching.UrlPattern;
+import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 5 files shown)


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
```diff
@@ -44,4 +46,40 @@ public static String urlToPathParts(PathAndQuery uri) {
...
+  /**
+   * Normalises the value matched by any kind of {@link UrlPattern} (e.g. one created via {@code
+   * urlEqualTo}, {@code urlMatching}, {@code urlPathEqualTo}, {@code urlPathMatching} or {@code
+   * urlPathTemplate}) into its path segments.
+   *
+   * <p>This works generically across all current and future {@link UrlPattern} subtypes by parsing
+   * the matcher's own {@code toString()} description, rather than switching on specific pattern
+   * classes.
+   */
+  public static List<String> urlPatternToPathSegments(UrlPattern urlPattern) {
```
leeturner, line 59: I guess these things don't change all that much but parsing a urlPattern toString feels a little brittle.  Maybe we could use the expected value instead.  This is typed at least:

```java
    public static List<String> urlPatternToPathSegments(UrlPattern urlPattern) {                                                                                                                                                                                                                    
      if (!urlPattern.isSpecified()) {                                                                                                                                                                                                                                                              
        return List.of();                                                                                                                                                                                                                                                                           
      }                                                                                                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                    
      String pathOnly = urlPattern.getExpected().split("\\?", 2)[0];                                                                                                                                                                                                                                
      return Arrays.stream(pathOnly.split("/")).filter(part -> !part.isEmpty()).toList();                                                                                                                                                                                                           
    } 

public static String urlPatternToPathParts(UrlPattern urlPattern) {                                                                                                                                                                                                                             
      return String.join("-", urlPatternToPathSegments(urlPattern));                                                                                                                                                                                                                                
    } 
```

tomakehurst: Good catch, that's much nicer

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/common/Urls.java
```diff
@@ -44,4 +46,40 @@ public static String urlToPathParts(PathAndQuery uri) {
...
+        description.startsWith("path and query ")
+            ? description.substring("path and query ".length())
+            : description.startsWith("path ")
+                ? description.substring("path ".length())
+                : description;
+
+    int firstSpaceIndex = withoutScopePrefix.indexOf(' ');
+    String matchedValue =
+        firstSpaceIndex >= 0 ? withoutScopePrefix.substring(firstSpaceIndex + 1) : "";
+    String pathOnly = matchedValue.split("\\?", 2)[0];
```
leeturner: Will this break for matchers like `urlMatching` and `urlPathMatching` where ? is part of the regex?  Perhaps it should be gated on `urlPattern.isRegex()` ?

leeturner: Also wondering if `PathAndQuery.parse` would be a good way to do this ?

tomakehurst: Added better logic and some more tests for this

leeturner (approved):

PR #3584, opened 2026-09-08
Clear the request cache when request handling fails

FuriousRec:
## Problem

`AbstractRequestHandler.handle()` calls `RequestCache.onRequestEnd()` only when processing completes normally.

If request handling, response rendering, sending, or an after-response callback throws, cleanup is skipped. Since `RequestCache` uses a `ThreadLocal`, the cache remains attached to the worker thread. A subsequent request on that thread can reuse the previous cache, and cached objects remain referenced.

This was identified through code inspection, not a reported production incident.

## Change

Move cache cleanup into a `finally` block around the existing request lifecycle. This ensures cleanup happens on both success and failure while preserving processing order and allowing the original exception to propagate.

## Validation

Added five regression tests covering successful completion and failures during handling, rendering, sending, and the after-response callback.

The Java tests and formatting checks have **not been run**. Source comparison confirmed the existing processing logic is unchanged apart from the cleanup wrapper, and `git diff --check` passed.

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandlerTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/http/AbstractRequestHandlerTest.java
@@ -1,0 +1,140 @@
+/*
+ * Copyright (C) 2011-2026 Thomas Akehurst
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


Mahoney:
I don't think you can expect a PR like this to be considered seriously without offering some explanation for the change you are making. Neither the commit message nor the PR body explain why you want this.

FuriousRec:
Thanks for pointing this out. I submitted the PR with the template still empty—sorry for omitting the reasoning.

`AbstractRequestHandler.handle()` clears `RequestCache` only after normal completion. If handling, rendering, sending, or an after-response callback throws, cleanup is skipped. Because the cache is stored in a `ThreadLocal`, it remains attached to the worker thread and can carry over to the next request on that thread.

The proposed `finally` block ensures cleanup also happens on failure while preserving the original exception. I added regression tests covering those paths, but have not run them yet, so the change still needs validation.

This was identified through code inspection, rather than a production incident.

PR #3590, opened 2026-09-12
Honor the configured HTTPS trust store type

### LINKED ISSUE #1484: WireMock doesn't allow KeyStore and TrustStore of different types
I am trying to create a WireMock configuration:

`WireMockServer wireMockServer = new WireMockServer(wireMockConfig()`
`    .httpsPort(8443)`
`    .needClientAuth(true)`
`    .trustStorePath(System.getProperty("javax.net.ssl.trustStore"))`
`    .trustStorePassword(System.getProperty("javax.net.ssl.trustStorePassword"))`
`    .trustStoreType(System.getProperty("javax.net.ssl.trustStoreType"))`
`    .keyStorePath(System.getProperty("javax.net.ssl.keyStore"))`
`    .keyStorePassword(System.getProperty("javax.net.ssl.keyStorePassword"))`
`    .keyStoreType(System.getProperty("javax.net.ssl.keyStoreType")));`
`  `
`wireMockServer.start()`

Where the trustStoreType is set to "JKS" and the keyStoreType is set to "pkcs12".

When doing this, I receive the error:

`java.io.IOException: DerInputStream.getLength(): lengthTag=109, too big.`

Tracing the issue, it is being thrown from the `org.eclipse.jetty.util.ssl.SslContextFactory.loadTrustStore(Resource resource)` method.  This method looks for the trustStoreType, and if that is null it uses the keyStoreType when it tries to load the store.  In my scenario, the trustStoreType is null even though I set it in the configuration.  So in this case, it is using the trustStorePath, trustStorePassword, and keyStoreType to try to load the truststore and this throws the exception.

I believe the reason that the trustStoreType is null occurs in the `com.github.tomakehurst.wiremock.jetty94.SslContexts.setupClientAuth(SslContextFactory.Server sslContextFactory, HttpsSettings httpsSettings)` method.  In that method, when the httpsSettings have a trustStore, the path and the password are set but the type is not set.  I believe the trustStoreType needs to be set here in order to allow a trustStoreType that is different than the keyStoreType.

Can you verify if this is a bug?  If it is not a bug can you recommend a way to use a trustStore and keyStore of different types?

Thank you!

xujiantop-crypto:
Fixes #1484.

When HTTPS client authentication uses different key-store and trust-store formats, WireMock can fail to start with an invalid key-store format error. `HttpsSettings` carries `trustStoreType`, but the Jetty SSL context only receives the trust-store path and password, leaving Jetty to fall back to the key-store type.

Pass the configured trust-store type through the shared client-authentication setup used by the HTTP/1.1, HTTP/2 and browser-proxy SSL contexts. The acceptance test uses a JKS key store and generates a JCEKS trust store from the existing client certificates, then makes an authenticated HTTPS request with HTTP/2 enabled and disabled. JCEKS makes the missing type reproducible even with JDKs that auto-detect JKS and PKCS12.

Both new cases fail before the production change with `Invalid keystore format` during server startup.

Validation on Windows / Temurin 25: 29 tests, 0 failures, 27 passed and 2 skipped (the existing platform-specific socket-reset test and disabled unused-method rule). Command:

```
./gradlew :test --tests '*HttpsAcceptanceTest' --tests '*HttpsBrowserProxyClientAuthAcceptanceTest' --tests '*Http2DisabledAcceptanceTest'
```

The changed Java files pass Google Java Format 1.17.0, the version configured by the project. The formatter was invoked directly because Spotless cannot locate the Git repository in a linked worktree (and the build disables it there by default).

### DIFF
--- a/src/test/java/com/github/tomakehurst/wiremock/HttpsAcceptanceTest.java
+++ b/src/test/java/com/github/tomakehurst/wiremock/HttpsAcceptanceTest.java
@@ -32,11 +32,15 @@ import com.github.tomakehurst.wiremock.http.client.apache5.ApacheHttpClientFacto
 import com.github.tomakehurst.wiremock.stubbing.ServeEvent;
 import java.io.FileInputStream;
 import java.io.IOException;
+import java.io.OutputStream;
 import java.net.SocketException;
+import java.nio.file.Files;
+import java.nio.file.Path;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
 import java.security.NoSuchAlgorithmException;
 import java.security.cert.CertificateException;
+import java.util.Collections;
 import javax.net.ssl.SSLContext;
 import javax.net.ssl.SSLException;
 import javax.net.ssl.SSLHandshakeException;
@@ -60,6 +64,9 @@ import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.DisabledOnOs;
 import org.junit.jupiter.api.condition.OS;
+import org.junit.jupiter.api.io.TempDir;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ValueSource;
 
 class HttpsAcceptanceTest {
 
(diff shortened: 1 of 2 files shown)


tomakehurst:
Hi, thanks for creating this.

Please can you do `./gradlew spotlessApply` and push the result?

Then this will be ready to merge if the build is green.

tomakehurst:
Please can you `./gradlew spotlessApply` and push the results, then I think we'll be ready to merge

PR #3592, opened 2026-09-14
fix: keep the first part when a multipart form repeats a field name

### LINKED ISSUE #3591: Multipart form data requests fails with global templating on.
### Summary

When the Wiremock has 'globalTemplating' set to true and receives a multipart form data request with multiple parts having the same name it fails with  500 error.

### Reproduction steps

Using latest stable version 3.13.2.

- Enable 'globalTemplating'
- Make a multipart form data request with multiple parts having same name

### References

The issue comes from this code in `TemplateEngine`
```java
return request.getParts().stream()
            .collect(
                Collectors.toMap(
                    Request.Part::getName,
                    part ->
                        new RequestPartTemplateModel(
                            part.getName(),
                            part.getHeaders().all().stream()
                                .collect(
                                    Collectors.toMap(
                                        HttpHeader::key,
                                        header -> ListOrSingle.of(header.values()))),
                            part.getBody())));```
`Collectors.toMap` hitting the same key again throws `IllegalStateException` and request processing fails early in the pipeline.

### PR DESCRIPTION
**Pull request by Voyagerroc-Lab**
### What

A `multipart/form-data` request whose parts repeat a field name (`curl -F 'thename=one' -F 'thename=two' ...`) answered **500** whenever templating was on. `TemplateEngine.buildRequestPartModel` collected the parts into a map with a merge function that always threw:

```java
Collectors.toMap(..., (e1, e2) -> { throw new IllegalStateException("Duplicate request part name"); }, LinkedHashMap::new)
```

Repeating a field name is ordinary form encoding, so the model now keeps the first part with that name - the same choice `Request#getPart` already makes with `findFirst()` - rather than failing the response.

Fixes #3591.

### Evidence

New test in `MultipartTemplatingAcceptanceTest`: two `thename` text parts posted to a stub with `globalTemplating(true)`.

Before the change:

```
Expected: is <200>
     but: was <500>
```

After:

```
multipart:true
part count = 1
first=one
```

### Tests run

- `./gradlew :test --tests "com.github.tomakehurst.wiremock.MultipartTemplatingAcceptanceTest"` - 8/8 pass.
- `./gradlew :test --tests "com.github.tomakehurst.wiremock.extension.responsetemplating.*"` - 563 tests, 1 failure: `ResponseTemplateTransformerTest.templatizeBodyFile` (`NotFoundException: Not found in blob store: /greet-Ram.txt`). It fails identically on an unmodified checkout on this Windows machine, so it is pre-existing and unrelated to this change.
- `./gradlew spotlessApply` (Google Java format) applied.

### Scope note

`request.parts` is keyed by name, so a duplicated name can only surface as one entry - `part count = 1` above is that shape showing through, not a second defect. Making repeated parts individually addressable would change the template model's shape and is deliberately left out of this fix.


tomakehurst:
Thanks!

stlhrt:
Will it be fixed in 3.x version too?

tomakehurst:
No, 3.x is frozen now. High priority security patches only.

lukasz-wozniak-ipfd:
Can you please release next beta release with this fix?

PR #3594, opened 2026-09-15
Server sent events

tomakehurst:
Add support for server-sent events as an async message type


wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/HttpStubServeEventListener.java
```diff
@@ -69,9 +70,8 @@ public void afterMatch(ServeEvent serveEvent, Parameters parameters) {
     }
 
     List<MessageStubMapping> matchingStubs = findMatchingMessageStubs(serveEvent);
-    for (MessageStubMapping stub : matchingStubs) {
-      executeActions(stub, serveEvent);
-    }
+    Optional<MessageStubMapping> firstMatch = matchingStubs.stream().findFirst();
+    firstMatch.ifPresent(stub -> executeActions(stub, serveEvent));
```
RafeArnold, line 74: we now only want the first matching stub to execute its actions? why has this changed?

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/WireMockHandlerDispatchingServlet.java
```diff
@@ -269,6 +274,37 @@ private void performWebSocketUpgrade(Request request) {
...
+      } catch (IOException e) {
+        throwUnchecked(e);
+      }
+      AsyncContext asyncContext = httpServletRequest.startAsync();
+      asyncContext.setTimeout(0);
+      try {
+        JettySseSession sseSession = new JettySseSession(asyncContext);
+        LoggedRequest snapshot = LoggedRequest.createFrom(request);
+        SseMessageChannel channel = new SseMessageChannel(snapshot, sseSession);
+        messageStubRequestHandler.getMessageChannels().add(channel);
```
leeturner, line 302: The Websocket code has onWebSocketClose to allow the channel to be removed. We don't seem to have any equivalent here. It looks like they never get removed

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/HttpStubServeEventListener.java
tomakehurst: This was a bug unrelated to this PR. To behave in an equivalent way to HTTP stubbing we only want to match one stub, stop and execute its actions, not match every stub and execute every action.

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/WireMockHandlerDispatchingServlet.java
tomakehurst: Fixed in 59263fb — JettySseSession now takes an onClose callback, invoked when the session terminates (explicit close, or a write failure after the client disconnects), which removes the channel from the registry. That's the equivalent of onWebSocketClose for SSE. Covered by the channelIsRemovedOnSseDisconnect acceptance test.

tomakehurst:
**Updates since the initial push:**

- **Message headers** — `MessageDefinition`/`Message` now carry multi-valued `MessageHeaders` (`MessageHeader extends MultiValue`, same JSON conventions as `HttpHeaders`), replacing the SSE-specific `eventName`/`id` fields. Reserved keys use native wire names and are interpreted per channel type — SSE reads `event` and `id`; a future Kafka binding would own `key`/`topic` the same way. Header values are templated alongside the body.
- **SSE event name & id** — new DSL: `sendSse("{...}").withEventName("userLogin").withEventId("42")`, plus generic `withHeader(...)`. Multi-valued `event`/`id` is rejected at mutation time; the journal serialises messages in object form (`{"body": ..., "headers": {...}}`).
- **Channel cleanup** — SSE channels are now removed on client disconnect, the equivalent of `onWebSocketClose` (addresses @leeturner's review comment).
- Misc: `onSseChannelFromRequestMatching()` DSL parity with WebSocket; test client consumes SSE via the stable `okhttp-sse` module instead of a hand-rolled parser.

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/sse/JettySseSession.java
```diff
@@ -0,0 +1,118 @@
...
+
+  @Override
+  public void sendEvent(String eventName, String data, String id) {
+    if (!open) {
+      return;
+    }
+    try {
+      if (id != null) {
+        ServletOutputStream out = asyncContext.getResponse().getOutputStream();
+        out.write(("id: " + id + "\r\n").getBytes(StandardCharsets.UTF_8));
```
leeturner, line 55: Does the id need to be sanitised in any way? Is the id templatable? Just wondering thinking about if the id contains `\r\n` for example

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/sse/JettySseSession.java
```diff
@@ -0,0 +1,118 @@
...
+    this.emitter = HELPER.newEmitter(asyncContext);
+  }
+
+  @Override
+  public boolean isOpen() {
+    return open;
+  }
+
+  @Override
+  public void sendEvent(String eventName, String data, String id) {
```
leeturner, line 48: If a message contains an id but no body (as in the body is null when Entity.EMPTY) then this will corrupt the event stream because the id only event will not be terminated correctly via the standard blank line

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/MessageHeaders.java
```diff
@@ -0,0 +1,150 @@
...
+  }
+
+  public MessageHeader getHeader(String key) {
+    if (!headers.containsKey(new CaseInsensitiveKey(key))) {
+      return MessageHeader.absent(key);
+    }
+
+    Collection<String> values = headers.get(new CaseInsensitiveKey(key));
+    return new MessageHeader(key, values);
+  }
```
leeturner, line 77:
```suggestion
public MessageHeader getHeader(String key) {                                                                                                                                                                                                                                                                                   
    Collection<String> values = headers.get(new CaseInsensitiveKey(key));                                                                                                                                                                                                                                                        
    return values.isEmpty() ? MessageHeader.absent(key) : new MessageHeader(key, values);                                                                                                                                                                                                                                        
  } 
```

wiremock-core/src/main/java/com/github/tomakehurst/wiremock/message/sse/SseMessageChannel.java
```diff
@@ -0,0 +1,88 @@
...
+  public boolean isOpen() {
+    return session != null && session.isOpen();
+  }
+
+  @Override
+  public void sendMessage(Message message) {
+    if (session != null && session.isOpen()) {
+      String data;
+      if (message.isBinary()) {
+        data = java.util.Base64.getEncoder().encodeToString(message.getBodyAsBytes());
```
leeturner, line 71: Still don't understand why AI uses fully qualified classes sometimes and not others

wiremock-jetty/src/main/java/com/github/tomakehurst/wiremock/jetty/sse/JettySseSession.java
tomakehurst: Probably a good idea to reject values that have line breaks in them at all when the stubs are created

